PR #684, opened 2020-10-19
[WIP] Android OS: how to integrate 1DS C++ SDK using Binder IPC

maxgolov:
This is minimal trivial Binder IPC example of Android OS (platform-level) client-server application.

It implements the min API that is somewhat similar to 1DS C++ SDK C API (`mat.h`). But also architected to allow sending binary blobs of data of given content type...

What it gives exactly?
- ability to open Telemetry System remotely (multi-tenancy).
- ability to pass configuration and obtain a handle (which can be directly bound to 1DS C API `evt_open` as-is).
- ability to pass any event data (binary blob, e.g. Bond serialized, or MessagePack, or JSON) from a Binder client to Binder server (privileged daemon running on device), where server decides how to handle the incoming telemetry event. Wired to `writeEvent` Binder IPC call.
- we can also implement a method in `LogManagerImpl` to accept already serialized blobs of data, such as remote client performs API call, serialization, decoration, then once the blob is prepared (we do it on calling thread), we momentarily pass it to server over Binder. Binder is fast. Binder is optimized for high-perf. This API call would also be wired to the `writeEvent` Binder API shown below.

```
interface ITelemetryAgent {

    // open TelemetrySystem instance with given configuration
    long open(String config);

    // close TelemetrySystem instance
    long close(long id);

    // clear data
    long clear(long id);

    // (re)configure TelemetrySystem instance
    long config(long id, String config);

    // log event data in serialized format (Bond, Text, JSON, MessagePack)
    long writeEvent(long id, long contentType, in byte[] data);

    // log blob data
    long writeBlob(long id, in FileDescriptor pfd, String name, String contentType, long byteCount);

    // log additional metadata that could indicate the agent to gather additional system-level detail
    long writeMetadata(long id, in String[] metadata);

    // pause TelemetrySystem
    long pause(long id);

    // resume TelemetrySystem
    long resume(long id);

    // upload TelemetrySystem
    long upload(long id);

    // stop TelemetrySystem
    long stop();

    // get current operational mode: enabled, disabled, paused, offline
    long opmode(long mode);

    // Obtain TelemetrySystem version
    String version();

    // TODO: nice to have to get some feedback from the Agent, such as property lists or counters
}
```

What is done:
- example binder client
- example binder server
- 1DS C API bound to binder client API, which invokes IPC method call
- binder server forwards incoming 1DS C API client requests to the actual 1DS C++ SDK instance

That way we split the 1DS SDK into two parts:
- one is lean 1DS API client; and
- the other one is the actual SDK - *Telemetry Agent* - the IPC server that handles event offline storage, event batching and upload. That *Telemetry Agent* process may also enrich event contents with additional system-level diagnostic data.

What is missing:
- I added 1DS C API interface ( `mat.h` API ) - same API that our lean Linux customers are using. BUT I haven't yet added the full-featured 1DS C++ SDK that most of our application-level clients are using. Nevertheless, this concept 'as-is' is already usable for native C code components that may forward their telemetry thru Android OS system-level agent (daemon).
- I did not add the actual wiring for the daemon to start from `init` (Android's service management layer), so it gets started manually as `root` user from adb shell. Script is provided that shows how to start it.
- SELinux is off, thus the agent can do anything. The real-world integration on actual Surface Duo device requires the proper SELinux rules setup for the agent process, to confine it to what's allowed and restrict it from doing anything it wants... There has to be clear definition of what the Agent is allowed to perform (net, binder, certain scoped storage location access, obtaining kernel log, obtaining crash dumps / "tombstones", etc.), but not necessarily accessing any user files nor data. This is gonna be a bad tone from privacy / compliance perspective if the Agent has unrestricted device access!

End-game goal:

Onboard Surface Duo OS components to 1DS C++ SDK, using Binder IPC as channel allowing components to send data to centralized Telemetry Agent.

Telemetry Clients (apps and services) instrumented with 1DS C++ SDK (emitters of telemetry) are:
- extremely slim and thin
- clients no longer need SQLite
- clients no longer need RAM queue - they synchronously pass events to Telemetry Agent
- clients no longer need HTTP stack - Agent would upload the data; thus clients could be totally offline components, network-restricted
- clients could be system-level min-size binary services
- clients could be kernel mode drivers - using a 10KB small-sized user mode helper binary (`sendEvent`), that knows how to send telemetry blobs to Telemetry Agent over the Binder protocol...

Telemetry Agent:
- acts as system daemon Binder server.
- reuses existing 1DS C++ SDK for storage, batching and uploads.

It may be practical to split the Telemetry Agent into more modular design, where:
- `Telemetry Agent` - is privileged enough to collect system data, counters, encrypt and save in some trusted location; while
- `Telemetry Uploader` - acts as a separate lower-privileged Java layer App Service that can only perform data upload, fetching data from some known "drop box" location on a filesystem, where `Telemetry Agent` stores it.

This much better, more modular design provides better security and trust model: if somebody breaks into one of the elements, breaks into one of the rooms, they won't be able to exploit / take over the whole house. For example, if there is exploitable flaw in Telemetry Agent, it would still be impossible to use it for any remoting or networking - we can restrict Telemetry Agent from accessing network completely using SELinux rules. And only Telemetry Uploader in that model would be capable of pushing data, and only the data stored in some location. All this can be configured via SELinux ruleset.

### DIFF
--- a/Solutions/Clienttelemetry/Clienttelemetry.vcxitems
+++ b/Solutions/Clienttelemetry/Clienttelemetry.vcxitems
@@ -66,6 +66,7 @@
     <ClCompile Include="$(MSBuildThisFileDirectory)..\..\lib\utils\FileUtils.cpp" />
     <ClCompile Include="$(MSBuildThisFileDirectory)..\..\lib\utils\StringConversion.cpp" />
     <ClCompile Include="$(MSBuildThisFileDirectory)..\..\lib\utils\StringUtils.cpp" />
+    <ClCompile Include="$(MSBuildThisFileDirectory)..\..\lib\utils\ZlibUtils.cpp" />
     <ClCompile Include="$(MSBuildThisFileDirectory)..\..\lib\utils\Utils.cpp" />
   </ItemGroup>
   <ItemGroup>
@@ -195,6 +196,7 @@
     <ClInclude Include="$(MSBuildThisFileDirectory)..\..\lib\utils\FileUtils.hpp" />
     <ClInclude Include="$(MSBuildThisFileDirectory)..\..\lib\utils\StringConversion.hpp" />
     <ClInclude Include="$(MSBuildThisFileDirectory)..\..\lib\utils\StringUtils.hpp" />
+    <ClInclude Include="$(MSBuildThisFileDirectory)..\..\lib\utils\ZlibUtils.hpp" />
     <ClInclude Include="$(MSBuildThisFileDirectory)..\..\lib\utils\Utils.hpp" />
   </ItemGroup>
   <ItemGroup>
--- a/build-xamarin.sh
+++ b/build-xamarin.sh
@@ -1,5 +1,19 @@
 #!/bin/sh
 
+if [ "$1" == "help" ] || [ "$1" == "?" ]; then
+    echo
+    echo "build-xamarin.sh usage:"
+    echo "./build-xamarin.sh [debug|release] [cleanAll|cleanXamarin] [xamarinOnly] [package]"
+    echo
+    echo "- debug|release: build configuration to use. Default to release of not specified"
+    echo "- cleanAll: deletes output and temporary directories for all platforms"
+    echo "- cleanXamarin: deletes output and temporary directories for the Xamarin solution only"
+    echo "- xamarinOnly: only builds the Xamarin solution, assuming that iOS and Android SDKs have already been built"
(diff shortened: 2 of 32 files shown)


maxgolov:
I might need to update this, and would like to propose a feature / contribution for 1DS that'd carry this on..
Might be needing it for my current assignment. Thus, not closing it yet.

PR #733, opened 2020-11-23
Windows perf fix for expensive string conversion method when logging.

jeffsuth:
UWP apps running on HoloLens (both x86 and ARM) were seeing huge performance hits causing stalls of upwards of 300ms for logs through Logger::LogEvent due to string conversion in WstringToUTF8.

The constructor for std::wstring_convert<std::codecvt_utf8<wchar_t>, wchar_t> was determined to be the culprit. The main expense was compounded by being inside the inner loop of EventProperties::PopulateEventProperties for dynamic types (gets called many times per log). This was profiled and determined to be very expensive on PC as well as HoloLens, in x86, x64 and ARM targets.

This change is for the conversion from the STL conversion method to the platform method via WideCharToMultiByte. The reverse analog MultiByteToWideChar is already being used for the reverse conversion in the function UTF8ToWstring below in the file.

The performance improvement on this function on the inner loop of the conversion on HoloLens 1 (x86) is about 400x, and on HoloLens 2 (ARM) about 200x. Our frame stalls due to telemetry have been completely eliminated by this change.


maxgolov:
@bliptec did another fix recently in #787 .. need to consolidate all usage of C++11 codecvt , and use Win32 API again (back to the old school!).

ThomsonTan:
Close the PR for now. Feel free to reopen it if it is still valid.

PR #758, opened 2021-01-05
Mitigation for OfflineRoom crashes

larvacea:
Attempt to mitigate crashes during upload on Android.

We are seeing crashes in production on Android, with GetAndReserveRecords in the call stacks. Let's attempt to prevent the crashes via try/catch, and log the errors.

### DIFF
--- a/lib/jni/LogManager_jni.cpp
+++ b/lib/jni/LogManager_jni.cpp
@@ -10,12 +10,12 @@
 #endif
 #endif
 
+#include <utils/Utils.hpp>
 #include "JniConvertors.hpp"
 #include "LogManagerBase.hpp"
 #include "WrapperLogManager.hpp"
 #include "android/log.h"
 #include "config/RuntimeConfig_Default.hpp"
-#include <modules/dataviewer/DefaultDataViewer.hpp>
 
 using namespace MAT;
 
--- a/lib/offline/OfflineStorage_Room.cpp
+++ b/lib/offline/OfflineStorage_Room.cpp
@@ -3,21 +3,15 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 #include "OfflineStorage_Room.hpp"
+#include "pal/PAL.hpp"
 #include <cmath>
 #include <exception>
 #include <jni.h>
 
-#ifndef NDEBUG
 namespace
 {
     static constexpr bool s_throwExceptions = true;
 }
-#else
-namespace
-{
-    static constexpr bool s_throwExceptions = false;
-}
-#endif
 
 namespace MAT_NS_BEGIN
 {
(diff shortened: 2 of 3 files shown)


lib/jni/LogManager_jni.cpp
```diff
@@ -10,12 +10,12 @@
 #endif
 #endif
 
+#include <utils/Utils.hpp>
```
maxgolov, line 13: Very minor nit: Edge team usually reserved the `<...>` for system headers. Thus, we usually have the `"..."` elsewhere. But I do not personally have a strong opinion on enforcing this stylistic requirement.

lib/offline/OfflineStorage_Room.cpp
```diff
@@ -82,9 +76,17 @@ namespace MAT_NS_BEGIN
...
-            ++push_count;
+        try {
+            if (env->PushLocalFrame(frameSize) == JNI_OK) {
+                ++push_count;
+            }
+        }
+        catch (std::exception e) {
+            auto what = e.what();
+            if (!what) {
+                what = "*nothing*";
```
maxgolov, line 87: Love it! ❤️

lib/offline/OfflineStorage_Room.cpp
```diff
@@ -330,121 +336,141 @@ namespace MAT_NS_BEGIN
...
+                                                                             reservedUntil_id));
+                ThrowLogic(env, "get reservedUntil");
+                auto blob_java = static_cast<jbyteArray>(env->GetObjectField(record, blob_id));
+                ThrowLogic(env, "get blob");
+                uint8_t *start = reinterpret_cast<uint8_t *>(env->GetByteArrayElements(blob_java,
+                                                                                       nullptr));
+                ThrowLogic(env, "get blob storage");
+                uint8_t *end = start + env->GetArrayLength(blob_java);
+                StorageRecord dest(
+                        std::to_string(id_java),
```
maxgolov, line 440: Isn't this a temporary, and/or would this be memcpy'ied into `StorageRecord` ?

maxgolov (approved):
Looks good. Only Q: I assume that the constructor is doing the string copy, and we're not trying to refer to string buffer of a temporary?

lib/offline/OfflineStorage_Room.cpp
```diff
@@ -82,9 +76,17 @@ namespace MAT_NS_BEGIN
...
     {
-        if (env->PushLocalFrame(frameSize) == 0)
-        {
-            ++push_count;
+        try {
+            if (env->PushLocalFrame(frameSize) == JNI_OK) {
+                ++push_count;
+            }
+        }
+        catch (std::exception e) {
```
mkoscumb, line 84: Catch by const&, C++17 mitigates this with Guaranteed Copy Elision, which allows copy free catch by value. As the SDK only requires C++11, we should adhere to best practices for that language version.

mkoscumb: Applies to all catch blocks :)

mkoscumb (approved):

lib/offline/OfflineStorage_Room.cpp
mkoscumb: Also, since std::exception is a base class, we may get into some slicing problems if we catch by value here :)

PR #760, opened 2021-01-15
[WIP] ILogConfiguration in Obj-C layer should remain in scope for the lifecycle of its LogManager

maxgolov:
ILogConfiguration in Obj-C layer should remain in scope for the lifecycle of its LogManager. Speculative fix.


eduardo-camacho (approved):

kindbe (approved):

EduardoVaca (approved):

sid-dahiya (approved):
I locally edited the file on my iOS build and _nothing bad happened._ Unfortunately, I don't have a repro for the crash itself, so I cannot definitively state that this fixed the crash, but it doesn't seem to cause a regression. Based on that, I'm signing off.

PR #763, opened 2021-01-16
Add support MacOS arm64

vgerasimovms:


build.sh
```diff
@@ -91,14 +91,14 @@ if [ "$LINK_TYPE" == "shared" ]; then
 fi
 
 # Set target MacOS minver
-export MACOSX_DEPLOYMENT_TARGET=10.10
+export MACOSX_DEPLOYMENT_TARGET=11.10
```
maxgolov, line 94: I would like others to chime in if it is OK to upgrade that one. I do realize that for arm64 silicon you need the new tools / SDK.

maxgolov (approved):
Please get approval from either @kindbe or @eduardo-camacho

build.sh
vgerasimovms: I made adjustments: 11.1 just for arm64

PR #765, opened 2021-01-20
Java: add public access modifier to EventPropertyXXXValue classes

theta3:
So these classes are exposed to users for casting EventProperty.getEventPropertyValue()


maxgolov (approved):
Would be good if you add some write-up or usage example... Maybe a Unit Test?

PR #766, opened 2021-01-20
Suppress Clang inconsistent-missing-override warnings around GMock Macros

mkoscumb:
`MOCK_METHOD` macros and their ilk appear to not append the `override` keyword to their generated method declarations. As such I'm disabling Clang's[ inconsistent-missing-override](https://clang.llvm.org/docs/DiagnosticsReference.html#winconsistent-missing-override) warning around only those types.

I also cleaned up the vcproj, let me know if I missed anything and I'm happy to undo that.

### DIFF
--- a/tests/common/MockIEcsClient.hpp
+++ b/tests/common/MockIEcsClient.hpp
@@ -8,6 +8,10 @@
 
 namespace testing {
 
+#if defined(__clang__)
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Winconsistent-missing-override"  // GMock MOCK_METHOD* macros don't use override.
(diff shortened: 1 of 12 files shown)


maxgolov (approved):
Looks good.

Indirectly related to this PR: our `.filters` files in general are a little bit messed up at the moment. They show all files as flat list right now (at least for the main SDK libraries). I found a plugin that allows to reconstruct the filters - visual layout of project in Visual Studio IDE - to match the actual directory layout, i.e. tree view rather than a flat list. I can submit a separate no-code-changes PR to fix this consistently for all filters, so don't worry if your PR breaks the visual layout.

PR #768, opened 2021-01-23
Build is broken if compiling without modules

maxgolov:
Motivation for this PR:

- make sure that the entire set of tests for all platforms is functional and passing even if compiling without private modules.

- remove private submodules cloning from GitHub Actions. Azure DevOps / VSO workflow may still be able to clone all if build account is explicitly authorized (for local runs in your own Docker). But we are not longer using PAT token for the clone action here.

While testing these changes I also did a full build in Visual Studio to cover additional extra projects that are not presently covered by CI, e.g. `ClientTelemetry3s.lib` - super-slim static build. It turned out that this slim build got broken some time ago due to Offline Storage factory refactor in `OfflineStorageFactory.cpp`. I fixed it too. We need to add slim build to CI later on, but right now I verified that it works with an ad-hoc test.

### DIFF
--- a/.github/workflows/build-android.yml
+++ b/.github/workflows/build-android.yml
@@ -19,6 +19,7 @@ jobs:
       uses: actions/checkout@v2
       with:
         submodules: 'true'
+      continue-on-error: true
     - name: Chocolatey
       run: |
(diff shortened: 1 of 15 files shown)


larvacea (approved):
Thanks!

PR #750, opened 2020-12-16
Add missing methods and properties to obj-c wrapper with corresponding Xamarin bindings

bemartin:
Changes in this PR are:

- Added obj-c method to set the cache file path
- Added obj-c methods for flush and flushAndTearDown that return a Status for consistency with Android/Java 
- Added Xamarin bindings for added methods
- Added code in Xamarin bindings to unify the APIs between the Obj-C and Java implementations

PR #770, opened 2021-01-25
Adding return value statements  in flush and flushAndTearDown methods of ODWLogManager.mm

dipatid:
Description:

While updating the pod spec with version 3.5.19, the local build verification failed with below error:

`- ERROR | [iOS] xcodebuild:  OneDsCppSdk.iOS/wrappers/obj-c/ODWLogManager.mm:207:1: error: non-void function does not return a value in all control paths [-Werror,-Wreturn-type]`

When I looked at recent changes I found [a recent change](https://github.com/microsoft/cpp_client_telemetry/pull/750/files#diff-e120c7a402598bfdf01df679f15c6d3d87dc63c9a59354d7fd0a2ec188040b94R197) in which we do not return an instance of ODWStatus in these two methods in `ODWLogManager.mm`
```
+(ODWStatus)flush
+(ODWStatus)flushAndTeardown 
```

Solution: 

1. Return ODWEfail when exception occurs in these two methods.
2. Return ODWStatus in value returned from `LogManager::FlushAndTeardown()` in `flushAndTeardown`

### DIFF
--- a/wrappers/obj-c/ODWLogManager.mm
+++ b/wrappers/obj-c/ODWLogManager.mm
@@ -187,6 +187,7 @@ static BOOL _initialized = false;
             [ODWLogger raiseException: e.what()];
         }
         [ODWLogger traceException: e.what()];
+        return ODWEfail;
     }
 }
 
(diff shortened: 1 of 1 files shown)


maxgolov (approved):
Looks good, assuming you tested it all compiles with warnings enabled.
And there are no similar instances elsewhere.

maxgolov:
@dipatid - one minor ask for the next PR: please create your branches on the main repo to avoid forking. You should have the proper permissions to create a branch, then send in a PR. We intentionally do not run CI for PRs that were sent in from forks.

PR #772, opened 2021-01-26
Move string methods from Utils.hpp to StringUtils.hpp

mkoscumb:
Moving the string functions to StringUtils seems like a better place for these methods to live. While doing so I moved the method bodies to the .cpp file. The combination of the two fixes here allows for us to change the underlying implementation of the method bodies without invalidating much of the build DAG, which means faster incremental builds for everyone.

I will remove the `#include "utils/StringUtils.hpp"` from Utils.hpp, but I need that there now to make fixes to the private modules repo. That'll be coming in subsequent PRs.

Additional thoughts moving forward, neither of these are part of this PR in order to keep it scoped. If people have strong opinions I'm happy to integrate:
1. I'm not sure of the value of the StringUtils static class. As it didn't have any state, I moved it to a namespace. I'd like to know if people have strong opinions on whether the other methods in StringUtils.hpp should be in the StringUtils namespace.
2. The code style guide indicates that public methods are `PascalCased` and other methods are `camelCased`. I'd like to change all the StringUtils.hpp methods to match `PascalCase`, but that seemed like it'd distract from the central point of the PR (moving methods).


lib/api/ContextFieldsProvider.cpp
```diff
@@ -7,7 +7,8 @@
 #include "ContextFieldsProvider.hpp"
 #include "LogSessionData.hpp"
 
-#include "utils/Utils.hpp"
+#include "utils/StringUtils.hpp"
+// #include "utils/Utils.hpp"
```
mkoscumb, line 11: Whoops!

lib/system/EventProperty.cpp
```diff
@@ -4,7 +4,7 @@
 //
 #include "EventProperty.hpp"
 
-#include "utils/Utils.hpp"
+#include "utils/StringUtils.hpp"
```
mkoscumb, line 7: From the loops, this looks like this needs `utils/annex_k.h` as well, I'll add it.

larvacea (approved):
I hate the preprocessor. #ifdef max #undef max indeed.

mkoscumb:
> I hate the preprocessor. #ifdef max #undef max indeed.

Yeah, I'm going to `#define NOMINMAX` in the Windows build scripts.

PR #773, opened 2021-01-26
Speculative change: crashes on low-DRAM, older Android API devices

larvacea:


lib/offline/OfflineStorage_Room.cpp
```diff
@@ -336,6 +336,8 @@ namespace MAT_NS_BEGIN
         EventLatency minLatency,
         unsigned maxCount)
     {
+        constexpr int64_t chunkSize = 1024;
+        int64_t requested = maxCount ? maxCount : INT64_MAX;
```
mkoscumb, line 340: `std::numeric_limits<int64_t>::max()` rather than `INT64_MAX`

mkoscumb (approved):

PR #776, opened 2021-01-28
Fixed documentation warnings in  ObjC layer which cause XCode build to fail

dipatid:
Fixed documentation warnings in  ObjC layer which cause XCode build to fail when "Documentation Comments" build setting is "YES" for a build target.

### DIFF
--- a/wrappers/obj-c/ODWDiagnosticDataViewer.h
+++ b/wrappers/obj-c/ODWDiagnosticDataViewer.h
@@ -23,7 +23,7 @@ NS_ASSUME_NONNULL_BEGIN
 /*!
  @brief Enables Data Viewer.
  @param endpoint A string that contains endpoint to route events.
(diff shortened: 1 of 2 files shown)


mkoscumb (approved):
LGTM, but I'd also get EduardoC's review.

maxgolov (approved):

maxgolov:
Is this complete set? Are there any other issues that trigger documentation warnings? is there anything in the private modules that needs the same clean-up? If so, you need to clean it up in the modules, then push a zero-change commit id PR to upgrade the submodule to refer to latest.

eduardo-camacho (approved):

dipatid:
@maxgolov , I did not find any other [-WDocumentattion] at any other place while building the code. So, this PR should address current set of documentation warnings.

Issue #775, opened 2021-01-27
Version.hpp Shouldn't Define MAT* and PAL* Namespace Macros


mkoscumb:
**Problem**
Version.hpp does two things right now:
1. `#define` the MAT* and PAL* Namespace macros.
2. Initialize the `Version` constant value and generated strings.

As a result of 1 above, Version.hpp is `#include`d in nearly - if not all - files in the SDK. This means that any change to Version.hpp will invalidate any previous builds of anything which includes it. This can be frustrating for developers, as Version.hpp will be updated on their local device every day as part of running the build scripts, or can be updated by `git pull`ing in an updated copy of Version.hpp as part of a release. Thus leaving work one night and starting again the next morning may force a complete rebuild of the SDK, where only a couple files may have changed.

**Proposed Solution**
Move `#define`s for the MAT* and PAL* Namespace macros to another file (possibly CTMacros.h, possibly a new file). This will allow faster incremental builds for developers, as Version.hpp updates will only require rebuilds of files which **use** the Version constant or strings (rather than the entire SDK).

mkoscumb:
Resoved with PR 779

PR #779, opened 2021-02-01
Move MAT* and PAL* namespace macros out of version.hpp

mkoscumb:
The underlying problem is better explained in this issue: https://github.com/microsoft/cpp_client_telemetry/issues/775

I moved the MAT* and PAL* namespace macros to CTMacros.hpp, as it seemed to be the most reasonable place for their addition, though I wouldn't be opposed to moving them to a separate file (NamespaceMacros.hpp perhaps?).

This ostensibly should improve incremental builds after running `git pull`, after a day change, or anything which updates the Version number, as only files which directly depend on the Version.hpp headers should be invalidated by those operations.


kindbe (approved):
LGTM

sid-dahiya (approved):
LGTM

maxgolov (approved):

PR #783, opened 2021-02-16
Add Support for Multiple IDataInspectors in LogManager

bwhybel:
Also adds support for LiveEventInspector module

### DIFF
--- a/lib/api/LogManagerImpl.cpp
+++ b/lib/api/LogManagerImpl.cpp
@@ -538,12 +538,12 @@ namespace MAT_NS_BEGIN
         m_context.SetCustomField(name, prop);
         {
             LOCKGUARD(m_dataInspectorGuard);
-            auto it = std::find_if(m_dataInspectors.begin(), m_dataInspectors.end(), [](std::shared_ptr<IDataInspector> inspector){
-                return inspector->InspectorDescriptor() == "PrivacyGuard";
-            });
-            if (it != m_dataInspectors.end())
+            for(const auto& inspector : m_dataInspectors)
             {
-                (*it)->InspectSemanticContext(name, value, /*isGlobalContext: */ true, std::string{});
+                if(inspector->InspectorDescriptor().compare("PrivacyGuard") == 0)
+                {
+                    inspector->InspectSemanticContext(name, value, /*isGlobalContext: */ true, std::string{});
+                }
             }
         }
         return STATUS_SUCCESS;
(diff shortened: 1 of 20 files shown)


lib/api/LogManagerImpl.cpp
```diff
@@ -525,9 +538,12 @@ namespace MAT_NS_BEGIN
         m_context.SetCustomField(name, prop);
         {
             LOCKGUARD(m_dataInspectorGuard);
-            if (m_dataInspector)
+            for(const auto& inspector : m_dataInspectors)
             {
-                m_dataInspector->InspectSemanticContext(name, value, /*isGlobalContext: */ true, std::string{});
+                if(inspector->InspectorDescriptor().compare("PrivacyGuard") == 0)
```
sid-dahiya, line 543: I don't believe we should have this check. All `IDataInspector`'s should have an `InspectSemanticContext` method and they should decide if they want to ignore it or inspect it. The logic should be an implementation details. This check is leaking the implementation detail into other components.

lib/api/LogManagerImpl.cpp
```diff
@@ -804,15 +820,37 @@ namespace MAT_NS_BEGIN
...
-    void LogManagerImpl::SetDataInspector(const std::shared_ptr<IDataInspector>& dataInspector)
+    void LogManagerImpl::AddOrUpdateInspector(const std::shared_ptr<IDataInspector>& dataInspector)
     {
         LOCKGUARD(m_dataInspectorGuard);
-        m_dataInspector = dataInspector;
+        if(dataInspector == nullptr)
+        {
+            m_dataInspectors.clear();
+            return;
+        }
```
sid-dahiya, line 830: This seems weird. Someone passes a nullptr and just clears everything?

The original design made sense for a single inspector, but I don't believe this makes sense for clearing all anymore. We probably should have a separate `ClearDataInspectors` or some method like that.

lib/api/LogManagerImpl.cpp
```diff
@@ -804,15 +820,37 @@ namespace MAT_NS_BEGIN
...
+
+        auto itInspector = std::find_if(m_dataInspectors.begin(), m_dataInspectors.end(), [&dataInspector](const auto& inspector){
+            return inspector->InspectorDescriptor().compare(dataInspector->InspectorDescriptor()) == 0;
+        });
+
+        if (itInspector != m_dataInspectors.end())
+        {
+            m_dataInspectors.erase(itInspector);
```
sid-dahiya, line 838: Probably should add a `LOG_INFO` or something that we're replacing the dataInspector here. Or maybe add a boolean flag to denote if we want to replace if it exists.

lib/include/public/ILogManager.hpp
```diff
@@ -377,13 +377,14 @@ namespace MAT_NS_BEGIN
...
         /// <param name="dataInspector">Shared Ptr to an instance of IDataInspector</param>
-        virtual void SetDataInspector(const std::shared_ptr<IDataInspector>& dataInspector) = 0;
+        virtual void AddOrUpdateInspector(const std::shared_ptr<IDataInspector>& dataInspector) = 0;
 
         /// <summary>
         /// Get the current instance of IDataInspector
         /// </summary>
-        /// <returns>Current instance of IDataInspector if set, nullptr otherwise.</returns>
-        virtual std::shared_ptr<IDataInspector> GetDataInspector() noexcept = 0;
+        /// <param name="inspectorDescriptor">String that describes the type of inspector</param>
```
sid-dahiya, line 385: `String that identifies the inspector`

lib/include/public/IDataInspector.hpp
```diff
@@ -119,6 +119,12 @@ namespace MAT_NS_BEGIN
         /// <param name="isGlobalContext">Whether this is a global/logmanager Context or local ILogger context</param>
         /// <param name="associatedTenant">(Optional) Tenant associated with the Context</param>
         virtual void InspectSemanticContext(const std::string& contextName, GUID_t contextValue, bool isGlobalContext, const std::string& associatedTenant) noexcept = 0;
+
+        /// <summary>
+        /// Get the descriptive name for inspector.
+        /// </summary>
+        /// <returns>String descriptor for instance of IDataInspector.</returns>
+        virtual const std::string& InspectorDescriptor() const noexcept = 0;
```
sid-dahiya, line 127: Description usually reads as you're giving details about the inspector. This is more of a Unique Identifier, probably should call this `UniqueIdentifier` or just `GetName`.

sid-dahiya (changes requested):
You need to finish the modules change before this one, but with the failing runs there, this is feeling like a chicken and the egg problem. I'll defer to @maxgolov about what is the best step forward. Also note that before you finish this change, you need to complete the modules change and make sure the submodules in this change point to the merge commit (or later commit on master branch) as otherwise it'll break the builds here.

lib/api/LogManagerImpl.cpp
```diff
@@ -804,15 +820,37 @@ namespace MAT_NS_BEGIN
...
     {
         LOCKGUARD(m_dataInspectorGuard);
-        m_dataInspector = dataInspector;
+        if(dataInspector == nullptr)
+        {
+            m_dataInspectors.clear();
+            return;
+        }
+
+        auto itInspector = std::find_if(m_dataInspectors.begin(), m_dataInspectors.end(), [&dataInspector](const auto& inspector){
```
kindbe, line 832: "There can only be one of each kind and I find/replace it based on some key" sounds more like a map than a vector.

bwhybel: Are there any performance differences to be considered that would favor `vector`? Because the `m_dataInspectors` vector will always be quite small

lib/api/LogManagerImpl.cpp
bwhybel: yes I added that for both, but only removed it for one call of the `InspectSemanticContext`, will definitely just make the call and if it isn't necessary that call will be a no-op determined by that implementation of `IDataInspector`

lib/api/LogManagerImpl.cpp
bwhybel: This was just taken from the documentation: "Note: A `nullptr` may be passed to `SetInspector` if the intent is to remove any registered data inspectors."

Adding a ClearDataInspectors makes more sense and having addorupdate just do nothing when a nullptr is passed in

lib/include/public/IDataInspector.hpp
bwhybel: that completely makes sense, changed everywhere

lib/api/LogManagerImpl.cpp
bwhybel: What would the situation be where we wouldn't want to update? When I'm thinking through it, in making a call to AddOrUpdate the desired behavior would be for the instance that you passed into that call to be present in that LogManager. I prefer the additional `LOG_INFO` to adding a boolean parameter

PR #785, opened 2021-02-19
Expose LogConfiguration.getLogConfigurationCopy() and LogManager.loggerWithTenant() with custom config on iOS

### LINKED ISSUE #784: Expose `LogManager.getLogConfigurationCopy()` and `LogManagerProvider.createLogManager(logConfiguration)` as custom instances on iOS
On iOS, I would like to have `ODWLogger` instances that target custom endpoint side by side with default endpoint. On Android, it's possible to do this:
``````
ILogConfiguration logConfiguration = LogManager.getLogConfigurationCopy();
// call set()
ILogManager logManager = LogManagerProvider.createLogManager(logConfiguration);
ILogger logger = logManager.getLogger(tenant);
``````
Is it possible to expose these methods on iOS?

I'm not familiar with the codebase and obj-c and the Android implementation looks like a lot. So I'm looking for some guidance here

### PR DESCRIPTION
**Pull request by theta3**
Currently iOS only uses default config, which makes side-by-side loggers with different configuration not possible. close #784

### DIFF
--- a/wrappers/obj-c/ODWLogConfiguration.h
+++ b/wrappers/obj-c/ODWLogConfiguration.h
@@ -312,114 +312,114 @@ extern NSString * _Nonnull const ODWCFG_BOOL_SESSION_RESET_ENABLED;
 @interface ODWLogConfiguration : NSObject
 
 /*!
- @brief Return a copy of the default configuration
+@brief Return a copy of the default configuration
 */
 +(nullable ODWLogConfiguration *)getLogConfigurationCopy;
 
 /*!
- @brief Sets the URI of the event collector.
- @param eventCollectorUri A string for event collector uri.
+@brief Sets the URI of the event collector.
+@param eventCollectorUri A string for event collector uri.
 */
 +(void)setEventCollectorUri:(nonnull NSString *)eventCollectorUri;
 
 /*!
- @brief Returns the URI of the event collector.
+@brief Returns the URI of the event collector.
 */
 +(nullable NSString *)eventCollectorUri;
 
 /*!
- @brief Sets the RAM queue size limit in bytes.
- @param cacheMemorySizeLimitInBytes  A long value for memory size limit in bytes.
+@brief Sets the RAM queue size limit in bytes.
+@param cacheMemorySizeLimitInBytes  A long value for memory size limit in bytes.
 */
 +(void)setCacheMemorySizeLimitInBytes:(uint64_t)cacheMemorySizeLimitInBytes;
 
 /*!
- @brief return the RAM queue size limit in bytes.
+@brief return the RAM queue size limit in bytes.
 */
 +(uint64_t)cacheMemorySizeLimitInBytes;
 
 /*!
- @brief Sets the size limit of the disk file used to cache events on the client side.
- @param cacheFileSizeLimitInBytes  A long value for cache file size limit.
+@brief Sets the size limit of the disk file used to cache events on the client side.
+@param cacheFileSizeLimitInBytes  A long value for cache file size limit.
 */
 +(void)setCacheFileSizeLimitInBytes:(uint64_t)cacheFileSizeLimitInBytes;
 
 /*!
- @brief Returns the size limit of the disk file used to cache events on the client side.
+@brief Returns the size limit of the disk file used to cache events on the client side.
 */
 +(uint64_t)cacheFileSizeLimitInBytes;
 
 /*!
- @brief Sets max teardown upload time in seconds.
- @param maxTeardownUploadTimeInSec An integer that time in seconds.
+@brief Sets max teardown upload time in seconds.
+@param maxTeardownUploadTimeInSec An integer that time in seconds.
 */
 +(void)setMaxTeardownUploadTimeInSec:(int)maxTeardownUploadTimeInSec;
 
 /*!
- @brief Sets if trace logging to file is enabled.
- @param enableTrace True if tracing is enabled.
+@brief Sets if trace logging to file is enabled.
+@param enableTrace True if tracing is enabled.
 */
 +(void)setEnableTrace:(bool)enableTrace;
 
 /*!
- @brief Sets if console logging from the iOS wrapper is enabled.
- @param enableConsoleLogging True if logging is enabled.
+@brief Sets if console logging from the iOS wrapper is enabled.
+@param enableConsoleLogging True if logging is enabled.
 */
 +(void)setEnableConsoleLogging:(bool)enableConsoleLogging;
 
 /*!
- @brief Sets the internal SDK debugging trace level.
- @param TraceLevel one of the ACTTraceLevel values.
+@brief Sets the internal SDK debugging trace level.
+@param TraceLevel one of the ACTTraceLevel values.
 */
 +(void)setTraceLevel:(int)TraceLevel;
 
 /*!
- @brief Returns true if tracing is enabled.
+@brief Returns true if tracing is enabled.
 */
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: wrappers/obj-c/ODWLogConfiguration.h [hunk 1]
```diff
@@ -4,111 +4,425 @@
...
+ @brief Returns the cache file path.
 */
 +(nullable NSString *)cacheFilePath;
 
+/*!
+ @brief Sets a config key to a string value
+ @param key A key.
+ @param value A value.
+*/
+-(void)set:(nonnull NSString *)key withValue:(nonnull NSString *)value;
```
eduardo-camacho, line 424: This method has misleading behavior. It updates a new internal instance defined at ODWLogConfiguration, while other methods like setCacheMemorySizeLimitInBytes update directly instance tracked by LogManager::GetLogConfiguration(), not the internal instance. This leads to a race condition.

wrappers/obj-c/ODWLogConfiguration.mm
```diff
@@ -4,16 +4,341 @@
...
+#import "ODWLogConfiguration_private.h"
 #import "ODWLogger_private.h"
 #import "LogManager.hpp"
 
 using namespace Microsoft::Applications::Events;
 
+/*!
+  Enable analytics.
+*/
+NSString *const ODWCFG_BOOL_ENABLE_ANALYTICS = @"enableLifecycleSession";
```
eduardo-camacho, line 16: Why did you not define these at ODWLogConfiguration.h directly?

wrappers/obj-c/ODWLogConfiguration.h
```diff
@@ -4,111 +4,425 @@
...
 
 /*!
-@brief Returns true if tracing is enabled.
+ @brief Returns true if tracing is enabled.
 */
 +(bool)enableTrace;
 
 /*!
-@brief Returns true if console logging is enabled.
+ @brief Returns true if console logging is enabled.
```
eduardo-camacho, line 382: nit: please revert extra space on all these comments

eduardo-camacho (changes requested):
Left few comments

wrappers/obj-c/ODWLogConfiguration.h
theta3: Why is this misleading? `_wrappedConfiguration` is set in private init, which is called through the static method `getLogConfigurationCopy`, which creates a *copy* of the default configuration. This instance method modified wrappedConfiguration which is a copy of the default configuration. No race condition here. Calling static methods modifies default config. Calling instance methods modified copied config. This is the same behavior as Android

wrappers/obj-c/ODWLogConfiguration.mm
theta3: Isn't that how public static vars are defined in objc? https://stackoverflow.com/a/23653584/4092412 https://stackoverflow.com/a/3117562/4092412 https://stackoverflow.com/a/16331068/4092412

wrappers/obj-c/ODWLogConfiguration.h
eduardo-camacho: Spaces are still there, same comment applies to other files in this PR. I know this might not have on execution, but let's keep change size smaller

theta3: Hi, you are looking at an OUTDATED commit, please take a look at latest commit https://github.com/microsoft/cpp_client_telemetry/pull/785/files

wrappers/obj-c/ODWLogManager.mm
```diff
@@ -52,6 +55,47 @@ +(nullable ODWLogger *)loggerWithTenant:(nonnull NSString *)tenantToken
     return [[ODWLogger alloc] initWithILogger: logger];
 }
 
++(nullable ODWLogger *)loggerWithTenant:(nonnull NSString *)tenantToken
```
eduardo-camacho, line 58: This method  is short-circuiting LogManager initialization logic (look at initializeLogManager method), resulting not only on code duplication but also skiping Log::ManagerInitialize, which is required. Try to create an implementation that reuses existing initialization logic as much as possible

thomasameisel: The issue is we do not want to update the static functions here, and it looks like most/all of the code in the `initializeLogManager` is using the static `LogManager` functions

thomasameisel: @theta3 I noticed there is an existing `initForTenant:withConfig:` [function](https://github.com/microsoft/cpp_client_telemetry/blob/master/wrappers/obj-c/ODWLogManager.mm#L55) here. Would we be able to use that instead of this new function?

eduardo-camacho: According to your scenario you are getting a second logger, so I would assume consumer called an existing loggerWithTenant method already, and you should at minimum for _initialized and fail otherwise. If my assumption is incorrect then initialization should not be skipped, because also things like turning off statistics, and setting semantics for AppVersion, AppLanguage and UserLanguage happens here

eduardo-camacho: Answering your question on "I noticed there is an existing... ", I think calling this it would override existing default configuration, which it's not what you are looking for

hruizhang: That's a good point @eduardo-camacho.  That said, I'm not sure I see the reason why there needs to be a static instance of the configuration at all.  It seems to make sense to pass the configuration into the LogManager, and have the LogManager instance be associated 1-to-1 with the specified configuration.   Would you mind clarifying the intent there?

eduardo-camacho (approved):

wrappers/obj-c/ODWLogConfiguration_private.h
```diff
@@ -0,0 +1,33 @@
...
+//
+#include "objc_begin.h"
+#include "LogManager.hpp"
+#import "ODWLogConfiguration.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+using namespace MAT;
+
+#define EXCEPTION_TRACE_FORMAT @"C++ Exception: %s"
```
kindbe, line 13: Is this #define necessary? I don't see any usages of it.

wrappers/obj-c/ODWLogConfiguration.mm
```diff
@@ -4,16 +4,341 @@
...
+
+/*!
+ Parameter that allows to check if the SDK is running on UTC mode
+*/
+NSString *const ODWCFG_BOOL_UTC_ENABLED = @"enabled";
+
+/*!
+ Parameter that allows to check if the SDK is running on UTC mode
+*/
+NSString *const ODWCFG_BOOL_UTC_ACTIVE = @"active";
```
kindbe, line 61: UTC options don't make sense on non-Windows platforms

wrappers/obj-c/ODWLogConfiguration.mm
```diff
@@ -4,16 +4,341 @@
...
+
+/*!
+ Pointer to the Android app's JavaVM
+*/
+NSString *const ODWCFG_PTR_ANDROID_JVM = @"android_jvm";
+
+/*!
+ JObject of the Android app's main activity
+*/
+NSString *const ODWCFG_JOBJECT_ANDROID_ACTIVITY = @"android_activity";
```
kindbe, line 218: Android options also don't make much sense to port to obj-c

kindbe (approved):
Approved with suggestions

PR #786, opened 2021-02-19
Add Support for Mulitple IDataInspectors within Log Manager

bwhybel:
the follow up changes will need to be to either the `IDataInspector` to have a public function that identifies the type of DataInspector is it (such as `UniqueIdentifier()` which will always describe the type of identifier) or we could change the ILogManager interface to include identifiers when adding new IDataInspectors (such as change `SetDataInspector(std::shared_ptr<IDataInspector>)` to `AddDataInspector(std::string, std::shared_ptr<IDataInspector>)`) or we could use a call to `typeid(currentInspector).name()` and compare that to the currently present `IDataInspector`'s however I'm not confident that using `typeid` wouldn't cause any issues down the road. This last solution I believe would reduce the number public interface changes

### DIFF
--- a/lib/api/LogManagerImpl.cpp
+++ b/lib/api/LogManagerImpl.cpp
@@ -807,7 +807,6 @@ namespace MAT_NS_BEGIN
     void LogManagerImpl::SetDataInspector(const std::shared_ptr<IDataInspector>& dataInspector)
     {
         LOCKGUARD(m_dataInspectorGuard);
-        m_dataInspectors.clear();
         if(dataInspector == nullptr)
         {
             LOG_WARN("Attempting to set nullptr as DataInspector");
@@ -816,7 +815,7 @@ namespace MAT_NS_BEGIN
 
         auto itDataInspector = std::find_if(m_dataInspectors.begin(), m_dataInspectors.end(), [&dataInspector](const std::shared_ptr<IDataInspector>& currentInspector)
         {
-            return strcmp(typeid(dataInspector).name(),typeid(currentInspector).name()) == 0;
+            return strcmp(typeid(*dataInspector).name(),typeid(*currentInspector).name()) == 0;
         });
 
         if (itDataInspector != m_dataInspectors.end())
@@ -847,11 +846,12 @@ namespace MAT_NS_BEGIN
     {
         for (const auto& dataInspector : m_dataInspectors)
         {
-            if (strcmp(uniqueIdentifier.c_str(), typeid(dataInspector).name()) == 0)
+            if (strcmp(uniqueIdentifier.c_str(), typeid(*dataInspector).name()) == 0)
             {
                 return dataInspector;
             }
         }
+        LOG_WARN("DataInspector requested does not exist");
         return  nullptr;
     }
 
--- a/lib/api/LogManagerImpl.cpp
+++ b/lib/api/LogManagerImpl.cpp
@@ -830,30 +830,30 @@ namespace MAT_NS_BEGIN
     void LogManagerImpl::ClearDataInspectors()
     {
         LOCKGUARD(m_dataInspectorGuard);
-        m_dataInspectors.clear();
+        std::vector<std::shared_ptr<IDataInspector>>().swap(m_dataInspectors);
     }
 
-    std::shared_ptr<IDataInspector> LogManagerImpl::GetDataInspector() noexcept
+    void LogManagerImpl::RemoveDataInspector(const std::string& name)
     {
-        if(m_dataInspectors.empty())
+        LOCKGUARD(m_dataInspectorGuard);
+        auto it = std::find_if(m_dataInspectors.begin(), m_dataInspectors.end(), [&name](const std::shared_ptr<IDataInspector>& inspector){
+            return strcmp(inspector->GetName(), name.c_str()) == 0;
+        });
+
+        if (it != m_dataInspectors.end())
         {
-            return nullptr;
+            m_dataInspectors.erase(it);
         }
-        return m_dataInspectors.at(0);
     }
 
-    std::shared_ptr<IDataInspector> LogManagerImpl::GetDataInspector(const std::string& uniqueIdentifier) noexcept
+    std::shared_ptr<IDataInspector> LogManagerImpl::GetDataInspector(const std::string& name) noexcept
     {
-        for (const auto& dataInspector : m_dataInspectors)
-        {
-            std::string inspectorIdentifier(typeid(*dataInspector).name());
-            if (inspectorIdentifier.find(uniqueIdentifier) != std::string::npos)
-            {
-                return dataInspector;
-            }
-        }
-        LOG_WARN("DataInspector requested does not exist");
-        return  nullptr;
+        LOCKGUARD(m_dataInspectorGuard);
+        auto it = std::find_if(m_dataInspectors.begin(), m_dataInspectors.end(), [&name](const std::shared_ptr<IDataInspector>& inspector){
+            return strcmp(inspector->GetName(), name.c_str()) == 0;
+        });
+
+        return it != m_dataInspectors.end() ? *it : nullptr;
     }
 
     status_t LogManagerImpl::DeleteData()
--- a/lib/api/LogManagerImpl.cpp
+++ b/lib/api/LogManagerImpl.cpp
@@ -402,7 +402,7 @@ namespace MAT_NS_BEGIN
             m_httpClient = nullptr;
             m_taskDispatcher = nullptr;
             m_dataViewer = nullptr;
-            std::vector<std::shared_ptr<IDataInspector>>{}.swap(m_dataInspectors);
+            ClearDataInspectors();
 
             m_filters.UnregisterAllFilters();
 
--- a/lib/api/LogManagerImpl.cpp
+++ b/lib/api/LogManagerImpl.cpp
@@ -846,7 +846,8 @@ namespace MAT_NS_BEGIN
     {
         for (const auto& dataInspector : m_dataInspectors)
         {
-            if (strcmp(uniqueIdentifier.c_str(), typeid(*dataInspector).name()) == 0)
+            std::string inspectorIdentifier(typeid(*dataInspector).name());
+            if (inspectorIdentifier.find(uniqueIdentifier) != std::string::npos)
             {
                 return dataInspector;
             }
--- a/lib/api/LogManagerImpl.cpp
+++ b/lib/api/LogManagerImpl.cpp
@@ -816,27 +816,6 @@ namespace MAT_NS_BEGIN
         m_dataInspectors.push_back(dataInspector);
     }
 
-    void LogManagerImpl::AddOrUpdateDataInspector(const std::shared_ptr<IDataInspector>& dataInspector)
-    {
-        LOCKGUARD(m_dataInspectorGuard);
-        if(dataInspector == nullptr)
-        {
-            LOG_WARN("Attempting to set nullptr as DataInspector");
-            return;
-        }
-
-        auto itDataInspector = std::find_if(m_dataInspectors.begin(), m_dataInspectors.end(), [&dataInspector](const auto& inspector){
-            return inspector->UniqueIdentifier().compare(dataInspector->UniqueIdentifier()) == 0;
-        });
-
-        if (itDataInspector != m_dataInspectors.end())
-        {
-            m_dataInspectors.erase(itDataInspector);
-        }
-
-        m_dataInspectors.push_back(dataInspector);
-    }
-
     void LogManagerImpl::ClearDataInspectors()
     {
         LOCKGUARD(m_dataInspectorGuard);
@@ -850,18 +829,6 @@ namespace MAT_NS_BEGIN
             return nullptr;
         }
         return m_dataInspectors.at(0);
-    }
-
-    std::shared_ptr<IDataInspector> LogManagerImpl::GetDataInspector(const std::string& uniqueIdentifier) noexcept
-    {
-        for (const auto& dataInspector : m_dataInspectors)
-        {
-            if (dataInspector->UniqueIdentifier().compare(uniqueIdentifier) == 0)
-            {
-                return dataInspector;
-            }
-        }
-        return nullptr;
     }
 
     status_t LogManagerImpl::DeleteData()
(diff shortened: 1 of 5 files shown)


lib/include/public/ILogManager.hpp
```diff
@@ -380,10 +380,22 @@ namespace MAT_NS_BEGIN
...
         /// <summary>
-        /// Get the current instance of IDataInspector
+        /// Clears all IDataInspectors
+        /// </summary>
+        virtual void ClearDataInspectors() = 0;
+
+        /// <summary>
+        /// Get the first instance of IDataInspector
         /// </summary>
         /// <returns>Current instance of IDataInspector if set, nullptr otherwise.</returns>
```
kindbe, line 390: "returns" description needs updating

lib/api/LogManagerImpl.hpp
```diff
@@ -297,8 +297,9 @@ namespace MAT_NS_BEGIN
         static size_t GetDeadLoggerCount();
 
         virtual void SetDataInspector(const std::shared_ptr<IDataInspector>& dataInspector) override;
-
+        virtual void ClearDataInspectors() override;
         virtual std::shared_ptr<IDataInspector> GetDataInspector() noexcept override;
+        virtual std::shared_ptr<IDataInspector> GetDataInspector(const std::string& uniqueIdentifier) noexcept override;
```
sid-dahiya, line 302: Since we can get specific `IDataInspector` based on the identifier, I'd suggest allowing for removing specific inspector based on that as well.

lib/api/LogManagerImpl.cpp
```diff
@@ -807,12 +807,53 @@ namespace MAT_NS_BEGIN
...
     {
-        return m_dataInspector;
+        if(m_dataInspectors.empty())
+        {
+            return nullptr;
+        }
+        return m_dataInspectors.at(0);
+    }
+
+    std::shared_ptr<IDataInspector> LogManagerImpl::GetDataInspector(const std::string& uniqueIdentifier) noexcept
```
sid-dahiya, line 845: We should call `uniqueIdentifier` as `name` as that's the parameter we're comparing with.

lib/api/LogManagerImpl.cpp
```diff
@@ -807,12 +807,53 @@ namespace MAT_NS_BEGIN
...
-        m_dataInspector = dataInspector;
+        if(dataInspector == nullptr)
+        {
+            LOG_WARN("Attempting to set nullptr as DataInspector");
+            return;
+        }
+
+        auto itDataInspector = std::find_if(m_dataInspectors.begin(), m_dataInspectors.end(), [&dataInspector](const std::shared_ptr<IDataInspector>& currentInspector)
+        {
+            return strcmp(typeid(*dataInspector).name(),typeid(*currentInspector).name()) == 0;
```
kindbe, line 818: Do we need to strcmp the typeid.name()s to compare the types?

lib/api/LogManagerImpl.cpp
```diff
@@ -807,12 +807,53 @@ namespace MAT_NS_BEGIN
...
+            return nullptr;
+        }
+        return m_dataInspectors.at(0);
+    }
+
+    std::shared_ptr<IDataInspector> LogManagerImpl::GetDataInspector(const std::string& uniqueIdentifier) noexcept
+    {
+        for (const auto& dataInspector : m_dataInspectors)
+        {
+            std::string inspectorIdentifier(typeid(*dataInspector).name());
```
sid-dahiya, line 849: I'm a bit curious about this approach. Instead of iterating over all inspectors, why can't we do something like:
```cpp
auto identifiedInspector = std::find_if(m_dataInspectors.begin(), m_dataInspectors.end(), [&uniqueIdentifier]
{
   return strcmp(inspector->GetName(), uniqueIdentifier) == 0;
});

return identifiedInspector != m_dataInspectors.end() ? identifiedInspector  : nullptr;
```

sid-dahiya (approved):
Left suggestions for improvements, but not worth blocking this. Still recommend addressing them, if possible.

kindbe (changes requested):

lib/api/LogManagerImpl.cpp
```diff
@@ -402,7 +402,7 @@ namespace MAT_NS_BEGIN
             m_httpClient = nullptr;
             m_taskDispatcher = nullptr;
             m_dataViewer = nullptr;
-            m_dataInspector = nullptr;
+            m_dataInspectors.clear();
```
mkoscumb, line 405: Clear doesn't guarantee that the underlying buffer is released. Prefer std::vector<T>()::swap(m_dataInspectors), as this guarantees that the underlying buffer is released.
See: https://www.cplusplus.com/reference/vector/vector/clear/

lib/api/LogManagerImpl.cpp
mkoscumb, line 818: Strongly discourage RTTI. Many consumers (including Office) disable this functionality intentionally as it bloats the binary. Further, it's not clear to me what you're trying to do here? Guarantee that no more than one instance of the same type of IDataInspector is registered? I can imagine multiple instances of the same type trying to send data to different endpoints, which should be supported.

lib/api/LogManagerImpl.cpp
mkoscumb: Actually, shouldn't this just call ClearDataInspectors()?

lib/api/LogManagerImpl.cpp
```diff
@@ -807,12 +807,53 @@ namespace MAT_NS_BEGIN
...
+            m_dataInspectors.erase(itDataInspector);
+        }
+
+        m_dataInspectors.push_back(dataInspector);
+    }
+
+    void LogManagerImpl::ClearDataInspectors()
+    {
+        LOCKGUARD(m_dataInspectorGuard);
+        m_dataInspectors.clear();
```
mkoscumb, line 833: Clear doesn't guarantee that the underlying buffer is released. Prefer std::vector()::swap(m_dataInspectors), as this guarantees that the underlying buffer is released.
See: https://www.cplusplus.com/reference/vector/vector/clear/

lib/api/LogManagerImpl.cpp
mkoscumb: Consider uniqueifying by name, otherwise don't bother?

lib/api/LogManagerImpl.cpp
mkoscumb, line 849: Same as above, I'm not sure that uniqueifying by type is the correct behavior.

lib/api/LogManagerImpl.cpp
```diff
@@ -807,12 +807,53 @@ namespace MAT_NS_BEGIN
...
     }
 
     std::shared_ptr<IDataInspector> LogManagerImpl::GetDataInspector() noexcept
     {
-        return m_dataInspector;
+        if(m_dataInspectors.empty())
+        {
+            return nullptr;
+        }
+        return m_dataInspectors.at(0);
```
mkoscumb, line 842: Use a lockguard here.

lib/api/LogManagerImpl.cpp
```diff
@@ -807,12 +807,53 @@ namespace MAT_NS_BEGIN
...
+    }
+
+    std::shared_ptr<IDataInspector> LogManagerImpl::GetDataInspector(const std::string& uniqueIdentifier) noexcept
+    {
+        for (const auto& dataInspector : m_dataInspectors)
```
mkoscumb, line 847: Lockguard here

mkoscumb (changes requested):

lib/api/LogManagerImpl.cpp
```diff
@@ -807,12 +807,53 @@ namespace MAT_NS_BEGIN
...
+        m_dataInspectors.push_back(dataInspector);
+    }
+
+    void LogManagerImpl::ClearDataInspectors()
+    {
+        LOCKGUARD(m_dataInspectorGuard);
+        m_dataInspectors.clear();
     }
 
     std::shared_ptr<IDataInspector> LogManagerImpl::GetDataInspector() noexcept
```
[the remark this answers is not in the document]
bwhybel: yes this will eventually be deleted, the problem is that there are tests in the private repo that uses `GetDataInspector` without specifying which one, so if I change the `ILogManager` by removing the `GetDataInspector()` then the private repo will fail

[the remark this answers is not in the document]
kindbe: Ah, makes sense

lib/api/LogManagerImpl.cpp
bwhybel: done

lib/api/LogManagerImpl.cpp
bwhybel: deleted code because the implementation of `GetDataInspector()` we put into the interface (for now, until the proper implementation is put into the modules)

lib/api/LogManagerImpl.cpp
bwhybel: typeid removed, using default GetName() (return "", non pure virtual)

lib/api/LogManagerImpl.cpp
bwhybel: done

lib/api/LogManagerImpl.cpp
bwhybel: changed to uniqueify by name

lib/api/LogManagerImpl.cpp
bwhybel: changed to call `ClearDataInspectors`

lib/api/LogManagerImpl.cpp
bwhybel: comparing by `GetName()` now

lib/include/public/ILogManager.hpp
bwhybel: updated

lib/api/LogManagerImpl.cpp
bwhybel: updated

lib/api/LogManagerImpl.cpp
bwhybel: changed

lib/api/LogManagerImpl.hpp
bwhybel: added

lib/api/LogManagerImpl.cpp
```diff
@@ -807,12 +807,53 @@ namespace MAT_NS_BEGIN
...
+            m_dataInspectors.erase(itDataInspector);
+        }
+
+        m_dataInspectors.push_back(dataInspector);
+    }
+
+    void LogManagerImpl::ClearDataInspectors()
+    {
+        LOCKGUARD(m_dataInspectorGuard);
+        std::vector<std::shared_ptr<IDataInspector>>{}.swap(m_dataInspectors);
```
sid-dahiya, line 833: This is backwards.
`m_dataInspectors.swap({});`

lib/include/public/IDataInspector.hpp
```diff
@@ -118,6 +118,12 @@ namespace MAT_NS_BEGIN
         /// <param name="isGlobalContext">Whether this is a global/logmanager Context or local ILogger context</param>
         /// <param name="associatedTenant">(Optional) Tenant associated with the Context</param>
         virtual void InspectSemanticContext(const std::string& contextName, GUID_t contextValue, bool isGlobalContext, const std::string& associatedTenant) noexcept = 0;
+
+        /// <summary>
+        /// Returns unique name for current Data Inspector
+        /// </summary>
+        /// <returns>Name of Data Inspector</returns>
+        virtual const char* GetName() const noexcept { return ""; }
```
sid-dahiya, line 126: Why are you returning a default empty string? For the interfaces, this should be set to `= 0` so the implementing class implements the method correctly.

lib/include/public/ILogManager.hpp
```diff
@@ -380,10 +380,28 @@ namespace MAT_NS_BEGIN
...
+        /// </summary>
+        /// <param name="name">String name that identifies IDataInspector</param>
+        /// <returns>Selected instance of IDataInspector if available, nullptr otherwise.</returns>
+        virtual std::shared_ptr<IDataInspector> GetDataInspector(const std::string& name) noexcept = 0;
+
+        /// <summary>
+        /// Get the current instance of IDataInspector specified by an empty string
+        /// </summary>
+        /// <returns>Selected instance of IDataInspector if available, nullptr otherwise.</returns>
+        virtual std::shared_ptr<IDataInspector> GetDataInspector() noexcept { return GetDataInspector(std::string{}); }
```
sid-dahiya, line 404: Same here, don't return defaults in interfaces.

sid-dahiya (changes requested):

lib/api/LogManagerImpl.cpp
sid-dahiya: http://www.cplusplus.com/reference/vector/vector/swap-free/

lib/include/public/IDataInspector.hpp
bwhybel: this was a suggestion from Matt because we need to keep PrivacyGuard building for this first PR, this will then get turned into a pure virtual function on the next update after we can fix the modules

sid-dahiya: Ah, doh, yes. This is fine then. I'm waiting on the swap FWIW.

lib/include/public/ILogManager.hpp
bwhybel: this is the same situation, this call will be deleted once the modules change goes in, so is it ok to leave it here?

lib/api/LogManagerImpl.cpp
bwhybel: Is it? Looking at the clear documentation it shows that `x` gets cleared and reallocated doing it the opposite way https://www.cplusplus.com/reference/vector/vector/clear/

bwhybel: there is also this: https://github.com/microsoft/cpp_client_telemetry/blob/125818fcf7637b956f0454420d3c7be87b58ad80/lib/filter/EventFilterCollection.cpp#L44
current example in the code

sid-dahiya: I believe using swap to clear is what threw me off. Yeah, your code works: https://godbolt.org/z/vh1Gxq

sid-dahiya (approved):

mkoscumb (approved):
LGTM, thanks!

kindbe (approved):

PR #790, opened 2021-02-25
Mitigate OOM errors on Android

larvacea:
We are seeing OOM crashes on the oldest and smallest (DRAM) devices.
If we OOM in the database or upload code, we should simply do nothing. We
may succeed on a subsequent attempt, but in any event we do not want to
crash the application because telemetry failed to upload.

### DIFF
--- a/lib/http/HttpClient_Android.cpp
+++ b/lib/http/HttpClient_Android.cpp
@@ -4,505 +4,569 @@
 //
 #include "HttpClient_Android.hpp"
 #include <algorithm>
-#include <jni.h>
 #include <cstdio>
+#include <jni.h>
 #include <sstream>
 #include <vector>
 
-namespace MAT_NS_BEGIN {
-
-constexpr static auto Tag = "HttpClient_Android";
-
-HttpClient_Android::HttpRequest::~HttpRequest() noexcept
+namespace MAT_NS_BEGIN
 {
-	EraseFromParent();
-	if (m_java_request)
-	{
-		JNIEnv* env = nullptr;
-		if (HttpClient_Android::s_java_vm->AttachCurrentThread(&env, nullptr) != JNI_OK)
-		{
-			return;
-		}
-		env->DeleteGlobalRef(m_java_request);
-	}
+    constexpr static auto Tag = "HttpClient_Android";
+
+    HttpClient_Android::HttpRequest::~HttpRequest() noexcept
+    {
+        EraseFromParent();
+        if (m_java_request)
+        {
+            JNIEnv* env = nullptr;
+            if (HttpClient_Android::s_java_vm->AttachCurrentThread(&env, nullptr) != JNI_OK)
+            {
+                return;
+            }
+            env->DeleteGlobalRef(m_java_request);
+        }
+    }
+
+    void HttpClient_Android::HttpRequest::SetMethod(std::string const& method)
+    {
+        m_method = method;
+    }
+
+    void HttpClient_Android::HttpRequest::SetUrl(std::string const& url)
+    {
+        m_url = url;
+    }
+
+    HttpHeaders& HttpClient_Android::HttpRequest::GetHeaders()
+    {
+        return m_headers;
+    }
+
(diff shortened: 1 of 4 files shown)


larvacea:
My apologies for the enormous whitespace diffs; I ran the formatter on this code to make it comply with the project standard, and, well, it really didn't comply before, apparently.

suryatn:
@larvacea please help in getting this merged to the repo post which we can use the updated SDK to resolve the crash issue.

lib/offline/OfflineStorage_Room.cpp
```diff
@@ -70,52 +70,57 @@ namespace MAT_NS_BEGIN
...
-            auto what = e.what();
-            if (!what) {
-                what = "*nothing*";
+        if (env->ExceptionCheck() == JNI_TRUE)
+        {
+            env->ExceptionDescribe();
+            env->ExceptionClear();
+            if (s_throwExceptions)
+            {
+                MATSDK_THROW(std::runtime_error("Push Local Frame"));
```
suryatn, line 89: please check if this will lead to a crash

suryatn:
@maxgolov @sid-dahiya @mkoscumb please help in reviewing the PR and getting the it merged to master.

lib/offline/OfflineStorage_Room.cpp
sid-dahiya: The exception, if thrown, is caught on line 306.

lib/offline/OfflineStorage_Room.cpp
```diff
@@ -684,227 +748,300 @@ namespace MAT_NS_BEGIN
...
+        }
+        catch (const std::runtime_error& error)
+        {
+            auto what = error.what();
+            if (!what)
             {
-                ResizeDbInternal(env);
+                what = "*nothing*";
             }
+            LOG_ERROR("Exception in StoreRecords: %s", what);
```
sid-dahiya, line 875: Is this sufficient information for debugging the exceptions? We will know `what` the exception is, but we won't know where it came from given the large scope of the try-block.

sid-dahiya: Applies globally in this file.

---
In reply to: [583813385](https://github.com/microsoft/cpp_client_telemetry/pull/790#discussion_r583813385) [](ancestors = 583813385)

sid-dahiya (approved):
:shipit:

maxgolov (approved):

lib/offline/OfflineStorageFactory.cpp
```diff
@@ -4,7 +4,6 @@
 //
 #include "mat/config.h"
 
-#pragma once
```
mkoscumb, line 7: Big thank.

mkoscumb (approved):
Seems reasonable, though I figure someone needs to do a whitespace cleanup PR at some point so some of these diffs are more reasonable.

lib/offline/OfflineStorage_Room.cpp
larvacea: In automation, we can see device logs, and this should be enough information to tell us where things went horribly wrong. In production, we won't see this at all because the catch block will do what catch blocks are supposed to do, and the exception will be a non-event.

lib/http/HttpClient_Android.cpp
```diff
@@ -4,505 +4,569 @@
...
+    IHttpRequest* HttpClient_Android::CreateRequest()
+    {
+        HttpRequest* local_request(new HttpRequest(*this));
+        std::lock_guard<std::mutex> lock(m_requestsMutex);
+        m_requests.push_back(std::move(local_request));
+        return local_request;
+    }
+
+    // we do NOT own the callback object
+    // we will segfault if callback is nullptr
```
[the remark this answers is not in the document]
larvacea: My theory is that callback == nullptr would represent a logic error in the calling code. I would rather let hardware deal with the joys of this exceptional condition, even if it means that we get a naked SIGSEGV. Debugging SIGSEGV-because-address-zero is not rocket science. Messing with branch prediction to avoid SIGSEGV is a performance hit.

PR #792, opened 2021-02-26
Re-enable SQLite tests and fix init/shutdown for multiple log managers

bliptec:
Previously with multiple log managers, all log managers needed skipSqliteInitAndShutdown, and SQLite needed to be initialized independently. Otherwise, even with careful sequencing of creating and shutting down log managers, operations like `recreate` could call shutdown, improperly tearing down SQLite while other log managers are active.

This ensures that `sqlite3_shutdown` isn't called if there are any other SDK instances of SQLite in the same memory space. I don't expect this to cause any issues, because shutting down SQLite is generally unnecessary. Added a related test, and re-enabled all the other tests 🙂

This change also fixes an issue where `MockIRuntimeConfig` could persist some state between tests.


jevery23 (approved):

lib/offline/OfflineStorage_SQLite.cpp
```diff
@@ -17,6 +17,9 @@ namespace MAT_NS_BEGIN {
 
     constexpr static size_t kBlockSize = 8192;
 
+    std::mutex OfflineStorage_SQLite::m_initAndShutdownLock;
+    int OfflineStorage_SQLite::m_instanceCount = 0;
```
maxgolov, line 21: Should we make it `atomic` ?

lib/offline/SQLiteWrapper.hpp
```diff
@@ -204,18 +204,34 @@ namespace MAT_NS_BEGIN {
...
         {
-            int result;
+            int result = SQLITE_OK;
 
             if (!m_skipInitAndShutdown) {
-                result = g_sqlite3Proxy->sqlite3_initialize();
+                if (m_initAndShutdownLock && m_instanceCount)
+                {
+                    LOCKGUARD(*m_initAndShutdownLock);
+                    *m_instanceCount = *m_instanceCount > 0 ? *m_instanceCount + 1 : 1;
```
maxgolov, line 225: nit: this construct looks unusual, at least to my eye 😄

lib/offline/SQLiteWrapper.hpp
```diff
@@ -273,6 +285,25 @@ namespace MAT_NS_BEGIN {
...
 
+        void shutdown_sqlite()
+        {
+            if (!m_skipInitAndShutdown)
+            {
+                if (m_initAndShutdownLock && m_instanceCount)
+                {
+                    LOCKGUARD(*m_initAndShutdownLock);
+                    *m_instanceCount -= 1;
+                    if (*m_instanceCount <= 0)
```
maxgolov, line 296: It feels a bit odd to see this check for `less than zero`. What are the preconditions for that? `FlushAndTeardown` should be rather safe, I believe it's already ref-counting properly.

maxgolov (approved):
I am generally Okay with this. But there are some interesting moments that I would like to get the answers for..

lib/offline/SQLiteWrapper.hpp
maxgolov: Why exactly do you first decrement it unconditionally, and then check if it turned out to be less than zero?

lib/offline/OfflineStorage_SQLite.cpp
bliptec: There's no need to, since there's already a lock, which is needed to prevent another initialization happening during shutdown (since shutdwon isn't thread safe for sqlite).

lib/offline/SQLiteWrapper.hpp
bliptec: I agree it wasn't very clear, updated the logic and added a check to not increment the count unless the initialization succeeds.

lib/offline/SQLiteWrapper.hpp
bliptec: Good catch. I think I originally did this out of a defensive habit, just in case the value isn't what was expected, but I updated this logic to be a bit clearer, and I changed it so shutdown only gets called when the count is exactly 1 (since the safer thing is to not shutdown multiple times)

maxgolov (approved):
I'm fine with this. It would really help if we can get a second pair of eyes on this.

@mkoscumb @larvacea @kindbe

lib/offline/OfflineStorage_SQLite.cpp
larvacea, line 21: Nit: it's conventional (but really not necessary) to make reference counts either size_t or uint64_t. Perhaps because we all remember the trauma (no, we don't ALL remember it) from designs where someone said "oh, we don't need that sign bit," or in the case of IBM's System 360 and 370, we don't need the high byte, who could ever need more than a 24 bit address. Don't ask me, I just work here. I believe that the probability that the SDK will need more than 2 billion SQLite-consuming-things is vanishingly small, so I am absolutely ok with the signed instance count here.

One can also make this work with std::shared_ptr, but I think your design is more straightforward.

larvacea (approved):

lib/offline/OfflineStorage_SQLite.cpp
bliptec: Good point. In this case I was slightly more worried about accidental underflows, but I'm not expecting either (underflow or 2 bil LogManager) to happen.

maxgolov: What is ironic - my understanding that based on Google Coding style, in some codebases they prefer `int` over `size_t` for that exact reason... 🤷 the underflow....

PR #797, opened 2021-03-01
Enable Warning as Error on GCC and Clang builds for symmetry with MSVC.

mkoscumb:
~~For now I'll be disabling any warnings which aren't trivially fixable.~~
Ok, I just fixed them or suppressed them all. Yay, no warnings!


lib/offline/OfflineStorage_SQLite.cpp
```diff
@@ -755,8 +755,17 @@ namespace MAT_NS_BEGIN {
             if (!stmt.select() || !stmt.getRow(m_pageSize)) { return false; }
         }
 
+#if defined(_MSC_VER)
 #pragma warning(push)
 #pragma warning(disable:4296) // expression always false.
```
larvacea, line 760: Why does the code want to compare uintxx_t x < 0? Just curious why this is happening.

larvacea (approved):

lib/offline/OfflineStorage_SQLite.cpp
mkoscumb: No idea, seems like it's unhittable code, but not something I want to address in this PR.

PR #798, opened 2021-03-01
Capturing ConnectivityManager exceptions in Android

kivarsha:
Calls into ConnectivityManager can sometimes throw exceptions, which causes crashes on HttpClient initialization. According to some posts online, some of these exceptions, such as a SecurityException due to package permissions (might also be relevant to Android Work Profile management), and RuntimeExceptions when too many callbacks are registered, are currently not handled in the Logger code. 

Capturing these calls in a try/catch loop ensures that we can still continue with the proper logger initialization, even if we don't have full insights into the changes in the network state of the device.

Similar issue is reported here: https://github.com/DataDog/dd-sdk-android/issues/413

### DIFF
--- a/lib/android_build/maesdk/src/main/java/com/microsoft/applications/events/HttpClient.java
+++ b/lib/android_build/maesdk/src/main/java/com/microsoft/applications/events/HttpClient.java
@@ -210,14 +210,23 @@ public class HttpClient {
     if (hasConnectivityManager()) {
       if (context.checkSelfPermission(permission.ACCESS_NETWORK_STATE)
           == PackageManager.PERMISSION_GRANTED) {
-        m_connectivityManager =
-            (ConnectivityManager) context.getSystemService(Context.CONNECTIVITY_SERVICE);
-        if (m_connectivityManager != null) {
-          boolean is_metered = m_connectivityManager.isActiveNetworkMetered();
-          m_callback = new ConnectivityCallback(this, is_metered);
-          onCostChange(is_metered); // set initial value in C++ side
-          m_connectivityManager.registerDefaultNetworkCallback(m_callback);
-        }
+            try {
+              m_connectivityManager =
+                  (ConnectivityManager) context.getSystemService(Context.CONNECTIVITY_SERVICE);
+              if (m_connectivityManager != null) {
+                boolean is_metered = m_connectivityManager.isActiveNetworkMetered();
+                m_callback = new ConnectivityCallback(this, is_metered);
+                onCostChange(is_metered); // set initial value in C++ side
+                m_connectivityManager.registerDefaultNetworkCallback(m_callback);
+              }
+            }
+            catch (Exception e) {
+              // Fetching CONNECTIVITY_SERVICE can throw a SecurityException, especially in Android Work Profile cases
+              // "package does not belong to xxxx"
+              // can also throw runtimeException: 
+              // https://developer.android.com/reference/android/net/ConnectivityManager#registerDefaultNetworkCallback(android.net.ConnectivityManager.NetworkCallback)
+              // in either case, we don't have access to ConnectivityInfo, so can't truly populate callback/isMetered, or react to network changes
+            }
       }
     }
     m_power_receiver = new PowerInfoReceiver(this);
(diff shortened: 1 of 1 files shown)


lib/android_build/maesdk/src/main/java/com/microsoft/applications/events/HttpClient.java
```diff
@@ -210,14 +210,23 @@ public HttpClient(Context context) {
...
+                  (ConnectivityManager) context.getSystemService(Context.CONNECTIVITY_SERVICE);
+              if (m_connectivityManager != null) {
+                boolean is_metered = m_connectivityManager.isActiveNetworkMetered();
+                m_callback = new ConnectivityCallback(this, is_metered);
+                onCostChange(is_metered); // set initial value in C++ side
+                m_connectivityManager.registerDefaultNetworkCallback(m_callback);
+              }
+            }
+            catch (Exception e) {
+              // Fetching CONNECTIVITY_SERVICE can throw a SecurityException, especially in Android Work Profile cases
```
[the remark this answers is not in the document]
kivarsha: LogCat is definitely doable. From what I see in the Android layer, it doesn't log anything itself, but rather, once the call makes it into the C++ maesdk components, those log to both logcat (if enabled) and mat*.log. The C++ debug logging is also controlled by a configuration setting, I fear straight logcatting from the android layer might spam logcat as the client cannot control its output.

And as the main logger depends on HttpClient, it seems cyclical to call into the debug logging on the C++ side from here....so not sure what to do/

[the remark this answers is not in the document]
theta3: Maybe @larvacea can chime in?

lib/android_build/maesdk/src/main/java/com/microsoft/applications/events/HttpClient.java
```diff
@@ -210,14 +210,23 @@ public HttpClient(Context context) {
...
+              m_connectivityManager =
+                  (ConnectivityManager) context.getSystemService(Context.CONNECTIVITY_SERVICE);
+              if (m_connectivityManager != null) {
+                boolean is_metered = m_connectivityManager.isActiveNetworkMetered();
+                m_callback = new ConnectivityCallback(this, is_metered);
+                onCostChange(is_metered); // set initial value in C++ side
+                m_connectivityManager.registerDefaultNetworkCallback(m_callback);
+              }
+            }
+            catch (Exception e) {
```
mkoscumb, line 223: Should we explicitly catch the below known exceptions, rather than generically catching and throwing away everything?

maxgolov: I think catching generic one is very good: that secures us from anything bad happening. But maybe we need to log that somewhere, in logcat.

maxgolov (approved):
I like it as-is, or at least - maybe we can have a follow-up PR that somehow logs the exception to logcat (in a separate PR, in my opinion).

kivarsha:
@maxgolov, the latest iteration calls into Log.e, which should log into logcat, but, unfortunately, that's not a log that can be turned off from the consumer side like we have with logcat logs from the C++ layer

lib/android_build/maesdk/src/main/java/com/microsoft/applications/events/HttpClient.java
kivarsha: I split out the different exceptions in case we want to handle them separately in the future + added logs for all of them in the latest iteration.

maxgolov:
> @maxgolov, the latest iteration calls into Log.e, which should log into logcat, but, unfortunately, that's not a log that can be turned off from the consumer side like we have with logcat logs from the C++ layer

Then let's not log anything at this point? to avoid polluting the logs with wasteful exceptions.. And consider some better / smarter debugging alternative in a follow-up PR. Stability & User Experience - P1. Debuggability - P2. I assume the issue itself happens on a permission-hardened device, where users kinda explicitly restrict / revoke the necessary permissions... i.e. not wide-spread.

PR #800, opened 2021-03-04
Prospective fix for crash

thomasameisel:
Prospective fix for crash when calling `ODWLogConfiguration.set`


eduardo-camacho (approved):

wrappers/obj-c/ODWLogConfiguration.mm
```diff
@@ -306,7 +306,7 @@ -(nullable ILogConfiguration*)getWrappedConfiguration
 +(nullable ODWLogConfiguration *)getLogConfigurationCopy
 {
     auto& config = LogManager::GetLogConfiguration();
-    ILogConfiguration configCopy(config);
+    static ILogConfiguration configCopy(config);
```
larvacea, line 309: I am guessing here because I don't have an ObjC reference on my phone (ironic, what was Apple thinking). I don't know for certain how static, the prior call to getlogconfiguration, and the initializer interact. I believe based on faith and hope that this will work to give you what you 
need: a copy or copies that will have static lifetime so log manager won't dereference garbage.

larvacea (approved):

PR #803, opened 2021-03-04
Move setting initialized to inner function

thomasameisel:
Right now we only set the initialized property when the `init` function is called. This property should be set in the inner function which is called from other functions as well.


eduardo-camacho (approved):

EduardoVaca (approved):

larvacea (approved):
lgtm

PR #806, opened 2021-03-16
Upgrade sqlite to v3.34.1

maxgolov:
One of our customers asked to upgrade to latest approved version of sqlite that is being used by other Microsoft products, e.g. same version that ships in latest Windows 10 (winsqlite3.dll). In their initial email they suggested **3.34.1** . These are the release notes:

https://www.sqlite.org/news.html
**SQLite version 3.34.1 is a patch releases that fixes a possible use-after-free bug that can be provoked by malicious SQL. Other minor issues in extensions and documentation are also fixed.**

There is a more recent version published last week ( **3.35.x** ), with additional new features. But my understanding that the latest hasn't got yet as many flight-hours as the one published earlier in January. We also don't need / use the new features added in 3.35.

The process as follows:
- obtain SQLite (recommended version 3.34.1) from https://www.sqlite.org/tmp/sqlite-win10-all-3340100.zip
- extract on top of the current version
- run tests
- commit


maxgolov:
There's one sketchy moment here:
https://github.com/microsoft/cpp_client_telemetry/pull/806/commits/3070a5ece565e5cb5a15bd8a7ddd156ea77eddf2

The official snapshot is obviously compatible with vs2019 compiler (used for building Win 10 SDK). But is not well-received by slightly older vs2017. We have to support both. Well, one cannot really do

```cpp
#if SQLITE_4_BYTE_ALIGNED_MALLOC
```
because `SQLITE_4_BYTE_ALIGNED_MALLOC` is not a number.

So the fix is:
```cpp
#ifdef SQLITE_4_BYTE_ALIGNED_MALLOC
``
instead of `if`

maxgolov:
Android runner is not feeling so good today (unrelated to my PR though).

larvacea (approved):

mkoscumb (approved):

PR #807, opened 2021-03-16
Fix PayloadDecoder broken for win32-mini-lib flavor

maxgolov:
Due to series of unfortunate events, one of the recent commits in the optional modules is now taking dependency on PayloadDecoder.cpp ... This is great, and usually it would work nicely, as in most build configurations we link both `zlib` and `json.hpp` libraries.

However, there's a build flavor of SKU (designed for installer.exe scenario), which we are yet to add to full build loop ( `win32-mini-lib` and corresponding `SampleCppMini` ), that is implementing a mini lib in under 350KB static linked, that does not include no zlib nor json for that matter. Thus, if you do the full build in MSBuild using Visual Studio IDE - you'd notice that those build flavors got broken.. sigh.. Because now we try to pull in a module, that is taking a dependency on something that is never gonna be available in that SKU. So I'm fixing those two again by stubbing out the missing feature.

After my change everything compiles and works again. I'll consider adding a build test for the `mini` flavor in a separate PR. Adding another set of functional / unit for that mini SKU won't scale..

Thanks to @utsabsingharoy for finding this issue.


maxgolov:
( Android CI is busted , unrelated to my changes :-/ )

utsabsingharoy (approved):

lalitb (approved):

mkoscumb (approved):

PR #809, opened 2021-03-18
Targeted fix to enable the arm64 static library build

rybest:
The current build doesn't support building Windows arm64 debug and retail libraries. This change fixes the solutions to enable the Windows arm64 builds that depend upon sqlite. This is a targeted fix and doesn't try to resolve any issues with other non-sqlite configurations.


trevorlacey-msft (approved):

Solutions/MSTelemetrySDK.sln
```diff
@@ -130,8 +130,6 @@ Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "SampleCppLogManagers", "..\
 EndProject
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "SampleCppMini", "..\examples\cpp\SampleCppMini\SampleCppMini.vcxproj", "{86AC752C-5687-4377-841E-943D9BEEF361}"
 EndProject
-Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "SampleCppUTC", "..\lib\modules\examples\cpp\SampleCppUTC\SampleCppUTC.vcxproj", "{C947B185-2B00-4073-A19D-483ED5C7EDC2}"
```
maxgolov, line 133: Why is this change needed?

build-win.ps1
```diff
@@ -76,7 +76,7 @@ foreach ($arch in $archs) {
         echo ""
         continue
       }
-      if ($binType -eq "lib" -and ($actualArch -eq "ARM" -or $actualArch -eq "ARM64")) {
```
maxgolov, line 79: If this is the only essential change necessary, can the other changes in .sln file all be safely reverted?

maxgolov:
> In addition to the targeted fix, this change removes the SampleCppUTC.vcxproj from the solution
> since it doesn't appear to be included in the repository anymore and can give build errors.

It's included if you build with additional private modules. You can specify the list of dependencies you want to build explicitly, as it's done in this script, for example:

https://github.com/microsoft/cpp_client_telemetry/blob/9aa9a8dd9cdda146a43b13874de6de3927d1cfd2/build-win10-min.bat#L32

where `/target:` option specifies exactly what project you need to build, while `/p:BuildProjectReferences=true` option specifies that all dependencies (references) of the project are also built, thus you build only the projects that are needed by MIP SDK.

Solutions/MSTelemetrySDK.sln
maxgolov: Since your `powershell` script already does:

```powershell
& cmd /c "msbuild $solution /target:$targetStr /p:BuildProjectReferences=true /maxcpucount:$cpuCount /p:Configuration=$actualConfig /p:Platform=$actualArch $actualCustomProps"
```

i.e. selective build of only the libraries you need via `/target:$targetStr` option, it seems to me that the changes made to Solution file here have no effect. You should be able to safely revert the .sln change. These are needed for customers who have extra modules under `lib/modules/` path.

rybest: I was getting errors when I tried to build the solution file (couldn't find the project). If SampleCppUTC is needed by other customers, I'll revert that part of the change and just ensure it doesn't get scheduled for the arm64 lib build.

build-win.ps1
rybest: No - the other solution changes for arm64 are also needed. The only optional one was to remove the references to SampleCppUTC.

Solutions/MSTelemetrySDK.sln
maxgolov: Sounds good.

maxgolov: You can also make sure that you can recursively clone with "private" module, that way your snapshot will include the UTC sample app too. It'd be best if we retain it in the solution.

maxgolov (approved):
Looks good to me. Thank you! Please merge.

Issue #781, opened 2021-02-09
Support Windows x64/x86/arm64 vcpkg for 1DS C++ SDK (mstelemetry)


lulululululu:
Before opening a feature request against this repo, consider whether the feature is generic enough for the SDK, or it should be a specific plugin/extension to the SDK.

**Is your feature request related to a problem?**
Currently 1DS C++ SDK only supports to build with Mac and Linux using vcpkg. 

**Describe the solution you'd like.**
What do you want to happen instead? What is the expected behavior?
Add support to build 1DS C++ SDK for Windows x64/x86/arm64 using vcpkg.

**Describe alternatives you've considered.**
Which alternative solutions or features have you considered?

**Additional context.**
Add any other context about the feature request here.
We are trying to add 1DS C++ SDK to our project as a dependency. We are targeting builds on Windows x64/x86/arm64.

maxgolov:
This is partially supported, but we have not yet fully tested and not yet pushed the control file to `vcpkg` repo.

Please try out some instructions here. These were tested on Linux:https://github.com/microsoft/cpp_client_telemetry/blob/master/docs/building-with-vcpkg.md

lulululululu:
Hi Max,
I tested on Windows by Cloning the repo and ran `vcpkg install --head --overlay-ports=`pwd`/tools/ports mstelemetry`. Got BUILD_FAILED. Checking log files, it's caused by `tsrc/sqlite3.h(35): fatal error C1083: Cannot open include file: 'stdarg.h': No such file or directory [C:\Repos\cpp_client_telemetry\sqlite\sqlite.vcxproj]`.

Looks like portfile.cmake is just calling `build-all.bat`. So I tried running this script myself and got the same error. Looking into this script, seems like it's trying to build four flavors: /MD, /MT, ARM, ARM64. I temporarily disabled /MD, ARM, ARM64 and build succeeded.

So looks like it's missing some logic to detect the build machine architecture? Also is it missing a step to copy the build output to a destination folder?

maxgolov:
The script right now builds once for all architectures, it does not intake the `vcpkg`-provided architcture.

[...]

PR #810, opened 2021-03-19
Updating mstelemetry portfile for windows

anaaru:
Got clienttelemetry.lib to build with MSVC v142 platform toolset.


tools/ports/mstelemetry/portfile.cmake
```diff
@@ -4,45 +4,80 @@ message("CMAKE_CURRENT_SOURCE_DIR=${CMAKE_CURRENT_SOURCE_DIR}")
...
-file(INSTALL ${SOURCE_PATH}/LICENSE DESTINATION ${CURRENT_PACKAGES_DIR}/share/${PORT} RENAME copyright)
+    vcpkg_install_msbuild(
+        SOURCE_PATH ${SOURCE_PATH}
+        PROJECT_SUBPATH Solutions/MSTelemetrySDK.sln
+        LICENSE_SUBPATH LICENSE
+        RELEASE_CONFIGURATION Release
+        DEBUG_CONFIGURATION Debug
+        OPTIONS /p:MATSDK_SHARED_LIB=1
+        PLATFORM ${VCPKG_TARGET_ARCHITECTURE}
+        PLATFORM_TOOLSET v142
```
maxgolov, line 73: I think we can add two options here:
- one is to respect env variable `PLATFORM_TOOLSET` , if it is already set (for those who still use vs2017-vc141)
- another option is to auto-detect the compiler. It's a bit more involved. But I can help with a script for that.

In general, maybe just replacing it with a custom variable logic - if set, use the value from var; else if not set, then assume the tool chain is v142 - would be good enough at first.

tools/ports/mstelemetry/portfile.cmake
```diff
@@ -4,45 +4,80 @@ message("CMAKE_CURRENT_SOURCE_DIR=${CMAKE_CURRENT_SOURCE_DIR}")
 message("CMAKE_MODULE_PATH=${CMAKE_MODULE_PATH}")
 message("CMAKE_CURRENT_LIST_DIR=${CMAKE_CURRENT_LIST_DIR}")
 
-execute_process(COMMAND "${CMAKE_CURRENT_LIST_DIR}/get_repo_name.sh" OUTPUT_VARIABLE REPO_NAME ERROR_QUIET)
+if (UNIX)
+    execute_process(COMMAND "${CMAKE_CURRENT_LIST_DIR}/get_repo_name.sh" OUTPUT_VARIABLE REPO_NAME ERROR_QUIET)
+else()
+    # execute_process(COMMAND git config --get remote.origin.url OUTPUT_VARIABLE REPO_URL ERROR_QUIET)
```
maxgolov, line 10: I can add a batch file `get_repo_name.cmd` here, but we also then need to assume that `git.exe` is also in the path.

maxgolov:
@anaaru - I think it's generally looking good. Let's merge it, and I'll add a few improvements in a separate PR, add you as reviewer.

maxgolov (approved):

maxgolov:
@anaaru - can you send in a separate follow-up PR that adds a bit of documentation to this file ?
https://github.com/microsoft/cpp_client_telemetry/blob/master/docs/building-with-vcpkg.md
It _should_ work, but maybe there are a few quirks in the process. I think I originally tested it on Linux only.

maxgolov:
@lulululululu - I think this PR should partially address what you asked for in #781

mkoscumb (approved):

Issue #16, opened 2019-06-18
Setup core SDK build agent that is similar to Edge gn setup


maxgolov:
Currently 1DS SDK is not built with the same set of compiler dependencies and warning levels as MS Edge. It would help if we can clean up OneDS warnings & Integrate a CI target with a very similar build setup to that of MS Edge (Windows clang-cl, Mac/Linux clang with strict warnings), with continuous run of BVT (unit and functional). That way all potential breaking issues can be caught early-on prior integrating a new SDK source code drop in MS Edge.

maxgolov:
I'm reopening it, but with slightly adjusted goal in mind:
- setup CMake build
- with clang
- that uses usual `ninja` to build the SDK

This should be more or less 95% functionally similar to `gn` build.

maxgolov:
Closing due to inactivity. Feel free to reopen. No plans to fix. Best solution would be to set up the usual cmake+ninja build (not gn).

PR #811, opened 2021-03-20
GitHub Action runner started hanging forever

maxgolov:
Installation phase of full build loop 'hangs forever', never actually getting to do the build. Build scripts haven't changed in a while. The only possible recent change could have beeen the GitHub Actions runner image change, which is beyond our control.

In order to unblock the PR reviews - I am temporarily disabling the CMake setup, and the build of deps from vcpkg packages. I'm also disabling the Windows-clang build of SDK for now. That is until we have better understanding of the culprit.

Windows-clang build best be done with built-in Visual Studio 2019 compiler + cmake + ninja anyways. And it would benefit from some refacting. Using CMake + ninja rather than MSBuild would be a better (and faster) option for Clang builds.. Thus, turning Windows-clang build off should not cause significant regression in our CI coverage.


tools/setup-buildtools.cmd
```diff
@@ -52,7 +56,6 @@ if ERRORLEVEL 1 (
 )
 
 REM Install it
-vcpkg integrate install
```
maxgolov, line 55: Re. this line - I honestly do not remember what was the merit of having it. It does cause some harm, however, because from the bundled zlib, sqlite, google test, and the ones installed in vcpkg directory - local build would improperly pick the vcpkg-integrated libraries after this command. Thus, even if we reinstate it at some point - this line seems to be detrimental.

maxgolov:
The hang is similar to this one: https://github.com/actions/virtual-environments/issues/2098

maxgolov:
@bliptec @kindbe @lalitb @larvacea @mkoscumb @reyang

If you guys can have a quick look - we really need this fix to unblock the CI. Otherwise all PRs show up as red.

bliptec (approved):
Should there be an issue to track re-enabling these parts of the build?

maxgolov:
> Should there be an issue to track re-enabling these parts of the build?

I'd like to get CMake + ninja setup, that would be covering the (currently disabled) clang build. And will be more closely resembling your build setup. I'll see if we have one in our backlog (we used to have it), so I'll assign it to myself and mention in this PR.

maxgolov:
#16 - original issue was about `gn` (Googe Ninja). But I think if we cover cmake+clang+ninja (on Windows) , that will be nearly identical and/or at least very similar. As for the other platforms - we can probably add ninja as an option, and have environment variable that forces `build.sh` to invoke CMake with Ninja rather than with GNU `make`.

PR #812, opened 2021-03-22
Fix CodeQL warnings

maxgolov:
Original customer issue reported by email.

There are three CodeQL 'errors' observed with **certain custom ruleset**. None of these issues get reported by our default ruleset on GitHub CodeQL run. And none of them, in essence, are security issues - since on all codepaths the issues described here are "impossible to happen". But I'm fixing these nevertheless to eliminate the 'noise' in the customer's build pipeline.

CodeQL issues are:

| Severity	| Code	| Description	| Project	| File	| Line	| Suppression | State	| Priority |
|---------|---------|---------------|---------|----|---------|--------------|----------|--------|
| Error	| SM02311	|  In GetCollectorUrl result of `[call](1)` to `[operator const char *](2)` is dereferenced here and may be null.	| 	| cpp_client_telemetry/lib/config/RuntimeConfig_Default.hpp	| 97	| Active	| Normal | 
| Error	| SM02311	|  In GetMetaStatsTenantToken result of `[call](1)` to `[operator const char *](2)` is dereferenced here and may be null. | 		|  cpp_client_telemetry/lib/config/RuntimeConfig_Default.hpp | 	117	|  Active	| Normal | 
| Error	| SM01733	|  Format expects 1 arguments but given 0 | 		|  cpp_client_telemetry/lib/offline/SQLiteWrapper.hpp | 334	|  Active	|  Normal |


larvacea (approved):
Hooray! impossible except why try to reason about the behavior of default-config-merge when we can just be nullptr-safe and also make the customer happy.

mkoscumb (approved):
LGTM

PR #813, opened 2021-03-23
Annotate JNI-accessed class and fields

larvacea:
Some applications are seeing a Java exception that the field `long nativeLogManager` is not present in the java class LogManagerImpl. Possibly R8 has minified the field out of existence; at any rate, the @Keep annotation should not hurt and might help.


maxgolov (approved):

mkoscumb (approved):
LGTM

PR #816, opened 2021-03-30
[ObjC] Add function to set host

thomasameisel:
Add a function to the log configuration to set the host


maxgolov (approved):

wrappers/obj-c/ODWLogConfiguration.mm
```diff
@@ -431,6 +431,11 @@ -(void)set:(nonnull NSString *)key withValue:(nonnull NSString *)value
     (*_wrappedConfiguration)[[key UTF8String]] = [value UTF8String];
 }
 
+-(void)setHost:(nonnull NSString *)host
```
eduardo-camacho, line 434: You might want to add a get method, to keep consistency with other properties functionality.

eduardo-camacho (approved):
Thanks

PR #818, opened 2021-04-01
Error checking for introspection failures

larvacea:
Add error checking in the getLogger method. Some applications are
seeing Java exceptions (no such field) from the nativeGetLogger method.

### DIFF
--- a/lib/jni/LogManager_jni.cpp
+++ b/lib/jni/LogManager_jni.cpp
@@ -869,10 +869,28 @@ Java_com_microsoft_applications_events_LogManagerProvider_00024LogManagerImpl_na
     jstring jSource,
     jstring jScope)
 {
+    if (!thiz) {
+        return 0;
+    }
     auto LogManagerProviderClassID = env->GetObjectClass(thiz);
+    if (env->ExceptionCheck()) {
+        env->ExceptionDescribe();
+        return 0;
+    }
+    if (!LogManagerProviderClassID) {
+        return 0;
+    }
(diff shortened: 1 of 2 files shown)


lib/jni/LogManager_jni.cpp
```diff
@@ -869,10 +869,28 @@ Java_com_microsoft_applications_events_LogManagerProvider_00024LogManagerImpl_na
     jstring jSource,
     jstring jScope)
 {
+    if (!thiz) {
```
mkoscumb, line 872: Should it be (!this) not sure if `thiz` is a typo.

mkoscumb (changes requested):

lib/jni/LogManager_jni.cpp
sid-dahiya: I'm surprised this didn't break the build

sid-dahiya (approved):
Please fix `thiz` to `this`. Looks good otherwise.

theta3:
`thiz` is the `jobject` passed in as parameter. It can be named anything no?

lib/jni/LogManager_jni.cpp
larvacea: thiz is the JNI conventional argument for the "this" Java pointer (jobject), because C++ dislikes it if you decide that some random variable should be named "this."

mkoscumb (approved):
Thanks for clarifying!

PR #820, opened 2021-04-05
Add CentOS8 and Debian10 support

vgerasimovms:


tools/setup-buildtools.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/sh
 
 if [ -f /bin/yum ]; then
+if [ `cat /etc/os-release | grep CentOS-7` != "" ]; then
```
maxgolov, line 4: nit: this one is a bit sketchy..

maxgolov (approved):

tools/setup-buildtools.sh
maxgolov: I mean maybe we need to have a separate notes / document in markdown format that elaborates on this, build toos setup depending on OS, etc. Maybe in the Linux build instructions doc.

maxgolov (approved):

tools/setup-buildtools.sh
vgerasimovms: I think it would be better to move all these installations to docker file, since docker image creation happens during the build

Issue #271, opened 2020-02-07
Support faster, leaky shutdown


bliptec:
For Edge, OneDS (if it's started) will live for as long as the browser, so it'll only ever get shut down right before the process exits. There's no reason to clean up resources (join threads, de-allocate memory, etc) that the OS is about to reclaim anyways. The only things that need to happen are:

1. Persistent storage has all the latest data reasonably flushed
2. Threads and callbacks are in a state where they won't behave poorly (ie. potentially cause access violations) as everything shuts down. Could be achieved by a combination of relying on weak pointers and ensuring pending work finishes.

PR #823, opened 2021-04-07
Add PauseManager to API

larvacea:
Adds a component called PauseManager to the SDK. The intended use case is a multithreaded application that wishes to pause the SDK and wait for it to quiesce.

This is compatible with the current API: if nothing calls the PauseManager::PauseActivity() method, the only effect is to add a compare_and_exchange_weak call and test to some API methods, and everything works as it does today.

Once PauseActivity() is called, future (for the usual sequential-consistency definition of "future") calls to protected methods will return immediately. Past (again, by the sequential-consistency definition) calls are allowed to complete and return normally.

The PauseManger QuiesceWait and QuiesceWaitFor methods will block the calling thread until all in-flight API calls complete.

### DIFF
--- a/lib/include/public/PauseManager.hpp
+++ b/lib/include/public/PauseManager.hpp
@@ -5,6 +5,7 @@
 #ifndef PAUSEMANAGER_HPP
 #define PAUSEMANAGER_HPP
 
+#include <atomic>
 #include <chrono>
 #include <condition_variable>
 #include <mutex>
--- a/lib/include/public/PauseManager.hpp
+++ b/lib/include/public/PauseManager.hpp
@@ -1,0 +1,47 @@
+//
+// Copyright (c) 2015-2020 Microsoft Corporation and Contributors.
+// SPDX-License-Identifier: Apache-2.0
+//
+#ifndef PAUSEMANAGER_HPP
+#define PAUSEMANAGER_HPP
+
+#include <chrono>
+#include <condition_variable>
+#include <mutex>
+
+#include "ctmacros.hpp"
+
+namespace MAT_NS_BEGIN
+{
+    class PauseManager {
+        public:
+
+        static void PauseActivity() noexcept;
+        static void ResumeActivity() noexcept;
+        static void QuiesceWait() noexcept;
+        static void QuiesceWaitFor(std::chrono::milliseconds timeout) noexcept;
+        static bool TryLock() noexcept;
+        static void ReleaseLock() noexcept;
+
+        class Lock {
+        public:
+
+            Lock();
+            ~Lock();
+
+            bool isPaused() const noexcept;
+
+        private:
+            bool acquired_lock;
+        };
+
+        private:
+
+        static std::mutex s_pause_manager_mutex;
+        static std::atomic<uint64_t> s_pause_manager_state;
+        static std::condition_variable s_pause_manager_cv;
+    };
+}
+MAT_NS_END
+
+#endif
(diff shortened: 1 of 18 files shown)


maxgolov:
I am a bit concerned about adding the new class. Since it'd also require adding extra projections for all languages.

Why don't we introduce the following logic instead:
- static `LogManager::OnSuspending` - this is also gonna address this ask: #271 , i.e. getting into safe suspendable state.
- static `LogManager::OnResuming` - this brings the SDK back into functional state.
- instance `LogManagerImpl::Suspend`
- instance `LogManagerImpl::Resume`
- plus, a way to register two lambda functions - platform-specific / PAL layer OS-specific routes, that would trigger the `OnSuspending` and `OnResuming` callbacks, that by-default may call into `Suspend` and `Resume` methods respectively. Maybe in PAL ... This is somewhat similar to what WinRT / Universal code has been doing for the Power Manager callbacks: https://github.com/microsoft/cpp_client_telemetry/blob/7e26c3964ad19186eff2be1026eabf7e65090fff/lib/pal/universal/WindowsRuntimeDeviceInformationImpl.cpp#L99

i.e. hiding `PauseManager` from user eyes / API surface almost entirely, only allowing the customers customize default per-platform behaviors by supplying custom `OnSuspending` and `OnResuming`. On Android and iOS this would have to be bound to corresponding lifecycle events in [Java](https://developer.android.com/guide/components/activities/activity-lifecycle) and [Obj-C](https://developer.apple.com/documentation/uikit/app_and_environment/managing_your_app_s_life_cycle) respectively.

lib/include/public/PauseManager.hpp
```diff
@@ -0,0 +1,48 @@
...
+#include <mutex>
+
+#include "ctmacros.hpp"
+
+namespace MAT_NS_BEGIN
+{
+    class PauseManager {
+        public:
+
+        static void PauseActivity() noexcept;
```
maxgolov, line 20: I think some of it can be implemented inside PAL layer. With unique OS-specific callback that monitors lifecycle events from Java or Obj-C layer. Then there is no need per se to have a separate public API. Also the corresponding "global app state" methods, such as `Suspend` + `Resume` / or `OnSuspending` + `OnResuming` can be added to existing `LogManager`. That makes it easier to project by adding methods  without adding a totally new class.

lib/include/public/PauseManager.hpp
```diff
@@ -0,0 +1,48 @@
...
+        static void ResumeActivity() noexcept;
+        static void QuiesceWait() noexcept;
+        static void QuiesceWaitFor(std::chrono::milliseconds timeout) noexcept;
```
maxgolov, line 23: Is there a need for `QuiesceWaitFor`? I am trying to understand the use-case here. In UWP the apps usually ask for time-limited deferral. It could be something that is managed internally in PAL, i.e. reading the default deferral timeout value from `ILogConfiguration`. That way the defaults can be applied, and customers may be able to override their defaults - to give a hint to SDK how long it may wait for certain activities (such as how long to wait for flush to compete, or how long to wait for HTTP requests cancellation).

lib/include/public/PauseManager.hpp
```diff
@@ -0,0 +1,48 @@
...
+        static bool TryLock() noexcept;
+        static void ReleaseLock() noexcept;
+
+        class Lock {
```
maxgolov, line 27: I'm wondering what value-add is to introduce a separate `Lock` - we can probably reuse the existing `LOCKGUARD` macro instead?

lib/include/public/PauseManager.hpp
larvacea: I could see adding this as instance methods to the ILogManager interface, and make the pause/resume a per-instance decision. I could add this to the macro-instantiated static log managers. Either way, the API change does need to be represented in the Java/ObjC wrapppers.

PR #824, opened 2021-04-08
Fix bool check for Objective-C

sapunhan:
Issue:
Our project was sending bools to kusto as longs instead of true, false. We are using the objective-c wrapper of the sdk. I believe there is an issue in the the sdk to check if an objective-c object is actually a bool

```
strcmp([num objCType], @encode(BOOL))==0 
```

However I found several posts saying this will not work for 64-bit systems (most iphones/simulators today) because an Obj-C Bool is actually an unsigned char.

https://stackoverflow.com/questions/23425187/nsnumber-with-bool-is-no-bool-on-64bit/23426415

Solution:
This post seemed to have an acceptable fix
https://github.com/gabriel/MPMessagePack/commit/c8aeebf820c92a7cd9dbd0fabfa758cca34488a8

### DIFF
--- a/wrappers/obj-c/ODWLogger.mm
+++ b/wrappers/obj-c/ODWLogger.mm
@@ -66,7 +66,7 @@ using namespace MAT;
         std::string strPropertyName = std::string([propertyName UTF8String]);
         if([value isKindOfClass: [NSNumber class]]){
             NSNumber* num = (NSNumber*)value;
-            if(strcmp([num objCType], @encode(BOOL))==0 ) {
+            if((id)num == (id)kCFBooleanTrue || (id)num == (id)kCFBooleanFalse) {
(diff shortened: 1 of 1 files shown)


eduardo-camacho (changes requested):
I suspect this might cause integer fields with 1-0 values to be treated as bools, did you test that scenario?

archahal:
> I suspect this might cause integer fields with 1-0 values to be treated as bools, did you test that scenario?

This doesn't seem to be the case - our loadTimeMs field is of type Int, and it did not go into that block of code we modified. We see it in data explorer as a 1.
Screen Shot 2021-04-08 at 12 25 34 PM

eduardo-camacho (approved):

kindbe (approved):

PR #826, opened 2021-04-13
Adding read me file for Apple podspec

sreejabanerjee:
I have published the new [macOS/iOS Podspec](https://msblox.visualstudio.com/_git/CocoaPods?path=%2FSpecs%2FOneDsCppSdk%2F3.5.67%2FOneDsCppSdk.podspec&version=GBmaster&_a=contents) to Msblox.
This file is tutorial on how to use the Podspec and the Obj-C wrappers in a iOS/macOS project.


docs/start_macOS_iOS.md
```diff
@@ -0,0 +1,61 @@
+# 1DS C++ SDK iOS/macOS Podspec Onboarding
+
+This tutorial guides you through the process of integrating the [1DS SDK](https://github.com/microsoft/cpp_client_telemetry) into your existing iOS and macOS app. Here we consume the [Msblox Podspec]() and published [Obj-C Wrappers](https://github.com/microsoft/cpp_client_telemetry/tree/master/wrappers/obj-c). If you want to consume the C++ Library directly, please follow the tutorials at [Getting started - iOS](https://github.com/microsoft/cpp_client_telemetry/blob/master/docs/cpp-start-ios.md) and [Getting Started - Mac OS X](https://github.com/microsoft/cpp_client_telemetry/blob/master/docs/cpp-start-macosx.md)
+
+## Add OneDsCppSdk Cocoapod to the Podfile
+
+The first version is `3.5.67`.
+
+```
+pod 'OneDsCppSdk’, '3.5.67'
```
maxgolov, line 10: Do we need some podspec file added here, or in the `/build/` directory as a reference?

docs/start_macOS_iOS.md
```diff
@@ -0,0 +1,61 @@
...
+// Log telemetry event
+logger?.logEvent(with: event)
+// …
+ODWLogManager.flushAndTeardown()
+```
+You're done! You can now compile and run your app, and it will send a telemetry event using your ingestion key to your tenant.
+
+## Authors
+
+- @srbanerj
```
maxgolov, line 61: I don't know if you really need to put your alias here -- since your name shows up in Git history anyways. I mean - as we change this, this may have more than one author, and it'd be easier to maintain if we remove the Authors section entirely.

docs/start_macOS_iOS.md
sreejabanerjee: We need to add a source in the Podfile
```
source 'https://msblox.visualstudio.com/DefaultCollection/_git/CocoaPods'
```

I didn't add it since you asked not to mention MSBLOX

maxgolov: I'm thinking of adding a "template" that refers to something like _YOUR-PUBLISHING-URL_ or _YOUR-NUGET-ARTIFACTS-URL_ . I am not familiar enough with Cocoapods, but is it possible to use either some variables or environment variables in the spec? Then we can show our generic template here, without mentioning concrete artifacts storage URL.

docs/start_macOS_iOS.md
```diff
@@ -0,0 +1,59 @@
...
+### Build your project
+
+Make sure your project builds after the previous steps. If you are having issues, then you might not have access to the private repos of 1DS C++ SDK. Make sure you have all the required access.
+
+### Use the Obj-C function signatures in the code. 
+- Initialize Logger with Aria Project Key
+
+``` swift
+var logger: ODWLogger?
+ODWLogManager.setTransmissionProfile(.realTime)
```
eduardo-camacho, line 43: two comments: first, this might falsely give impression that setting transmission profile is mandatory, which is not, better to put a comment. Second: please pass ODWRealTime, to match what enum has defined

docs/start_macOS_iOS.md
maxgolov: You can probably add a spec file that uses environment variable:
https://stackoverflow.com/questions/58795893/possibility-of-adding-environment-variables-on-podfile

Something like:

`source 'https://' + ENV['COCOAPOD_SOURCE_URL']`

in the example Podfile should do the trick, where in the README here - you'd say that environment variable `COCOAPOD_SOURCE_URL` can be used to specify the artifacts / Cocoapod repository. Folks familiar with the matter can adjust it accordingly.. OR we can just say the users specify the source URL as env var prior to running the `pod` command.

maxgolov: i.e.

```console
export COCOAPOD_SOURCE_URL=https://msblox.visualstudio.com/DefaultCollection/_git/CocoaPods
pod ...
```

docs/start_macOS_iOS.md
sreejabanerjee: I use this exact code. For some reason, ODWRealTime === .realTime in my project and I cannot just use ODWRealTime. 

I will remove this line so there is less confusion.

maxgolov:
It generally looks good. My only ask is to include a _TEMPLATE_ of the pod spec file you are using, i.e. create a copy of `OneDsCppSdk` like:

```text
platform :ios, '9.0'
inhibit_all_warnings!

target 'MyApp' do
  pod 'ObjectiveSugar', '~> 0.5'

  target 'MyAppTests' do
    inherit! :search_paths
    pod 'OCMock', '~> 2.0.1'
  end
end

post_install do |installer|
  installer.pods_project.targets.each do |target|
    puts "#{target.name}"
  end
end
```

But not mentioning the `msblox` URL in it. You can still use whatever environment variables in the pod file.

maxgolov:
@sreejabanerjee - I'm gonna merge your PR. It would be great if you can follow-up with a sample template for the actual pod spec file.

maxgolov (approved):

maxgolov (approved):

PR #828, opened 2021-04-16
[WIP] Add compiler warning helper macros.

mkoscumb:
I'm super tired of writing giant ifdef blobs to properly push/disable/pop compiler warnings. This makes things a bit easier and - ideally - more readable when we do need to suppress a warning. Intentionally not converting warnings to  macros in modules (yet), public headers (as I want to avoid collision with consumers), and third party code.

I'm not wild about macros, but this is one of the times where they're warranted.


ThomsonTan:
[@mkoscumb](https://github.com/mkoscumb), could you please resolve the conflicts in this PR?

PR #829, opened 2021-04-19
Pause and resume-or-terminate

larvacea:
Adds methods to ILogManager to support pause and resume. During pause, loggers do not log, flush does not flush, and tpm does not start uploads.

This allows client software to quiesce these activities, either for an operating-system "you are in the background" state change (iOS wants background apps to negotiate for time to do activities like persisting to the database) or a "you are about to tear down all objects" (iOS, again).

Happily, Android kills applications with SIGKILL and there is no such thing as a bad application exit, but even on that platform one might want to sleep-hard or know when the SDK is idle and paused.

No iOS ObjC wrapper in this PR because I do not have a Mac and therefore no xcode.'

This PR includes the Java wrapper changes and Java tests for the interface, and C++ unit tests.

### DIFF
--- a/lib/android_build/maesdk/src/main/java/com/microsoft/applications/events/ILogManager.java
+++ b/lib/android_build/maesdk/src/main/java/com/microsoft/applications/events/ILogManager.java
@@ -68,4 +68,10 @@ public interface ILogManager extends AutoCloseable {
   public boolean registerPrivacyGuard();
 
   public boolean unregisterPrivacyGuard();
+  
+  public void pauseActivity();
+  public void resumeActivity();
+  public void waitPause();
+  public boolean startActivity();
+  public void endActivity();
 }
--- a/lib/api/LogManagerImpl.cpp
+++ b/lib/api/LogManagerImpl.cpp
@@ -364,6 +364,8 @@ namespace MAT_NS_BEGIN
 
     void LogManagerImpl::FlushAndTeardown()
     {
+        PauseActivity();
+        WaitPause();
         LOG_INFO("Shutting down...");
         LOCKGUARD(m_lock);
         if (m_alive)
@@ -754,7 +756,7 @@ namespace MAT_NS_BEGIN
 
     void LogManagerImpl::ResetLogSessionData()
     {
-        if (m_logSessionDataProvider) 
+        if (m_logSessionDataProvider)
         {
             m_logSessionDataProvider->ResetLogSessionData();
         }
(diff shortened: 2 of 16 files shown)


lib/api/LogManagerImpl.cpp
```diff
@@ -364,6 +364,8 @@ namespace MAT_NS_BEGIN
 
     void LogManagerImpl::FlushAndTeardown()
     {
+        PauseActivity();
```
nishchith-cp, line 367: 1. If we are already calling PauseActivity and WaitPause in FlushAndTeardown, do we need to expose wrapper class methods?
2. Since flushAndTeardown is being called in other scenarios except for app termination, is it safe to have PauseActivity and WaitPause here?
cc: @aamittal1989 @karthika-msft

lib/android_build/maesdk/src/main/java/com/microsoft/applications/events/ILogManager.java
```diff
@@ -68,4 +68,10 @@
   public boolean registerPrivacyGuard();
 
   public boolean unregisterPrivacyGuard();
+  
+  public void pauseActivity();
+  public void resumeActivity();
+  public void waitPause();
+  public boolean startActivity();
```
sid-dahiya, line 75: Very minor nit, can this be on top? Then the list feels order in that all the voids are together and the first method is `startActivity` and last is `endActivity`

sid-dahiya: Below as well.

lib/api/LogManagerImpl.cpp
sid-dahiya: I can see `PauseActivity` being useful independently as well. For example, you can pause the logmanager if you know the app is going to suspend and try to resume later. You can also pause for, say, when the device is roaming or in low power mode and resume when things are normal.

lib/api/LogManagerImpl.hpp
```diff
@@ -340,10 +347,20 @@ namespace MAT_NS_BEGIN
         DataViewerCollection m_dataViewerCollection;
         std::vector<std::shared_ptr<IDataInspector>> m_dataInspectors;
         std::recursive_mutex m_dataInspectorGuard;
+
+        std::mutex m_pause_mutex;
+        std::condition_variable m_pause_cv;
+        uint64_t m_pause_active_count = 0;
+        enum class PauseState : uint8_t
+        {
+            Active,
```
sid-dahiya, line 356: Can we call this `NotPaused`? To me, `PauseState::Active` reads as we're actively paused.

sid-dahiya (approved):
A few suggested fixes, but none that are worthy of blocking this change to me.

mkoscumb (approved):

lib/api/LogManagerImpl.cpp
```diff
@@ -868,25 +870,93 @@ namespace MAT_NS_BEGIN
...
+        });
+    }
+
+    bool LogManagerImpl::StartActivity()
+    {
+        std::unique_lock<std::mutex> lock(m_pause_mutex);
+        if (m_pause_state != PauseState::Active) {
+            return false;
+        }
+        m_pause_active_count += 1;
```
kindbe, line 942: ++m_pause_active_count?

lib/api/LogManagerImpl.cpp
```diff
@@ -868,25 +870,93 @@ namespace MAT_NS_BEGIN
...
+        return true;
+    }
+
+    void LogManagerImpl::EndActivity()
+    {
+        std::unique_lock<std::mutex> lock(m_pause_mutex);
+        if (m_pause_active_count == 0) {
+            return;
+        }
+        m_pause_active_count -= 1;
```
kindbe, line 952: --

lib/include/public/ILogManager.hpp
```diff
@@ -402,10 +402,36 @@ namespace MAT_NS_BEGIN
...
+
+        /// <summary>
+        /// Start an activity
+        /// </summary>
+        /// <returns>True if we are not paused and the activity may continue.</returns>
+        virtual bool StartActivity() = 0;
+
+        /// <summary>
+        /// End an activity. StartActivity MUST have returned true: do not call this
+        /// method if StartActivity returned true.
```
kindbe, line 429: Should that last "true" be "false"?

kindbe (approved):
A couple of nits, but otherwise LGTM

PR #831, opened 2021-04-21
Fix deep-copy for `const char *`

maxgolov:
**Problem Description**

This is a fix that makes the following scenario `fool-proof` :
- customer creates a config object that points to temporary string, OR to temporary `const char *` buffer
- previous copy of `const char *` was shallow, i.e. the pointer was copied (assuming that the passed in config is immutable)..
- now if original buffer is kinda gone (e.g. pointer to temporary `std::string.c_str()`), we were running in issues while trying to accessing junk instead of real values.

Another possibility here is due to order of static object destructors, it could have happened that we were unintentionally hitting this on certain platforms. Just because the entire original (owning) object was living somewhere in higher-level abstraction projection was gone before the underlying `LogManager` referencing some pointers on it is gone. Then, despite the fact that it seemed immutable, it was gone - with a shallow copy of config referencing pointers on it.

**Solution**

Make all `const char *` value assignments to land in `std::string` owning object  within the `ILogConfiguration` collection. This may cause minor overhead, i.e. memcpy on assignment. But this is much safer than making assumptions about the immutability of original config.

I also added a small test to validate the first scenario mentioned in the problem description.

Two other unrelated changes:
- Latest Visual Studio 2019 IDE hates it when two projects refer to the same compilation unit. Since `PayloadDecoder.cpp` is included explicitly in `FuncTests.exe`, remove the reference to `decoder.vcxitems`. This fixes the compilation problem for me. Otherwise the project won't open in IDE.
- `Version.hpp` without `END-OF-LINE` is violently rejected by latest Resource Compiler process. This may be a quirk of the latest RC.exe, but I'm adding EOL just to make it happy.

I hope that this should address some recent issues we faced on force-termination of Mobile Apps on Obj-C. If it does not, then we have some other global static objects issue, OR, possibly, an issue with Apple HTTP stack not terminating HTTP requests before the `LogManager` is gone. Anyways, fingers crossed, deep copy of `const char *` into `std::string` seems like a reasonable thing to do regardless.


maxgolov:
One minor word of caution for the Obj-C projection, the objects get also copied usually in the original `LogManagerBase.hpp` template here. Note the LAST line in this snippet:

https://github.com/microsoft/cpp_client_telemetry/blob/884ac265733b9d88f75790081f7edd6417f6fc4b/lib/include/public/LogManagerBase.hpp#L238

```cpp

        /// <summary>
        /// Initializes the telemetry logging system with the specified ILogConfiguration.
        /// </summary>
        /// <param name="tenantToken">Token of the tenant with which the application is associated for collecting telemetry</param>
        /// <param name="configuration">ILogConfiguration to be used.</param>
        /// <returns>A logger instance instantiated with the tenantToken.</returns>
        static ILogger* Initialize(
            const std::string& tenantToken,
            ILogConfiguration& configuration)
        {
            LM_LOCKGUARD(stateLock());
            ILogConfiguration& currentConfig = GetLogConfiguration();
            if (nullptr == instance)
            {
                // Copy alternate configuration into currentConfig
                if (&configuration != &currentConfig)
                {
                    for (const auto& kv : *configuration)
                    {
                        currentConfig[kv.first.c_str()] = kv.second;
...
```

Just to make sure I am not missing the other scenarios, e.g. direct invocation of `CreateLogManager` API or alike. If necessary, the Obj-C code should also do the same thing as `LogManagerBase.hpp` singleton template is doing. I am not touching the Obj-C projection code here.

maxgolov:
@mkoscumb - this one contains the fix for the other issue you noticed.

tests/functests/APITest.cpp
```diff
@@ -290,6 +290,29 @@ TEST(APITest, LogManager_Initialize_Custom)
     LogManager::FlushAndTeardown();
 }
 
+#define SOME_CUSTOM_URL "https://custom.url"
+
+TEST(APITest, LogManager_Initialize_Custom_DeepCopy)
+{
+    ILogConfiguration* config = new ILogConfiguration();
```
mkoscumb, line 297: Why new this up? Doesn't need to be heap allocated.

mkoscumb: `ILogConfiguration config {};` :)

tests/functests/FuncTests.vcxproj
```diff
@@ -94,7 +94,7 @@
   <ImportGroup Label="Shared">
     <Import Project="..\..\Solutions\Clienttelemetry\Clienttelemetry.vcxitems" Label="Shared" />
     <Import Project="..\..\lib\pal\desktop\desktop.vcxitems" Label="Shared" />
-    <Import Project="..\..\lib\decoder\decoder.vcxitems" Label="Shared" />
+    <!-- Import Project="..\..\lib\decoder\decoder.vcxitems" Label="Shared" / -->
```
mkoscumb, line 97: Why comment? Why not just delete this, as it's a strict subset of ClientTelemetry?

tests/functests/APITest.cpp
```diff
@@ -290,6 +290,29 @@ TEST(APITest, LogManager_Initialize_Custom)
     LogManager::FlushAndTeardown();
 }
 
+#define SOME_CUSTOM_URL "https://custom.url"
```
mkoscumb, line 293: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-macros2

tests/functests/APITest.cpp
```diff
@@ -290,6 +290,29 @@ TEST(APITest, LogManager_Initialize_Custom)
...
+        std::string beingBadString = SOME_CUSTOM_URL;
+        // Assign operator using `const char *`
+        (*config)[CFG_STR_COLLECTOR_URL] = beingBadString.c_str();
+    }
+    // beingBadString is now out-of-scope! But we copied the buffer passed to us.
+    ILogger* result = LogManager::Initialize(TEST_TOKEN, *config);
+    EXPECT_EQ(true, (result != NULL));
+    auto& configuration = LogManager::GetLogConfiguration();
+    // Make sure we actually copied from origin to destination configuration on initialization
+    EXPECT_EQ(true, (const char*)(*config)[CFG_STR_COLLECTOR_URL] != (const char*)configuration[CFG_STR_COLLECTOR_URL]);
```
mkoscumb, line 308: Any chance these can be static_cast<const char*>?

larvacea: +1 for the static_cast suggestion. () cast is so kernighan-and-ritchie :).

tests/functests/APITest.cpp
larvacea: +1. the preprocessor is full of hate and bugs.

larvacea (changes requested):

PR #832, opened 2021-04-21
Add additional safety checks to `LogManager` API surface to prevent crashes if called after global static destructors run

maxgolov:
Extra safety checks to cover the case when SDK is attempted to be invoked from rogue runaway thread from context of `std::atexit` handler, e.g on app force termination. Customers should not be calling into SDK at this point. But if they do, we can make the code logic more fool-proof.


lib/include/public/LogManagerBase.hpp
```diff
@@ -36,6 +36,22 @@ ref class LogManagerLock
 
 #include "LogManagerProvider.hpp"
 
+
+/// Termination handler required for Apple Platforms
+static bool isLogManagerTerminated = false;
```
maxgolov, line 41: argghhh.....

lib/include/public/LogManagerBase.hpp
```diff
@@ -161,9 +177,16 @@ namespace MAT_NS_BEGIN
             return lock;
         }
 #endif
+        
+        static bool isGone()
+        {
+            return LogManagerExitHandler::isTerminated();
```
maxgolov, line 183: All of them share the same global static.. i.e. it has to run on any instance in case of multiple log managers, and so the flag trips.

lib/include/public/LogManagerBase.hpp
```diff
@@ -161,9 +177,16 @@ namespace MAT_NS_BEGIN
...
 #endif
+        
+        static bool isGone()
+        {
+            return LogManagerExitHandler::isTerminated();            
+        }
 
         static inline bool isHost()
         {
+            if (isGone())
```
maxgolov, line 188: Doesn't have to be thread-safe since the value of it is only changing when `atexit` processing is about to run. Won't ever flip to `false` during normal app lifecycle.

lib/include/public/LogManagerBase.hpp
```diff
@@ -273,6 +299,9 @@ namespace MAT_NS_BEGIN
         /// </summary>
         static status_t FlushAndTeardown()
         {
+            if (isGone())
+                return STATUS_EALREADY;
```
maxgolov, line 303: ...the process is dying at this point. No need to teardown. Trying to do teardown is detrimental since corresponding config object is already dead, and so we would have tripped on it. Maybe the `stateLock` below (also uses global static) should also have similar global state check.

maxgolov:
This PR also includes the other PR #831 . I am changing target branch to `master` to run the CI / Tests across all platforms with the proposed change to confirm that it does not negatively affect any platform other than Windows.

PR #838, opened 2021-04-23
Stop Unit and Func tests from recompiling all of the files that constitute clienttelemetry.lib

mkoscumb:
This yields better build perf as these projects only build the test files they need, rather than compiling the entirety of files contained within ClientTelemetry.lib. 

The tests will, of course, still depend on ClientTelemetry.lib, and any changes in constituent files will require a re-lib and re-link, but it means that we'll be compiling each file N-2 times, which is probably a win :)

On the build perf improvement: Looking at a couple PRs, the Test-Win32 Release loop looks like it's saving 6-7 minutes out of ~20 minutes (or roughly 35%). Not yet at the non-windows loop speeds, but makes my calculator show a smiley face.

### DIFF
--- a/tests/functests/FuncTests.vcxproj
+++ b/tests/functests/FuncTests.vcxproj
@@ -91,16 +91,7 @@
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.props" />
   <ImportGroup Label="ExtensionSettings">
   </ImportGroup>
-  <ImportGroup Label="Shared">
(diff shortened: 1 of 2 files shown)


kindbe (approved):

maxgolov:
Sorry, I clicked on **Merge** button, as I thought it's a fix for another build issue. I'm not 100% sure if this is safe.

mkoscumb:
Pretty sure it's OK, let me know if it isn't.

PR #841, opened 2021-04-24
Fix for no-offline storage build

### LINKED ISSUE #834: Exception being thrown during initialization of LogManager with  HAVE_MAT_STORAGE disabled.
Building `win32-lib` on VS2019 with MSVC 14.2. `sqlite` module is disabled via the `HAVE_MAT_STORAGE` flag being disabled in `config-custom.h`.

When initializing `LogManager` with `HAVE_MAT_STORAGE` disabled, an exception is thrown. It occurs at Line 89 of `OfflineStorageHandler.cpp`.
```
Exception thrown: read access violation. 
std::shared_ptr<Microsoft::Applications::Events::IOfflineStorage>::operator->**<Microsoft::Applications::Events::IOfflineStorage,0>**(...) returned nullptr.
```

I'm pretty sure this is happening because of `OfflineStorageFactory::Create()` in `OfflineStorageFactory.cpp`. If MAT Storage is disabled, it returns `nullptr` (Lines 39-40 of `OfflineStorageFactory.cpp`).  

Is this was intended behavior? Or am I doing something incorrectly in my initialization to cause this exception?

maxgolov:
Fixes #834

**Changes**

Looks like some recent changes, possibly refactor of storage creation into factory, caused the issue for the no-offline storage build.

I'm adding `nullptr` check to fix this.

**Tests**

Manually ad-hoc tested with `SampleCppMini` + `win32-mini-dll` (x2 times smaller build of SDK with no storage).

Ran the sample and it sends an event and no longer crashes.

**Unrelated Changes**

Ever since I upgraded my local Visual Studio 2019 to latest, I hit the following issues:

* RC compiler does not like `Version.hpp` without EOL.

* Latest compiler does not like no-RTTI option, saying that the parameter is not recognized. It could be that the 2019 IDE is spawning an older 2017 compiler with a new option that it does not recognize it or alike? I am re-enabling RTTI for `win32-min-dll` since its overhead is small. That eliminates the issue with noRTTI compiler option. I am not exactly sure how else this would work with 2017 compiler and 2019 IDE. Appears like we should assume that RTTI is now always on, even on small-sized build of SDK.. again, the overhead of leaving RTTI on (default) is rather small in my opinion.


maxgolov:
`C/C++ CI on Windows (vs2017) / Build (pull_request) In progress — This check has started...` - something wrong with optional vs2017 loop, it hangs on installing the tools (NOT code issue, but build infra).

mkoscumb (approved):
LGTM

PR #842, opened 2021-04-26
Temporarily disable Visual Studio 2017 "full" build loop

maxgolov:
Rename build-windows-vs2017.yaml to build-windows-vs2017.yaml.off

WHY it happens:

- There is no clear indication what is going on, it might be that one of the standard tools installed in Visual Studio 2017 full build loop triggers a "false positive" from Windows Defender, or alike... It could be the `cmake` installer, as there were some bugs reported against GitHub Action docker infra before. **Installation of the tooling now hangs**, whereas just last week it was okay with the same script.

image

HOW to solve:

- Temporarily disable Visual Studio 2017 full build loop. We know it compiles, we continue doing the full build loop with Visual Studio 2019, as well as individual Functional Test projects with "mini" loop that uses Visual Studio 2017 by default. So the coverage is there.. just not the "full build loop"-level of coverage.

Unfortunately I do not have energy to debug the GitHub Actions runner right now (it may take 2 days to get to the bottom of it, as it's a trial-and-error to find what is hanging, given the logs aren't showing, and how to fix it fully, considering then we need to try alternate versions of the build tooling). I'd recommend anyone who feels that they are impacted by turning this loop off - to take the ownership and debug it further, as to what exact portion of the installation script is hanging. I can't, of course, repro any of these hangs on my machine locally.

Meanwhile turning it off to unblock the PRs. Because ALL PRs now showing as failed, blocked by this issue. Root cause is definitely in the GitHub Actions image itself, as we have not touched the rest of infra on our end.


mkoscumb (approved):
LGTM

PR #844, opened 2021-04-27
Update Modules commit

sid-dahiya:
Update commit ID for modules.

Also fix Macro redefinition that is breaking gcc.


maxgolov (approved):
Hoping for the best..

mkoscumb (approved):
LGTM

PR #847, opened 2021-04-29
[iOS Wrapper] Added applicationWillTerminate notification Handler in ODWLogger.mm

dipatid:
Added applicationWillTerminate notification Handler in ODWLogger.mm

Issue:
In Teams iOS, the SDK events are added in a separate queue for asynchronous execution in background thread. In the case when app is force-quit, the main thread starts deallocating the SDK and static objects but the background thread still keeps accessing those in the SDK. This cases random crashes in force-quit scenarios.

Introduced an class method in ODWLogManager - 
`+(void)applicationWillTerminate;`
In this method, we pause event transmission, set the `canUseSDK` boolean flag to false and followed by flushAndTearDown.
In `PerformActionWithCppExceptionsCatch` function in ODWLogger.mm we check if `canUseSDK` is `false`, we immediately return from the function without executing the `block()` in the function.

Host apps should call this method when the app receives `applicationWillTerminate` notification. Calling this API ensures safe termination of the logging library on exit.


wrappers/obj-c/ODWLogger.mm
```diff
@@ -4,6 +4,7 @@
 //
 #include "ILogger.hpp"
 #import <Foundation/Foundation.h>
+#import <UIKit/UIKit.h>
```
eduardo-camacho, line 7: This might not work for Mac

eduardo-camacho (changes requested):
This code is used by both Mac and iOS, but UIKit.h might not work for Mac

wrappers/obj-c/ODWLogger.mm
kindbe: UIKit (iOS) and IOKit (OSX) are both problematic for the shared cocoapod:  https://dev.azure.com/msblox/_git/CocoaPods?path=%2FSpecs%2FOneDsCppSdk%2F3.5.117.1%2FOneDsCppSdk.podspec

wrappers/obj-c/ODWLogger.mm
```diff
@@ -31,6 +34,12 @@ -(instancetype)initWithILogger:(ILogger*)logger
 	        NSLog(@"Logger initialized successfully");
 		}
         _semanticContext = [[ODWSemanticContext alloc] initWithISemanticContext:_wrappedLogger->GetSemanticContext()];
+
+        [[NSNotificationCenter defaultCenter] addObserver:self
+                                                 selector:@selector(applicationWillTerminate:)
```
sid-dahiya, line 42: Since the headers don't agree between iOS and Mac, what about having a callback here that is called when the host application identifies the `applicationWillTerminate` scenario? Then, in the future as the platforms converge/diverge, the SDK won't need to have any changes and the host application can work out the details of know the scenario and update accordingly.

wrappers/obj-c/ODWLogger.mm
```diff
@@ -14,6 +15,8 @@
 
 using namespace MAT;
 
+static BOOL canUseSDK = YES;
```
sid-dahiya: I'm not an Obj-C dev, so would appreciate some clarity here.

Why is `BOOL` preferred over `bool` or `Boolean` here? As someone reading about this now, it seems weird to have `YES` and `NO` values, instead of the `true`/`false` that is peppered across this SDK.

sid-dahiya (changes requested):

wrappers/obj-c/ODWLogger.mm
dipatid: Had thought about this approach to have a static / class method which the app can call when it encounters the appWillTerminate scenario.
I an offline discussion @eduardo-camacho suggests to keep as is. Thoughts?

wrappers/obj-c/ODWLogger.mm
dipatid: Updated with std::atomic<bool> and true /false

wrappers/obj-c/ODWLogger.mm
sid-dahiya: hmm, lemme think a bit. My overall intent is to avoid the need for any platform specific log here. I understand that there may be instance where we must do some platform specific implementation, and in that case, I would recommend covering both iOS and Mac.

maxgolov: Isn't there `#ifdef` you can use to make it iOS only? I think though that force termination applies to Mac as well, i.e. you can `Force Quit` if you press that shortcut (Option while right-clicking on icon). Press these three keys together: Option, Command, and Esc (Escape). Or choose Force Quit from the Apple menu  in the upper-left corner of your screen. (This is similar to pressing Control-Alt-Delete on a PC.) Then select the app in the Force Quit window and click Force Quit...

I don't know though if it triggers the same UX notification or not.

dipatid: One more disadvantage of listening  to notification in the `ODWLogger` is that not every host app may need or want this behaviour. Since, we are adding the check in `PerformActionWithCppExceptionsCatch` function  which is kind of a bottle-neck to access the SDK in anyway.  Other apps may want to access the SDK for some other purposes in appWillTerminate notification and the current check will make all such accesses no-op, giving a false sense to the host app.

dipatid: @sid-dahiya , @maxgolov created an API in `ODWLogManager` instead of listening to the notification in the SDK. The host app can call the API on `appWillTerminate` notification.

wrappers/obj-c/ODWLogger.mm
dipatid: created an API in `ODWLogManager` instead of listening to the notification in the SDK. The host app can call the API on `appWillTerminate` notification.

wrappers/obj-c/ODWLogManager.h
```diff
@@ -149,6 +149,10 @@ typedef NS_ENUM(NSInteger, ODWStatus)
 +(void)setContextWithName:(nonnull NSString*)name
               stringValue:(nonnull NSString*)value
                   piiKind:(enum ODWPiiKind)piiKind;
+/*!
+ @brief This method notifies the  SDK of applicationWillTerminate. The SDK makes all the methods in ODWLoggerManager and ODWLogger a no-op.
```
dipatid, line 153: Need to improve this comment with example of usage.

wrappers/obj-c/ODWLogManager.mm
```diff
@@ -311,4 +311,7 @@ +(void)setContextWithName:(nonnull NSString*)name
         LogManager::SetContext(strKey, strValue, piiValue);
     });
 }
++(void)applicationWillTerminate {
+    canUseSDK = false;
```
dipatid, line 315: [ODWLogManager pauseTransmission];
canUseSDK = false;
[ODWLogManager FlushAndTeardown];

dipatid: Updated.

eduardo-camacho (approved):

maxgolov (approved):

PR #849, opened 2021-04-30
enables building iOS armv7 and armv7s archs

dipidoo:
This change enables building armv7 and armv7s architectures for iOS.


CMakeLists.txt
```diff
@@ -46,15 +46,20 @@ if(APPLE)
     endif()
 
     if(${IOS_ARCH} STREQUAL "x86_64")
-      set(IOS_PLATFORM "iphonesimulator")
```
maxgolov, line 49: I would like to get @kindbe and @eduardo-camacho to sign-off on it.

Where do you move `iphonesimulator` and `iphoneos`, is this intentional removal?

dipidoo: Line 42-46 definitively sets `IOS_PLATFORM` and so not only that the lines I removed seem to be a leftover of a previous logic, it also seems it would not allow the newly introduced arm64 iphonesimulator binaries to be built (I have not verified this empirically as our team has no access to M1 Macs yet).

eduardo-camacho: This was added as a safety check when updated script to support M1, to ensure iOS x86_64 is always simulator (never device), this since ${IOS_PLAT} is coming from sh script. I do not recommend removing it.

eduardo-camacho (changes requested):
You are removing few lines added to support Xcode development on M1.
Please do not remove existing lines unless you fully test all Intel and Silicon scenarios for iOS.

CMakeLists.txt
dipidoo: I suppose I am a little confused as to why these lines exist—I personally would have preferred I receive a FATAL_ERROR or at the very least a warning when a project overrides my explicitly specified argument IOS_PLAT. But you are right that this commit is overreaching in that regard, so I will remove this deletion portion.

For my education once we got our M1s, could you share how one would create an arm64 simulator binary? Is it by specifying `IOS_PLAT=iphoneos` and `IOS_ARCH=arm64`? I would have specified `IOS_PLAT=iphonesimulator` if I hadn’t taken a look at what this file is doing on the arm64 section.

lalitb: @dipidoo - As we are triaging the old PRs, want to check if this PR still needs to be active and merged?

dipidoo: I still do not have access to M1 so I will close this

PR #855, opened 2021-05-03
Fix error: ctmacros.hpp file not found with <angled> include

jevery23:
Cherry-picking from release branch to prevent us from having to cherry-pick explicitly to multiple releases.

### DIFF
--- a/lib/include/public/Version.hpp
+++ b/lib/include/public/Version.hpp
@@ -10,7 +10,7 @@
 #define BUILD_VERSION 3,5,117,1
 
 #ifndef RESOURCE_COMPILER_INVOKED
-#include <ctmacros.hpp>
+#include "ctmacros.hpp"
 #include <stdint.h>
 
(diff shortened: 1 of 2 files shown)


maxgolov (approved):

mkoscumb (changes requested):
This should also update the .template file, so that any subsequent change regenerates the correct file.

bliptec:
> This should also update the version.hpp.template file, so that any subsequent change regenerates the correct file.
> 
> In addition, where is this failing? Missing the `#include` when using <>'s indicates that something's wrong with your INCLUDE paths.

@mkoscumb On Edge, there's stricter checking for using bracketed includes on standard headers and parenthesis for project headers. There was a bit of cleanup related to this in 2018, but any new files added that don't follow that pattern will bubble up as an issue for us.

jevery23:
Updated the template file with the change.

mkoscumb (approved):

PR #857, opened 2021-05-04
Set the NOMINMAX Macro in the Build

mkoscumb:
Windows.h, for a number of probably good reasons at the time, `#define`s `min` and `max`. These of course conflict with std::min and std::max methods.

Windows also offers NOMINMAX, which avoids those defines. 

If there are situations where the Windows SDK min and max macros are necessary (currently none, but there might be some in the future) we should `#undef NOMINMAX` in those specific .cpp files:
```C++
#pragma push_macro("NOMINMAX")
#undef NOMINMAX
#include <windows.h> // or whatever windows header
... // use the icky macros of badness.
#pragma pop_macro("NOMINMAX")```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -171,6 +171,8 @@ if(PAL_IMPLEMENTATION STREQUAL "WIN32")
  add_definitions(-DZLIB_WINAPI)
 endif()
 
+add_definitions(-DNOMINMAX)
+
 ################################################################################################
 # Build prefix and version
 ################################################################################################
--- a/Solutions/net40/net40.vcxproj
+++ b/Solutions/net40/net40.vcxproj
@@ -109,7 +109,7 @@
       </PrecompiledHeader>
       <WarningLevel>Level4</WarningLevel>
       <Optimization>Disabled</Optimization>
-      <PreprocessorDefinitions>ZLIB_WINAPI;WIN32;MATSDK_SHARED_LIB=1;_CRT_SECURE_NO_WARNINGS;_WIN32_WINNT=0x0601;_DEBUG;_WINDOWS;_USRDLL;%(PreprocessorDefinitions)</PreprocessorDefinitions>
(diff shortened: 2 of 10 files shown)

### CODE REVIEW

**Comment by maxgolov** [comment]
Is there a reason why you prefer

`<AdditionalOptions>/DNOMINMAX</AdditionalOptions>`

instead of 

`<PreprocessorDefinitions>...;DNOMINMAX</PreprocessorDefinitions>`

?

I think it would be more consistent to see all build options under the `PreprocessorDefinitions`, in one central spot. Because it is easy to not notice the extra option if it is only explicitly added to additional options. It also forces you to go into a separate UI screen in Visual Studio to see it, i.e. additional opportunity for a failure.

If we can avoid doing that `#define NOMINMAX`, completely.. I feel we'd have less unpredictable interactions with other code that may be building the SDK from source for Windows OS, in non-standard build environments where they may also be needing the original definitions of these macros. We could have wrapped all usages of `std::min` and `std::max` like this:

```cpp
(std::min)(a, b); // added parentheses around function name
(std::max)(a, b);
```

This is search-and-replace in 11 files. But should be rather simple change. We'd still have to keep an eye on any other new usage of `std::min` and `std::max`. But it would not require defining a `NOMINMAX`. Rather hard to tell what exact spot in Windows still require their original macro definition.

maxgolov:
It's about the same scope: 10 project files with define versus (less than) 11 code file targeted fixes to use `(std::min)` and `(std::max)` - that'd fix it in _less invasive_ manner, with about the same amount of total lines changed.

mkoscumb:
There's a `<PreprocessorDefinitions>` blob? Heck yes I'll use that!

maxgolov:
> There's a <PreprocessorDefinitions> blob? Heck yes I'll use that!

Yeah, that'd be much cleaner. I mean - I am OK if you do that. I would have probably preferred if we did not need the macro definition at all, but having it in `PreprocessorDefinitions` should be fine too.

maxgolov (approved):

maxgolov (approved):

PR #859, opened 2021-05-06
ID in /etc/os-release on openSUSE contains quotes

HIROSN:
ext_os_name string on openSUSE may contain quotes for example,
"opensuse-leap"
"opensuse-tumbleweed"
We would like to remove these quotes.

### DIFF
--- a/lib/pal/posix/sysinfo_sources.cpp
+++ b/lib/pal/posix/sysinfo_sources.cpp
@@ -225,6 +225,14 @@ sysinfo_sources_impl::sysinfo_sources_impl() : sysinfo_sources()
     // add("proc_loadavg", {"/proc/loadavg", "(.*)[\n]*"});
     // add("proc_uptime", {"/proc/uptime", "(.*)[\n]*"});
 
+    // osName may contain quotes on openSUSE
+    if (get("osName").find('"') == 0)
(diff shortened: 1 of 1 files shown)


jevery23 (approved):

lib/pal/posix/sysinfo_sources.cpp
```diff
@@ -225,6 +225,14 @@ sysinfo_sources_impl::sysinfo_sources_impl() : sysinfo_sources()
     // add("proc_loadavg", {"/proc/loadavg", "(.*)[\n]*"});
     // add("proc_uptime", {"/proc/uptime", "(.*)[\n]*"});
 
+    // osName may contain quotes on openSUSE
+    if (get("osName").find('"') == 0)
```
maxgolov, line 229: It looks Okay. You could have also probably just patched the contents of cache, because `add` would force a file re-read. But it's a minor thing anyways.

maxgolov (approved):

lib/pal/posix/sysinfo_sources.cpp
HIROSN: Fixed.

maxgolov:
Latest looks better!

PR #865, opened 2021-05-12
Module, Tests, and VS Solution files updates

sid-dahiya:
1. Update of the Modules commit with updated tests.
2. Update of the test infra to consume defaultdataviewer functional tests.
  a. tests/functests/CMakeLists.txt
  b. tests/functests/FuncTests.vcxproj
3. Update of a couple of VS files that are continuously being updated by VS.
  a. Solutions/Clienttelemetry/Clienttelemetry.vcxitems.filters
  b. ~Solutions/MSTelemetrySDK.sln~

### DIFF
--- a/Solutions/Clienttelemetry/Clienttelemetry.vcxitems.filters
+++ b/Solutions/Clienttelemetry/Clienttelemetry.vcxitems.filters
@@ -19,6 +19,7 @@
     <ClCompile Include="$(MSBuildThisFileDirectory)..\..\lib\bond\BondSerializer.cpp" />
     <ClCompile Include="$(MSBuildThisFileDirectory)..\..\lib\callbacks\DebugSource.cpp" />
     <ClCompile Include="$(MSBuildThisFileDirectory)..\..\lib\compression\HttpDeflateCompression.cpp" />
+    <ClCompile Include="$(MSBuildThisFileDirectory)..\..\lib\decoder\PayloadDecoder.cpp" />
     <ClCompile Include="$(MSBuildThisFileDirectory)..\..\lib\decorators\BaseDecorator.cpp" />
     <ClCompile Include="$(MSBuildThisFileDirectory)..\..\lib\filter\EventFilterCollection.cpp" />
     <ClCompile Include="$(MSBuildThisFileDirectory)..\..\lib\http\HttpClient_CAPI.cpp" />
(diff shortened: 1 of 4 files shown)


Solutions/Clienttelemetry/Clienttelemetry.vcxitems.filters
```diff
@@ -1,199 +1,196 @@
...
-    <ClInclude Include="$(MSBuildThisFileDirectory)..\..\lib\utils\StringConversion.hpp" />
-    <ClInclude Include="$(MSBuildThisFileDirectory)..\..\lib\utils\StringUtils.hpp" />
-    <ClInclude Include="$(MSBuildThisFileDirectory)..\..\lib\utils\ZlibUtils.hpp" />
-    <ClInclude Include="$(MSBuildThisFileDirectory)..\..\lib\utils\Utils.hpp" />
-  </ItemGroup>
-  <ItemGroup>
-    <None Include="$(MSBuildThisFileDirectory)..\..\lib\include\public\Version.hpp.template" />
-  </ItemGroup>
-</Project>
+﻿<?xml version="1.0" encoding="utf-8"?>
```
maxgolov, line 1: Is there a way to revert it, or re-normalize to same line endings to avoid the large diff? What are you changing here exactly?

Solutions/MSTelemetrySDK.sln
```diff
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.28307.645
+# Visual Studio Version 16
```
maxgolov, line 2: I usually don't commit that change because I am not 100% sure how that'd be handled by Visual Studio 2017.

sid-dahiya: Ah, that's what is generating it. Eh, I can revert this.

tests/functests/CMakeLists.txt
```diff
@@ -15,6 +15,12 @@ if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/modules/privacyguard/ AND BUILD_PRIVACYGUA
   )
 endif()
 
+if(EXISTS ${CMAKE_SOURCE_DIR}/lib/modules/dataviewer/)
```
maxgolov, line 18: This should work, but also double-check what happens if both main + modules have been checked out. But the build excludes the dataviewer. I think Edge browser does not use your dataviewer implementation, as they have their own that is based on pure debug event callback.

Solutions/Clienttelemetry/Clienttelemetry.vcxitems.filters
sid-dahiya: VS generated this, I think the actual change is line 55, else everything is just a whitespace change. Let me try this again.

tests/functests/CMakeLists.txt
sid-dahiya: hmm, this is the same way we were doing it in UTs though.

Solutions/Clienttelemetry/Clienttelemetry.vcxitems.filters
sid-dahiya: Cleaned up this diff.

tests/functests/CMakeLists.txt
maxgolov: Okay. Let's see what happens next.

maxgolov (approved):

tests/functests/CMakeLists.txt
maxgolov: May need to double-check with @bliptec on that

bliptec: For better or worse, we maintain our own GN build script for Edge, so the CMakeLists changes don't directly affect us (other than needing to reconcile them against our GN build file when updating)

@jevery23 fyi

sid-dahiya: Awesome. In that case, I'll leave it as it is for now.

PR #866, opened 2021-05-13
Fix missing header in PayloadDecoder

sid-dahiya:
This causes the ZLIB flag to be false and LiveDataInspector, amongst other thing, is unable to decode the data.


maxgolov:
lol, it seems like the same fix as Edge folks did in #867 ?

PR #867, opened 2021-05-13
Add mat/config.h include to PayloadDecoder.cpp as it is causing undesired behavior in MS Edge.

jevery23:
Add mat/config.h include to ensure that we have access to CONFIG_CUSTOM_H for clients to use customized defines, such as Edge.

### DIFF
--- a/lib/decoder/PayloadDecoder.cpp
+++ b/lib/decoder/PayloadDecoder.cpp
@@ -2,9 +2,9 @@
 // Copyright (c) 2015-2020 Microsoft Corporation and Contributors.
 // SPDX-License-Identifier: Apache-2.0
 //
-#include "PayloadDecoder.hpp"
-
 #include "mat/config.h"
+
+#include "PayloadDecoder.hpp"
 
 #if !defined(HAVE_MAT_ZLIB) || !defined(HAVE_MAT_JSONHPP)
 
--- a/lib/decoder/PayloadDecoder.cpp
+++ b/lib/decoder/PayloadDecoder.cpp
@@ -4,6 +4,8 @@
 //
 #include "PayloadDecoder.hpp"
 
+#include "mat/config.h"
+
 #if !defined(HAVE_MAT_ZLIB) || !defined(HAVE_MAT_JSONHPP)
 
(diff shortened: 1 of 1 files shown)


maxgolov (approved):
I also see you updated the submodule. I assume this is intentional.

lib/decoder/PayloadDecoder.cpp
```diff
@@ -4,6 +4,8 @@
 //
 #include "PayloadDecoder.hpp"
 
+#include "mat/config.h"
```
maxgolov, line 7: It probalby doesn't matter much. But it's maybe nicer to add it before the `PayloadDecoder.hpp` header. Generally the config goes first, and also in terms of alphabetic order - it should've been first, 'm' before 'P'.

maxgolov: That way if `PayloadDecoder.hpp` class itself makes some assumption about the features being present and what not, it could have had extra `ifdefs` in it - and it would've respected the set of features from `config.h`

jevery23: Oh okay, I thought the header for the unit went first, but I'll change that.

jevery23:
> I also see you updated the submodule. I assume this is intentional.

It wasn't actually, I think my local enlistment generated that change automatically. Is it required? - I just removed it btw.

lib/decoder/PayloadDecoder.cpp
jevery23: Fixed.

maxgolov (approved):
Latest looks even better. Thank you for fixing this.

maxgolov:
>  Is it required? - I just removed it btw.

I don't think it is required immediately. @sid-dahiya will probably send a separate PR for updating modules to latest, as I see it was his change last added to modules.

sid-dahiya (approved):
Thanks for the fix!

sid-dahiya:
> >  Is it required? - I just removed it btw.
> 
> 
> 
> I don't think it is required immediately. @sid-dahiya will probably send a separate PR for updating modules to latest, as I see it was his change last added to modules.

Yeah, I'll update the module once this change is in.

Issue #848, opened 2021-04-29
Make zombie loggers logic more robust


maxgolov:
**Describe your environment.**

This issue is reproducible in one popular app on **older models** of iOS devices with slower processor.

**Steps to reproduce.**

Steps:
- application exiting.
- main thread is calling `FlushAndTeardown`.
- at about the same time another thread is scheduled to perform logging on `ILogger`.
- both clash with a deadlock in zombie logger protection code in `Logger::RecordShutdown()` method.

**What is the expected behavior?**

Well, it is expected that applications do not abuse the logging API that way.. At the same time we have some protection mechanism in place, to allow the safe `use-after-free`. Just that protection mechanism is failing at extremely low rate, unique to the `concurrent-use-during-free`.

**What did you expect to see?**

I expect:
- the app should avoid doing what it is doing.
- the zombie logger logic MAY be improved to handle this race condition / deadlock in zombie-logger protection code in a better way.

**What is the actual behavior?**

Deadlock and hang on app termination, hang in the fool-proof code that is supposed to prevent a crash due to `use-after-free`. As of note, the code very reliably preventing the crash ... by hanging instead. Unfortunately that hang is eventually reported as a crash.

**Additional context.**

The crash rate right now is extremely low. It does not seem to affect newer devices.

I think we need to add the following stress test:
- `Initialize` / `FlushTeardown` in a tight loop on `LogManager` instance.
- rogue thread(s) attempting to obtain loggers via `GetLogger` and log massive volumes of data
Basic expectation here that the app should not crash after a 100,000 iterations like this. I am not sure if we can use some other fuzzy testing tools to artificially cause the deadlock.

Solution could be to perform timed-wait on mutex here:
https://github.com/microsoft/cpp_client_telemetry/blob/a924650883ecfd44f12dba131ca117f502f372b9/lib/api/Logger.cpp#L948

And when we see that the timeout happened, we return status back, and we avoid doing anything on that `ILogger` instance - discarding events that are timing out on that path.

maxgolov:
I think that the `Reinitialize` test in CI cross-plat is also accidentally tripping / hanging on that.. NOT very often, but does happen occasionally.

maxgolov:
The issue is that the loggers map itself is a global static map, and its destructor may be called at the time when `atexit` handlers are running. The force-termination scenario on iOS then may not rely on that zombie loggers map anymore.

PR #869, opened 2021-05-13
Modules commit update.

sid-dahiya:


sid-dahiya:
@maxgolov @mkoscumb is `APITest.LogManager_Reinitialize_UploadNow` test broken? I don't see why it would be failing with the modules update.

maxgolov:
Can you please rerun it? We can trace it back to when it was broken, but I suspect the root cause is related to #848 - getting stuck on logger in use while `FlushAndTeardown` is called. We need to fix the lock guard logic to avoid a deadlock. We can probably try to more precisely pin-point what exact commit started causing it, as it was quite solid, say 6 months ago. This test is abusive and exercises the nastiest possible path, using the logger while system goes up and down. I think it MAY be Okay to temporarily comment it out.

maxgolov (approved):

sid-dahiya:
> Can you please rerun it? We can trace it back to when it was broken, but I suspect the root cause is related to #848 - getting stuck on logger in use while `FlushAndTeardown` is called. We need to fix the lock guard logic to avoid a deadlock. We can probably try to more precisely pin-point what exact commit started causing it, as it was quite solid, say 6 months ago. I think it's around the `RecordShutdown` that was trying to keep loggers zombies, even if the system goes down/ This test is rather abusive and exercises the nastiest possible path, _something that customers should not do_ - using the logger while system goes up and down. I think it MAY be Okay to temporarily comment it out for now, but it'd be great to make this more fool-proof / robust. I also assume that there isn't something like registered callback from any of the live inspector or DDV that went out of scope.

I don't believe there is any DDV, PG or LiveInspector code that is impacting things here. FWIW, the test passed locally for me on two separate Windows boxes and Ubuntu WSLs on two machines. It is certainly something that is exaggerated in the Actions environments.

sid-dahiya:
> Can you please rerun it? We can trace it back to when it was broken, but I suspect the root cause is related to #848 - getting stuck on logger in use while `FlushAndTeardown` is called. We need to fix the lock guard logic to avoid a deadlock. We can probably try to more precisely pin-point what exact commit started causing it, as it was quite solid, say 6 months ago. I think it's around the `RecordShutdown` that was trying to keep loggers zombies, even if the system goes down/ This test is rather abusive and exercises the nastiest possible path, _something that customers should not do_ - using the logger while system goes up and down. I think it MAY be Okay to temporarily comment it out for now, but it'd be great to make this more fool-proof / robust. I also assume that there isn't something like registered callback from any of the live inspector or DDV that went out of scope.

@maxgolov this seems to be a deterministic failure as I've rerun the loops a few times and it fails everytime: https://github.com/microsoft/cpp_client_telemetry/pull/869/checks?check_run_id=2578146000#step:3:2460

maxgolov:
@sid-dahiya - any changes you do that could have been leaving a handler for Live Inspector or DDV? Because the `master` branch now shows all tests passed. Here:
https://github.com/microsoft/cpp_client_telemetry/commit/b4acb9112393968c57199cb0d4b733eba4c2482a

You can click on green checkmark to see that all tests have been passed successfully:
https://github.com/microsoft/cpp_client_telemetry/runs/2572518349

sid-dahiya:
@maxgolov, I don't believe there are any changes in Modules that would impact this, or maybe I'm missing some subtlety here. This is the only net new change here: https://github.com/microsoft/cpp_client_telemetry_modules/commit/0176efa694b4c15f0ffafb177980f58326e74863

maxgolov:
`
[ RUN      ] APITest.LogManager_Reinitialize_UploadNow
/Users/runner/work/cpp_client_telemetry/cpp_client_telemetry/tests/functests/APITest.cpp:1048: Failure
Expected: (shutdownSec + 1) >= (total_time), actual: 21 vs 25
`

Can you temporarily disable the test. This is something timing related, and appears like a transient failure related to possibly Mac host being too busy.

sid-dahiya:
> `[ RUN ] APITest.LogManager_Reinitialize_UploadNow /Users/runner/work/cpp_client_telemetry/cpp_client_telemetry/tests/functests/APITest.cpp:1048: Failure Expected: (shutdownSec + 1) >= (total_time), actual: 21 vs 25`
> 
> Can you temporarily disable the test. This is something timing related, and appears like a transient failure related to possibly Mac host being too busy.

Done and opened #870.

maxgolov:
@sid-dahiya - the rest looks good. Transient failure in re-initialize still happens in another test. It is not as frequently tripping, so I'm merging your change.

PR #873, opened 2021-05-19
Fix Visual Studio 2019 build issues in samples

maxgolov:
In short, the essence of this fix:
1. Try to autodetect the latest Windows 10 SDK.
2. Try to autodetect the Visual Studio version:
- if Visual Studio 2019 is detected, then use `v142` tooling for the samples.
- if Visual Studio 2017 is detected, then use `v141` tooling for the samples.

This ensures that if we "Build All..." in Visual Studio 2019 IDE, even without v141 (vs2017) tools installed, all projects including samples are built successfully.

### DIFF
--- a/examples/cpp/SampleCppUWP/SampleCppUWP.vcxproj
+++ b/examples/cpp/SampleCppUWP/SampleCppUWP.vcxproj
@@ -8,7 +8,6 @@
     <AppContainerApplication>true</AppContainerApplication>
     <ApplicationType>Windows Store</ApplicationType>
     <ApplicationTypeRevision>8.2</ApplicationTypeRevision>
-    <WindowsTargetPlatformVersion>10.0.17763.0</WindowsTargetPlatformVersion>
     <WindowsTargetPlatformMinVersion>10.0.17763.0</WindowsTargetPlatformMinVersion>
     <ProjectName>SampleCppUWP</ProjectName>
     <LatestTargetPlatformVersion>$([Microsoft.Build.Utilities.ToolLocationHelper]::GetLatestSDKTargetPlatformVersion('Windows', '10.0'))</LatestTargetPlatformVersion>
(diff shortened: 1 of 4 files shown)


maxgolov:
@sid-dahiya - I know you've seen it on your machine too. While I'm working on `SetType` PR, I've experienced OCD related to not having `v141` tools installed.. I need all of my projects to build successfully. Thus, fixing it.

examples/cpp/SampleCppUWP/SampleCppUWP.vcxproj
```diff
@@ -11,6 +11,15 @@
     <WindowsTargetPlatformVersion>10.0.17763.0</WindowsTargetPlatformVersion>
```
maxgolov, line 11: Maybe need to remove this explicit dependency on particular Win 10 SDK here as well, and just fall-thru to the auto-detection logic. It happens that I do have this one installed on my machine ( `10.0.17763.0` ), as I believe it does come bundled as one of the defaults with the current version of Visual Studio 2019.

sid-dahiya: Yup, I've had to manually update or install SDKs. I don't believe this property is actually needed, at least not in VS2017 and 2019 as I removed and things still built in both.

examples/cpp/SampleCppUWP/SampleCppUWP.vcxproj
```diff
@@ -11,6 +11,15 @@
     <WindowsTargetPlatformVersion>10.0.17763.0</WindowsTargetPlatformVersion>
     <WindowsTargetPlatformMinVersion>10.0.17763.0</WindowsTargetPlatformMinVersion>
```
sid-dahiya, line 12: Is this line duplicated?

sid-dahiya (approved):

examples/cpp/SampleCppUWP/SampleCppUWP.vcxproj
maxgolov: I think we can remove that.. Then the rest of the flow should be fine, i.e. the logic below should be able to auto-detect the Win 10 SDK -latest automagically. What I'm worried about is if I should test the removal of it with 2017.

maxgolov:
@bliptec @kindbe @lalitb @larvacea @mkoscumb @reyang - can you folks quickly take a look? The change is no-code-change, build change that fixes some issues with vs2019 ide. I'd like to merge that to minimize my other PR.

reyang (approved):
LGTM.

PR #874, opened 2021-05-19
Allow to specify any custom prefix with SetType API

maxgolov:
@sid-dahiya 

https://github.com/microsoft/cpp_client_telemetry_modules/issues/136

**Problem Statement**

This has been discussed during Tuesday's community meeting as a P1 blocker for MS Teams.

Use-case is that the customer wants `EventProperties.SetType(...)` API to respect their own custom type.

Customer is expecting that no prefix should be added to their own base event type.

A little bit of history:

- old most prominently used Telemetry SDK **had a contract** that any custom type should start with `custom.`. The merit of that was that if customer _accidentally_ specifies one of the _standard_ types, used by the old most prominently used data visualization portal... Let's say, does: `evt.SetType("metric")`. And uses `LogEvent(evt)` instead of `LogMetric(evt)`. Then `evt` event does not actually precisely follow the semantics of _metric type_, i.e. not populating all the necessary _metric_ -related semantics / attributes. And then visualization would fail to render, without any reasonable explanation of why that happened so that event of a standard type comes in without the necessary semantic decoration. Thus, `custom.` was added as a defensive measure. And expected base type would be `custom.metric` , not `metric`. Design intent.

To summarize:

`event.SetType("MyEventType");`

was always expected to populate

`record.baseType="custom.myeventtype";`

In the old SDKs, across all language SDKs. Moreover, in prehistoric world, the only allowed value of that would have been `custom_myeventtype` - with all dots replaced by underscores. Because many years ago in quite a few storage systems you could not have a table name that contains a dot in it.. Thus, dots were discouraged and replaced by underscores, e.g.

`record.baseType="custom_myeventtype";`

NOTE: historically the value has also been forcedly lower-cased.

- ...now fast-forward +4 years from the moment when that contract existed. Now the modern SDK for one of the other prominent languages (not C++) obviously does not honor that old Telemetry SDKs contract w.r.t. always appending the prefix. And we are in a situation when the new customer also using that new (other language) SDK is now expecting the new behavior from C++ SDK. Behavior that is not in alignment with the old agreed-upon contract amongst the old SDKs.

And, unfortunately, - neither this new behavior nor the old contract aren't captured anywhere in schema documentation.

The modern (other language 1DS SDK) customer expectation is:

`event.SetType("MyEventType");`

is now expected to populate

`record.baseType="myeventtype";`

without the `custom.` prefix.

**Solution**

**Since we do not want to regress the legacy customer expectations - customers that may be expecting the `custom.` prefix**. And since we would rather honor the legacy contract, we introduce now a new runtime config option that allows to tune the prefix.

Default value of the prefix remains `custom` (with dot appended). New customers who would like to use their own (new behavior) prefix might set it to empty. That way they get precisely what they expect to see. And that way we preserve the legacy customers expected behavior.

New config option:

```cpp
        auto &config = LogManager::GetLogConfiguration();
        config[CFG_MAP_COMPAT][CFG_STR_COMPAT_PREFIX] = "";
        LogManager::Initialize(TOKEN);
        EventProperties myEvent("fooBar");
        myEvent.SetType("myFooBarTypedEvent");
        LogManager::GetLogger()->LogEvent(myEvent); // emits record.baseType="myfoobartypedevent"
```

The new config option now consistently applied to all `ILogger` object obtained from that `LogManager` instance configured to drop the old `custom.` prefix behavior.

NOTE: the reference here while obtaining config object - `auto &config` , NOT a copy of config object. This way we tune the current `LogManager` configuration without necessarily passing the config object back to `LogManager::Initialize`. 

Whereas default value of this map entry remains `custom`, that is to avoid breaking any of the existing / legacy customer expectations.

**Why are we in this state?**

Sometimes we are moving fast at the speed of light. As we move on, we rarely look back at previous experiences, expectations, legacy assumptions, etc. One of the keys to studying and learning history is to establish connections between facts. Fact: old behavior with `custom.` prefix - used to be the foundation of certain most prominent product data cooking pipeline. Breaking that was probably an oversight in the other new language SDK.

I would also reasonably accept the blame for rather poor documentation of this behavior 5 years ago. My old team did not have this contract documented. Got burned once by this already due to NOT having `custom.` prefix elsewhere! 😄 (got burned by something being quite the opposite of what the new customer is asking for).

Now a separate conversation needs to happen in some sort of schema forum, to cover the base type validation / expectations, what the valid values of that `record.baseType` field should be, as well as if data collector performs the necessary validation of it following some sort of input constraints. Be that the min length of the field, whether dots are allowed or not, and whether these should be replaced by underscores. Related parameter that enforces `dots vs. underscores` is `(bool) config[CFG_MAP_COMPAT][CFG_BOOL_COMPAT_DOTS]`. This is potentially not honored by the other fresh SDK, whereas certain pipelines may expect the type field value never contain dots. Such as, for some but not the others, `mycustom_event_type` is the only valid value. No dots.

Debugging the entire E2E flow of that when events don't flow is usually quite entertaining, exciting, and may take days if not weeks. Be that an option to toggle dots to underscores, an option to prepend `custom.`, or any other expectation, such as expecting preserved original case instead of lowercase.

Conclusion: we need more documentation, transparency, be honest. And do more E2E testing as-necessary whenever we introduce a new type or new event. Or when we enable some new event flow in the backend.

**What's next**

Projections for that new config value (2nd-level deep map) must be created for corresponding languages:
- [ ] Obj-C
- [ ] Java

This will be done by corresponding teams working on Mobile, not core part, in separate PRs.

### DIFF
--- a/Solutions/MSTelemetrySDK.sln
+++ b/Solutions/MSTelemetrySDK.sln
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.28307.645
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.31205.134
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "sqlite", "..\sqlite\sqlite.vcxproj", "{2EBC7B3C-2AF1-442C-9285-CAB39BBB8C00}"
 EndProject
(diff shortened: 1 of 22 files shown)


sid-dahiya (approved):
LGTM

maxgolov:
Just a moment, fix for `gcc` is coming. `gcc` is being a bit more finicky than latest `msvc`.

Solutions/MSTelemetrySDK.sln
```diff
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.28307.645
+# Visual Studio Version 16
```
maxgolov, line 2: Sorry, I violated my own principle of not committing Visual Studio 2019 files. If it opens in Visual Studio 2017, then we are still good. I can revert it, if you folks insist on reverting it.

suryatn (approved):

tests/functests/APITest.cpp
```diff
@@ -746,7 +746,7 @@ TEST(APITest, UTC_Callback_Test)
 {
     TestDebugEventListener debugListener;
 
-    auto configuration = LogManager::GetLogConfiguration();
+    auto &configuration = LogManager::GetLogConfiguration();
```
sid-dahiya, line 749: Minor Nit: `auto&`

sid-dahiya: Below as well.

sid-dahiya (approved):

tests/functests/APITest.cpp
maxgolov: I'll fix it.

maxgolov: Cleaned it up for this entire file.

kindbe (approved):

mkoscumb (approved):

PR #875, opened 2021-05-20
Add configuration example that includes A/B EXP feature

maxgolov:
New build config. Follow [this process](https://github.com/microsoft/cpp_client_telemetry/blob/master/docs/building-custom-SKU.md) to build with EXP. Use `CONFIG_CUSTOM_H=config-default-exp.h`.

@anaaru - alternatively you may also use this to diff against the standard `config-default.h` and just patch it as part of vcpkg process. The only change is removing gated check around `HAVE_MAT_EXP` and `HAVE_MAT_FIFOSTORAGE`.

Third option is to pass `HAVE_PRIVATE_MODULES`, I think you already handled that.

### DIFF
--- a/lib/include/mat/config-default-exp.h
+++ b/lib/include/mat/config-default-exp.h
@@ -1,0 +1,35 @@
+//
+// Copyright (c) 2015-2020 Microsoft Corporation and Contributors.
+// SPDX-License-Identifier: Apache-2.0
+//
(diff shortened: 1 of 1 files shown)


sid-dahiya (approved):
LGTM

mkoscumb (approved):

PR #877, opened 2021-05-20
Update Status.java to fix EALREADY

maxgolov:
@larvacea @robertwernsman

Since Java layer is unique to Android, let's use Linux / Unix / Android definition of the errno constant with value of `114`.

Using Chromium as a reference. And double-checking thru Android / Linux definition.

Example references:
- https://chromium.googlesource.com/chromiumos/docs/+/master/constants/errnos.md

MAYBE it would have been much cleaner to use this constant instead?
https://developer.android.com/reference/android/system/OsConstants#EALREADY
But it's a slightly bigger change.


sid-dahiya (approved):
LGTM

PR #879, opened 2021-05-24
Add Custom Prefix configuration to Java and ObjC

sid-dahiya:
Related #874


maxgolov (approved):
Looks good. Maybe if you can add an example (or even a markdown file with 2 paragraph example for Java and Obj-C?) in modules repo, that'd be good.

sid-dahiya:
> Looks good. Maybe if you can add an example (or even a markdown file with 2 paragraph example for Java and Obj-C?) in modules repo, that'd be good.

Lemme see where the existing config examples are and add to that. I realize now we don't have proper API docs generated for the wrappers, probably something to do as a follow-up.

maxgolov:
@sid-dahiya - I added C++ testcase. You may do the same for Java and Obj-C as part of your feature test cases?

PR #881, opened 2021-05-26
InformatonProviderImpl::UnRegisterInformationChangedCallback doesn't unregister the first callback 

### LINKED ISSUE #880: InformatonProviderImpl::UnRegisterInformationChangedCallback doesn't unregister the first callback
We're seeing crashes on Windows Store applications receiving power-state change notifications. The crash is in [InformationProviderImpl::OnChanged](https://github.com/microsoft/cpp_client_telemetry/blob/d2dd0e378a14286c5c4a1ad4778881821739e611/lib/pal/InformationProviderImpl.cpp#L56), specifically:

```c++
size_t count = local_callbacks.size();
for (size_t index = 0; index < count; ++index)
{
    IPropertyChangedCallback* cur_callback = local_callbacks[index];
    if (cur_callback != NULL)
    {
        cur_callback->OnChanged(propertyName, propertyValue);
    }
}
```

The `local_callbacks` contains an entry that is not valid, so our application crashes on `cur_callback->OnChanged(propertyName, propertyValue);`.

The bug appears to be in `InformatonProviderImpl::UnRegisterInformationChangedCallback`, the `index > 0` check looks like it should be `index >= 0`:

```c++
    void InformatonProviderImpl::UnRegisterInformationChangedCallback(int callbackToken)
    {
        size_t index = (size_t)(callbackToken - 1);

        std::lock_guard<std::mutex> lock(m_lock);
        size_t count = m_callbacks.size();
        if (index > 0 && index < count)
        {
            // Don't ever delete the item from the vector.
            // Just set it to NULL.
            m_callbacks[index] = NULL;
            --m_registeredCount;
        }
    }
```

Since the `m_callbacks` is indexed with a 0-base, then `0` is a valid entry - it's the first entry. 

**Steps to reproduce.**
1. Initialize a LogManager that uses an IInformationProvider implementation
2. `LogManager::FlushAndTeardown` the LogManager.
3. Cause a power-state change notification.

No crash.

Crash in `InformationProviderImpl::OnChanged`.

**Additional context.**
I've written unit tests for `InformatonProviderImpl` that cover this scenario and reproduce the problem, and have a fix that fixes the unit tests.

mschofie:
This PR fixes #880.

The implementation of `InformatonProviderImpl::UnRegisterInformationChangedCallback` doesn't correctly validate the `callbackToken`. The `callbackToken` should be the `index + 1` of the entry in the `m_callbacks` to be removed, but after subtracting 1 from the `callbackToken` (to get a 0-based index), `UnRegisterInformationChangedCallback` then checks `if (index > 0 && index < count)`, when it should be checking if `index` is `>=` 0.

Since I couldn't find tests for `InformatonProviderImpl` I added a couple of simple ones. Without the fix, the  `InformatonProviderImplTests.RegisterUnregisterTest` test will fail - the call to `UnRegisterInformationChangedCallback` does not unregister the first callback, so the provided `PropertyChangedCallbackSink` receives the second property-change notification when it shouldn't:

```
[----------] 3 tests from InformationProviderImplTests
[ RUN      ] InformationProviderImplTests.ConstructionTest
[       OK ] InformationProviderImplTests.ConstructionTest (0 ms)
[ RUN      ] InformationProviderImplTests.RegisterTest
[       OK ] InformationProviderImplTests.RegisterTest (0 ms)
[ RUN      ] InformationProviderImplTests.RegisterUnregisterTest
d:\source\github\cpp_client_telemetry\tests\unittests\informationproviderimpltests.cpp(68): error:       Expected: 1u
      Which is: 1
To be equal to: callbackSink.GetCallbacks().size()
      Which is: 2
[  FAILED  ] InformationProviderImplTests.RegisterUnregisterTest (0 ms)
[----------] 3 tests from InformationProviderImplTests (2 ms total)
```

With the fix, the test passes. Huzzah! I built and tested on Windows using `.\Solutions\MSTelemetrySDK.sln`. I've added the tests to `tests/unittests/CMakeLists.txt`, but I couldn't build them on my MacBook - I was seeing unrelated errors. Should things work?

In production, this bug manifests as a crash because the `IPropertyChangedCallback*` that is left when `UnRegisterInformationChangedCallback` doesn't clear the entry refers to a already destroyed instance. The next notification will crash in `InformatonProviderImpl::OnChanged`.

This is my first PR in this project! Please let me know if I've missed some due diligence! Thanks!


mschofie:
Ah... Clang is smarter than me, the breaks are correctly reporting that in "if (index >= 0 && index < count)", 'index is always bigger than 0', 'cos it's unsigned!.

lib/pal/InformationProviderImpl.cpp
```diff
@@ -40,11 +40,17 @@ namespace PAL_NS_BEGIN {
 
     void InformatonProviderImpl::UnRegisterInformationChangedCallback(int callbackToken)
     {
+        // The callbackToken cannot be less than 1.
+        if (callbackToken < 1)
+        {
+            return;
+        }
+
         size_t index = (size_t)(callbackToken - 1);
```
sid-dahiya, line 49: Can we do `static_cast<size_t>` instead of raw cast?

sid-dahiya (approved):
LGTM

lib/pal/InformationProviderImpl.cpp
mschofie: Yes. I was avoiding touching the code to keep things small, but can move to a static cast.

mschofie:
It looks like I have a test failure - I didn't see it locally, but I was (IIRC) only building Debug. The test failure is:

```
[ RUN      ] APITest.LogManager_Reinitialize_Test
Error: Process completed with exit code -1073740791.
```

`-1073740791` is `C000 0409` - STATUS_STACK_BUFFER_OVERRUN. Is this known? Otherwise I'll try repro'ing locally.

maxgolov:
> It looks like I have a test failure - I didn't see it locally, but I was (IIRC) only building Debug. The test failure is:

I apologies, that's a known issue. I do not believe it is actually affecting any of our customers. Let's rerun the test.

maxgolov (approved):

maxgolov:
@mschofie  - All looks good now after rerun. If you do not plan any additional changes, you can merge it now.

mschofie:
> @mschofie - All looks good now after rerun. If you do not plan any additional changes, you can merge it now.

Let me switch over to a C++ style-cast.

maxgolov:
You can apply the C++ cast, and assuming the tests all pass - feel free to merge.

PR #887, opened 2021-06-13
Adjust CMakeLists.txt to build for Windows with MSVC

eshany:
CMakeLists requires some adjustments to build properly for Windows and MSVC.
- Set proper build flags
- Adjust Debug vs non-Debug flags to allow for "RelWithDebInfo" and other release types

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,7 +97,7 @@ message("-- CMAKE_CXX_COMPILER_ID:  ${CMAKE_CXX_COMPILER_ID}")
 include(tools/ParseOsRelease.cmake)
 
 if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
-set(WARN_FLAGS "/W4")
+set(WARN_FLAGS "/W4 /WX")
 else()
 # No -pedantic -Wno-extra-semi -Wno-gnu-zero-variadic-macro-arguments
 set(WARN_FLAGS "-Wall -Werror -Wextra -Wno-unused-parameter")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,15 +92,22 @@ message("-- CMAKE_SYSTEM:           ${CMAKE_SYSTEM}")
 message("-- CMAKE_SYSTEM_VERSION:   ${CMAKE_SYSTEM_VERSION}")
 message("-- CMAKE_BUILD_TYPE:       ${CMAKE_BUILD_TYPE}")
 message("-- TARGET_ARCH:            ${TARGET_ARCH}")
+message("-- CMAKE_CXX_COMPILER_ID:  ${CMAKE_CXX_COMPILER_ID}")
 
 include(tools/ParseOsRelease.cmake)
 
+if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
+set(WARN_FLAGS "/W4")
+else()
 # No -pedantic -Wno-extra-semi -Wno-gnu-zero-variadic-macro-arguments
 set(WARN_FLAGS "-Wall -Werror -Wextra -Wno-unused-parameter")
+endif()
 
 if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
   # Using GCC with -s and -Wl linker flags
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -92,15 +92,22 @@ message("-- CMAKE_SYSTEM:           ${CMAKE_SYSTEM}")
 message("-- CMAKE_SYSTEM_VERSION:   ${CMAKE_SYSTEM_VERSION}")
 message("-- CMAKE_BUILD_TYPE:       ${CMAKE_BUILD_TYPE}")
 message("-- TARGET_ARCH:            ${TARGET_ARCH}")
+message("-- CMAKE_CXX_COMPILER_ID:  ${CMAKE_CXX_COMPILER_ID}")
 
 include(tools/ParseOsRelease.cmake)
 
+if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
+set(WARN_FLAGS "/W4")
```
mkoscumb, line 100: Should also be /WX for parity with GCC/Clang and MSBuild.

mkoscumb (commented):
Please enable Warning as error for MSVC Clang build as well.

CMakeLists.txt
```diff
@@ -114,7 +121,7 @@ endif()
 # Use libtcmalloc for Debug builds memory leaks detection
 set(DBG_FLAGS "-ggdb -gdwarf-2 -O0  ${WARN_FLAGS} -fno-builtin-malloc -fno-builtin-calloc -fno-builtin-realloc -fno-builtin-free")
 
-if (CMAKE_BUILD_TYPE STREQUAL "Release")
+if (NOT CMAKE_BUILD_TYPE STREQUAL "Debug")
```
sid-dahiya, line 124: What's the intent behind this particular change? Are new build type being introduced? If so, could you point to where that'd be as I didn't see any changes here. (_Compiler ID is not build type, correct?_)

eshany: CMAKE has several build types, such as RelWithDebInfo. See [CMAKE_BUILD_TYPE](https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html) for more info. this change is meant to allow them.

sid-dahiya: Just for clarification, Debug behaves as it did today, but all other flavors are now considered `Release` equivalent. Did I understand that correctly?

eshany: Correct

mkoscumb (approved):
LGTM

CMakeLists.txt
```diff
@@ -92,15 +92,22 @@ message("-- CMAKE_SYSTEM:           ${CMAKE_SYSTEM}")
 message("-- CMAKE_SYSTEM_VERSION:   ${CMAKE_SYSTEM_VERSION}")
 message("-- CMAKE_BUILD_TYPE:       ${CMAKE_BUILD_TYPE}")
 message("-- TARGET_ARCH:            ${TARGET_ARCH}")
+message("-- CMAKE_CXX_COMPILER_ID:  ${CMAKE_CXX_COMPILER_ID}")
 
 include(tools/ParseOsRelease.cmake)
 
+if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
+set(WARN_FLAGS "/W4 /WX")
```
mkoscumb, line 100: nit: probably should indent this line much like the bodies of other if blobs.

sid-dahiya (approved):
Shit It.

maxgolov (approved):

CMakeLists.txt
maxgolov, line 100: One minor piece of feedback here: when you add `/W4`, it may be different than the default flag CMake decided to pass for whatever build configuration. You'd be getting a warning, like `Command line warning D9025: overriding '/W3' with '/W4'` .. https://stackoverflow.com/questions/58708772/cmake-project-in-visual-studio-gives-flag-override-warnings-command-line-warnin ... that's because by default the flag for `/W3` is already there. So the more elegant solution could have been the one mentioned in the StackOverflow answer `string(REGEX REPLACE "/W[3|4]" "/w" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")`

PR #895, opened 2021-07-02
[WIP] Disable linter for Android builds to unblock / resolve CI failures

maxgolov:
This is slightly outside of my ownership domain. But since this is blocking our CI, I feel we need to solve the issue.

CI fails because it can't download `groovy-all-2.4.15.jar`. It is not clear what is triggering this issue. As we have changed nothing, and most likely it's the change in the GitHub runner image. I found this [StackOverflow thread](https://stackoverflow.com/questions/53108438/could-not-download-groovy-all-jar-org-codehaus-groovygroovy-all2-4-15-in-and) that recommends a few solutions.

One solution is to disable the app linter. Which will in turn should stop downloading the offending package, that for some strange reason, fails to download. I suspect that it's either very slow connection on the runner, or defender running on host that detects some false-positive in the .jar, and blocks it from being downloaded. So the "fix" is to avoid downloading it.

What it gives --> Android CI should start running tests again (fingers crossed).

@larvacea @anod 

Ad-hoc tested on Windows. Build still works fine with my change:

image

### DIFF
--- a/lib/android_build/app/build.gradle
+++ b/lib/android_build/app/build.gradle
@@ -3,6 +3,12 @@ apply plugin: 'com.android.application'
 apply from: "$rootProject.projectDir/tools.gradle"
 
 android {
+
+    lintOptions {
+        checkReleaseBuilds false
+        abortOnError false
+    }
+
     defaultConfig {
         applicationId "com.microsoft.applications.events.maesdktest"
         versionCode 1


maxgolov:
It still fails fetching some other package on Mac OS X runner. Not good. My local run on Windows box is successful.

anod:
To avoid download errors I've cached gradle in our build this might help since it doesn't changes often

```
- task: Cache@2
  enabled: ${{ parameters.buildCacheEnabled }}
  inputs:
    key: 'v1| gradle | "$(Agent.OS)" | Android/PowerBI/gradle/wrapper/gradle-wrapper.properties'
    path: "${{ parameters.gradleUserHome }}/wrapper/dists"
  displayName: Cache gradle binary
```

anod:
> To avoid download errors I've cached gradle in our build this might help since it doesn't changes often
> 
> ```
> - task: Cache@2
>   enabled: ${{ parameters.buildCacheEnabled }}
>   inputs:
>     key: 'v1| gradle | "$(Agent.OS)" | Android/PowerBI/gradle/wrapper/gradle-wrapper.properties'
>     path: "${{ parameters.gradleUserHome }}/wrapper/dists"
>   displayName: Cache gradle binary
> ```

Actually  I implemented cache in private module in GitHub format [build-android-aar.yml](https://github.com/microsoft/cpp_client_telemetry_modules/blob/master/.github/workflows/build-android-aar.yml)

```
      - name: Cache NDK
        uses: actions/cache@v2
        with: 
          path: /usr/local/lib/android/sdk/ndk/${{ env.ANDROID_NDK_VERSION }}
          key: ${{ runner.os }}-ndk-${{ env.ANDROID_NDK_VERSION }}

      - name: Cache Gradle
        uses: actions/cache@v2
        with: 
          path: ~/.gradle/wrapper/dists
          key: ${{ runner.os }}-gradle-${{ hashFiles('./../cpp_client_telemetry/**/gradle-wrapper.properties') }}
```

PR #900, opened 2021-07-14
Add Reset Privacy Guard API

sid-dahiya:
Also make re-initialization of Privacy Guard a no-op.

### DIFF
--- a/wrappers/obj-c/ODWPrivacyGuard.h
+++ b/wrappers/obj-c/ODWPrivacyGuard.h
@@ -71,6 +71,11 @@ typedef NS_ENUM(NSUInteger, ODWDataConcernType)
  */
 +(void)addIgnoredConcern:(NSString *) EventName withNSString:(NSString *)FieldName withODWDataConcernType:(ODWDataConcernType)IgnoredConcern;
 
+/*!
+ @brief Reset the Privacy Guard instance. This should be used after LogManager::FlushAndTeardown is called.
+ */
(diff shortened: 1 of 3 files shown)


mkoscumb (approved):
LGTM, but I'd like Eduardo to take a look prior to merge.

wrappers/obj-c/ODWPrivacyGuard.h
```diff
@@ -71,6 +71,11 @@ typedef NS_ENUM(NSUInteger, ODWDataConcernType)
  */
 +(void)addIgnoredConcern:(NSString *) EventName withNSString:(NSString *)FieldName withODWDataConcernType:(ODWDataConcernType)IgnoredConcern;
 
+/*!
+ @brief Reset the Privacy Guard instance. This should be used after LogManager::FlushAndTeardown is called.
```
eduardo-camacho, line 75: Not sure I follow, if FlushAndTearDown is usually last interaction client has with 1DS, why a reset is needed? If that is required, wouldn't be better to do that as part of FlushAndTearDown automatically?

eduardo-camacho: Sid clarified offline scenario. This is targeted to Teams, so better to keep it decoupled.

sid-dahiya: Discussed offline. We want to avoid strong dependency between LogManager and PrivacyGuard, and as such are going with the separate API call approach

PR #904, opened 2021-07-20
Fix no-sqlite build for various flavors

maxgolov:
**Problem Description**

The issue is that the SDK size with embedded SQLite is **1,384 KB**.
In some "always connected" scenarios this causes unnecessary overhead.

**How to build without SQLite for Windows using Visual Studio IDE**

```console
set NOSQLITE=TRUE
set CUSTOM_PROPS_VS=C:\work\cpp_client_telemetry\Solutions\build.compact-dll.props
tools\start-ide.cmd
```

Then in Visual Studio, build win10-dll (Microsoft.Applications.Telemetry.Windows.native.dll). Please use Visual Studio 2019, as recent Windows 10 SDK only works well with latest / recent Visual Studio. Resulting size without SQLite: **778 KB** (x2 times smaller DLL).

**How to build without SQLite for Windows using command line**

This batch file builds a custom SKU without SQLite. It also covers the build of Functional Tests without SQLite:

```console
tools\build-Win10-compact-dll.cmd
```

There were several minor quirks, e.g. some projects were not building with custom SKU. For those I explicitly add the proper build flavor they should be built with. Plus, clean-up the tests that were always compiling with assumption that SQLite is there. When it is not there, we should not be building SQLite tests.

Fixes https://github.com/microsoft/cpp_client_telemetry_modules/issues/152 for UWP only.

Android is NOT supported. You'd have to sort it out on your own.

@SahilAshar


Solutions/net40/dllmain.cpp
```diff
@@ -24,7 +24,7 @@ void LockInMemory()
 {
 #ifndef _WINRT_DLL
     HMODULE mstelModule;
-    ::GetModuleHandleExW(GET_MODULE_HANDLE_EX_FLAG_PIN, L"ClientTelemetry.dll", &mstelModule);
+    ::GetModuleHandleExW(GET_MODULE_HANDLE_EX_FLAG_PIN, L"ClientTelemetry3m.dll", &mstelModule);
```
maxgolov, line 27: Actually maybe we need to revert this. The name of the DLL would be different depending on SKU..

lalitb (approved):
LGTM

PR #909, opened 2021-07-30
Update EventPropertiesDecorator to respect the value set by SetTimestamp

kindbe:
EventPropertiesDecorator ignores any value specified in EventProperties::SetTimestamp and all records get the timestamp value set in BaseDecorator (PAL::getUtcSystemTimeinTicks()).  This change updates EventPropertiesDecorator::decorate to set record.time to eventProperties.GetTimestamp() if it is non-zero.

### DIFF
--- a/lib/decorators/EventPropertiesDecorator.hpp
+++ b/lib/decorators/EventPropertiesDecorator.hpp
@@ -114,6 +114,9 @@ namespace MAT_NS_BEGIN {
                 record.data.push_back(data);
             }
 
+            if (eventProperties.GetTimestamp() != 0)
+                record.time = eventProperties.GetTimestamp();
+
             record.popSample = eventProperties.GetPopSample();
 
             // API surface tags('flags') are different from on-wire record.flags
--- a/lib/decorators/EventPropertiesDecorator.hpp
+++ b/lib/decorators/EventPropertiesDecorator.hpp
@@ -114,8 +114,9 @@ namespace MAT_NS_BEGIN {
                 record.data.push_back(data);
             }
 
-            if (eventProperties.GetTimestamp() != 0)
-                record.time = eventProperties.GetTimestamp();
+            auto timestamp = eventProperties.GetTimestamp();
+            if (timestamp != 0)
+                record.time = timestamp;
 
             record.popSample = eventProperties.GetPopSample();
 


maxgolov (approved):

lib/decorators/EventPropertiesDecorator.hpp
```diff
@@ -114,6 +114,9 @@ namespace MAT_NS_BEGIN {
                 record.data.push_back(data);
             }
 
+            if (eventProperties.GetTimestamp() != 0)
```
maxgolov, line 117: Minor nit: I would've cached in a local variable to avoid calling `GetTimestamp()` twice.

maxgolov: e.g.
```cpp
auto ts = eventProperties.GetTimestamp();
if (ts != 0)
  record.time = ts;
```

sid-dahiya: Is the timestamp guaranteed to be 0 when not set? Would it make sense to define this as a `const` variable in the header and compare against that?

sid-dahiya (approved):

lib/decorators/EventPropertiesDecorator.hpp
kindbe: The underlying value is an int64_t, so by definition if it's unset it's 0

maxgolov (approved):

mkoscumb (approved):
Any test updates with this?

maxgolov:
> Any test updates with this?

We have some tests that simply validate the record timestamp. Maybe we can add a few lines / an extra test in that same test, that sets the timestamp 1 hour behind, and verify that it matches.

kindbe:
> > Any test updates with this?
> 
> We have some tests that simply validate the record timestamp. Maybe we can add a few lines / an extra test in that same test, that sets the timestamp 1 hour behind, and verify that it matches.

Drat, I missed Matt's comment prior to hitting the button.  I had looked for EventPropertiesDecorator tests but didn't find any.  @maxgolov if you can point me in the right direction, I can look at updating the record test(s).

maxgolov:
@kindbe - something around here, and/or maybe a callback similar to that one:
https://github.com/microsoft/cpp_client_telemetry/blob/477292da8fb569eab529fe403c6114f6b1fe55ae/tests/functests/APITest.cpp#L680

PR #916, opened 2021-08-19
enable MultiProcessorCompilation `/MP` to speed up compilation

yulin-li:
make a full usage of CPU


lalitb:
Are we seeing improvement in build time with these changes ? It would be interesting to see if any numbers are captured.

maxgolov:
Historically there was some issue with parallel builds in Visual Studio 2017. Maybe it's good in 2019. I think the best investment could be in enabling CMake+ninja+vs2019 based build, that'd be the fastest of them all..

mkoscumb (approved):
LGTM

lalitb (approved):
Approved as discussed in meeting.

yulin-li:
@lalitb The CI build is reduced to 30+ mins by enable `/MP`, and the build is more faster on my machine with more cores.
@maxgolov  it will be great if we can update the windows build to use cmake. I am using the telemetry sdk as a submodule in my project, and I can use CMake to manage the project on Linux, but I need to build the lib and then link to my dlls on windows.

PR #918, opened 2021-08-20
Adding Config Option to Disable Zombie Logger Logic

SahilAshar:
We want to add the ability to disable deferred cleanup of the loggers, so that XAL's cleanup properly cleans up, and our clients can test for memory leaks correctly. 

Happy to change flag name if it's not inline with your style guide.

@lubeltra for vis

### DIFF
--- a/lib/api/LogManagerImpl.cpp
+++ b/lib/api/LogManagerImpl.cpp
@@ -368,22 +368,30 @@ namespace MAT_NS_BEGIN
         LOCKGUARD(m_lock);
         if (m_alive)
         {
-            // before we do anything else, move our Logger instances
-            // to the shut-down state and the s_deadLoggers graveyard.
-            // Calls to these loggers will be benign: before we complete
-            // their RecordShutdown() call, this LogManagerImpl is alive
-            // and well and ILogger methods should work. After that
-            // RecordShutdown(), those ILogger methods do nothing and in
-            // particular do not touch this now-defunct LogManagerImpl.
-            for (auto& kv : m_loggers)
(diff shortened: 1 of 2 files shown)


lib/api/LogManagerImpl.cpp
```diff
@@ -368,22 +368,30 @@ namespace MAT_NS_BEGIN
...
         {
-            // before we do anything else, move our Logger instances
-            // to the shut-down state and the s_deadLoggers graveyard.
-            // Calls to these loggers will be benign: before we complete
-            // their RecordShutdown() call, this LogManagerImpl is alive
-            // and well and ILogger methods should work. After that
-            // RecordShutdown(), those ILogger methods do nothing and in
-            // particular do not touch this now-defunct LogManagerImpl.
-            for (auto& kv : m_loggers)
+            if (m_logConfiguration[CFG_BOOL_DISABLE_ZOMBIE_LOGGERS])
```
lalitb, line 371: Just wondering if this should be config option, or a compile time macro. As disabling the zombie logic can have unexpected behaviour if customer is using the Logger instance incorrectly. With macro, we also have option to remove the dead logger logic completely and no initial memory allocation for `s_deadLoggers`.

Though I know this have been discussed and agreed with other members over teams chat before raising this PR. Approving this PR based on the discussions there.

lalitb (approved):

lib/api/LogManagerImpl.cpp
maxgolov: I think config option is probably fine: keeps it that way for those who wanted it (Office, Teams), at the same time allows the rest to reconfigure?

maxgolov (approved):

PR #929, opened 2021-09-15
Set locale and country code only when supported by Apple OS Version

eduardo-camacho:
languageCode and contryCode are not supported for Mac 10.11 or earlier versions as per Apple documentation: https://developer.apple.com/documentation/foundation/nslocale/1643026-languagecode?language=objc

Added branching to only call APIs when supported


sid-dahiya (approved):
LGTM

PR #937, opened 2021-10-05
Use latest googletest for builds on macOS 11.0 or higher

lalitb:
As discussed over 1ds.sdk.cpp channel in Teams, the functional/unit tests are currently failing for the latest macOS version. The clang compiler on this OS is perhaps not happy with the older version of google-test brought as part of this repo. This PR ensures that we use the latest google-test version for macOS 11.0 or higher.

The older gtest/gmock version is coming from: https://github.com/microsoft/cpp_client_telemetry/tree/main/googletest .
The latest gtest/gmock version is used from third_party/googletest ( either cloned as submodule or else downloaded and copied in that path during build)

Note: github VM for `macos-latest` points to `macos-10.15`, so the problem is not caught in our CI pipeline.


lalitb:
@sid-dahiya @reyang Can you please approve this PR. As GH virtual machines for macOS are upgraded from 10.15 to 11.6, the CI pipeline is failing. This PR fixes to use the latest googletest for ubuntu 20.04 and macOS 11.x, and existing older for other platforms. For #994 which upgrade googletest for all the platforms, let's discuss it before, as it may break build for older platforms not part of CI and which we still support.

The CI pipeline is sometimes failing for #969 (segfault while running c-api tests), which I will address separately.

PR #942, opened 2021-10-14
Fix android CI build ( migrate to maven )

lalitb:
As jcenter is no more accessible from github CI actions ( refer: https://blog.gradle.org/jcenter-shutdown ), this PR migrates the repository to download Gradle build dependencies from jcenter to maven.


lalitb:
Looks like CodeScan is failing for SQL Injection. Will check, and if required add a fix for that too in this PR.

Issue #949, opened 2021-11-03
Potential build break for Ubuntu 16 LTS


reyang:
**Describe your environment.**

Ubuntu 16 LTS

**Steps to reproduce.**

Based on what I heard from @vgerasimovms, it's just running a vanilla build under Ubuntu 16 LTS.

**What is the expected behavior?**

1. Build pass.
2. Need to clarify in the [Build Matrix](https://github.com/microsoft/cpp_client_telemetry#build-environments) whether Ubuntu 16 LTS is supported or not. Currently this is a lack of clarity (e.g. in [Target Platforms](https://github.com/microsoft/cpp_client_telemetry#target-platforms) only Ubuntu 14.04x LTS and Ubuntu (latest) are listed).
3. If Ubuntu 16 LTS is supported, make it part of the required CI, any build break should cause the PR to be blocked.

**What is the actual behavior?**

Build fail.

**Additional context.**

@vgerasimovms believes the build break was introduced by #797.
FYI @lalitb @mkoscumb.

lalitb:
Adding a few more details:
 Build error on ubuntu16 LTS / gcc 5.4.0:
```
[ 63%] Building CXX object lib/CMakeFiles/mat.dir/pal/PAL.cpp.o

/mnt/s/lib/pal/PAL.cpp: In member function 'std::__cxx11::string Microsoft::Applications::Events::PlatformAbstraction::PlatformAbstractionLayer::formatUtcTimestampMsAsISO8601(int64_t) const':

/mnt/s/lib/pal/PAL.cpp:402:32: error: unknown option after '#pragma GCC diagnostic' kind [-Werror=pragmas]

#pragma GCC diagnostic ignored "-Wformat-truncation" // error: 'T' directive output may be truncated writing 1 byte into a region of size between 0 and 16 [-Werror=format-truncation=]

^~~~~~~~~~~~~~~~~~~~~
```
1. The pragma support for ignoring "-Wformat-truncation" was added in gcc8, so even build for ubuntu14.04 LTS should break.  Our CI pipeline is only building on ubuntu-latest, so we don't see the issue for older platforms.

2. We can discuss this in the next community meeting, but ideally SDK should build successfully with gcc4.8 and more( i,e compiler with c++11 support, which includes ubuntu14.04 to ubuntu20.04). We should at least add ubuntu14.04 in the CI.

vgerasimovms:
[...]

PR #950, opened 2021-11-03
Fix the build break for Ubuntu 16 LTS #949

vgerasimovms:
The issue #949


lalitb (approved):
Thanks for the fix.

reyang (approved):
LGTM.
We need to keep #949 open since it has 3 parts (the actual build break, CI coverage, and clarification on the support matrix). Maybe change the PR title to "Fix the build break for Ubuntu 16 LTS" and in the description link to #949.

PR #956, opened 2021-11-15
Update supported platforms and Build Environments

lalitb:
Trying to bring clarity on build/supported matrix as mentioned in #949
 - CI coverage is only supported for ubuntu-18.04 and ubuntu-20.04 as these are the only environment available under Github. We can separately bring ubuntu-14.04 and ubuntu-16.04 under CI coverage by running under container env. 
 - 1DS SDK build breaks under ubuntu 14.04 with gcc 4.8.x ( default compiler ) and gcc 4.9.x. It is successful with gcc5.5. `Build Environments` section is updated to reflect that.
  - 1DS SDK builds successfully on ubuntu 16.04 ( with gcc 5.5.x ) and ubuntu 18.04 (with gcc 7.5.x ) and this is updated in the `Build Environments` section.

### DIFF
--- a/.github/workflows/build-posix-latest.yml
+++ b/.github/workflows/build-posix-latest.yml
@@ -26,7 +26,7 @@ jobs:
     strategy:
       matrix:
         config: [release, debug]
-        os: [ubuntu-latest, macOS-latest]
+        os: [ubuntu-latest, ubuntu-18.04, macOS-latest]
     
     steps:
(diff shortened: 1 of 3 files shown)


sid-dahiya (approved):

lalitb:
I am not sure why tests are crashing on ubuntu 18.04 CI, as they work fine in my local env. Looking into it.

Issue #953, opened 2021-11-06
potential bug when parse .ses files


yulin-li:
**Describe your environment.** Describe any aspect of your environment relevant to the problem, including your SDK version, platform, OS version, etc. If you're reporting a problem with a specific version of a library in this repo, please check whether the problem has been fixed on main brach.

**Steps to reproduce.**
I am testing this SDK on Linux and found the SDKUid changes every time I restart my app.
I think I found the reason after debugging for a while.

In `LogSessionDataProvider.cpp` L117. The SDK try to parse the content from a .ses file, and it has
```c++
StringUtils::SplitString(content, '\n', v);
if (v.size() != 2) {
   return false;
}
```

But the ses file is written by 
```c++
void LogSessionDataProvider::writeFileContents(
        const std::string &path,
        uint64_t sessionFirstTimeLaunch,
        const std::string &sessionSDKUid)
    {
        std::string contents;
        contents += toString(sessionFirstTimeLaunch);
        contents += '\n';
        contents += sessionSDKUid;
        contents += '\n';
```

So the content should be `xx\nxx\n`, then the `StringUtils::SplitString` will split it to 3 (`{"xx", "xx", ""}` (empty string at end)
And then the check `v.size() == 2` always failed. Therefore, .ses file always parsed failed and create a new session id every run.

**What is the expected behavior?**
What did you expect to see?

**What is the actual behavior?**
What did you see instead?

**Additional context.**

lalitb:
> Therefore, .ses file always parsed failed and create a new session id every run.

Agree, this looks to be issue and will create `ses` file with the new `SDKUid` with every restart.

lalitb:
#957 fixes this

PR #957, opened 2021-11-16
Fix session-file parsing logic

lalitb:
A small fix for #953, and add unit/func tests to validate the fix.

The session file parsing logic doesn't take into consideration the \<newline\> character at end of the file, which leads to file-parsing failure and the file getting repopulated with new session data (SDKUid, first-time-launch ts) every time the app is restarted.


lib/offline/LogSessionDataProvider.cpp
```diff
@@ -184,8 +184,9 @@ namespace MAT_NS_BEGIN
         contents += toString(sessionFirstTimeLaunch);
         contents += '\n';
         contents += sessionSDKUid;
+        // Valid line ends with newline as per posix specs ( https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206)
         contents += '\n';
-
+
```
reyang, line 189: nit: remove these trailing spaces

lib/offline/LogSessionDataProvider.cpp
```diff
@@ -140,7 +140,7 @@ namespace MAT_NS_BEGIN
         }
         std::vector<std::string> v;
         StringUtils::SplitString(content, '\n', v);
-        if (v.size() != 2) {
+        if (v.size() != 3) {
```
reyang, line 143: Is there a compat concern (e.g. the file was written by a previous version of the sdk, and the reader is new version)?

lalitb: No, it should be fine as we are only changing the parsing logic in the new sdk. The file was correctly written with older sdk, and newer sdk ensure that it is now read correctly too.

reyang: Thanks.

reyang (approved):
LGTM.

ThomsonTan (approved):
:shipit:

tests/unittests/LogSessionDataTests.cpp
```diff
@@ -54,9 +55,32 @@ TEST(LogSessionDataTests, parse_TwoLinesFirstLaunchTooLarge_ReturnsFalse)
...
+   ASSERT_EQ(sessionSDKUid, "bar");
+}
+
+TEST(LogSessionDataTests, getLogSessionData_ValidInput_SessionDataPersists)
+{
+   TestLogSessionDataProvider logSessionDataProvider1(PathToNonEmptyTestSesFile);
+   logSessionDataProvider1.CreateLogSessionData();
+   auto logSessionData1 = logSessionDataProvider1.GetLogSessionData();
+
+   // Create another provider instance and valiate session data is not re-generated
```
ThomsonTan, line 78:
```suggestion
   // Create another provider instance and validate session data is not re-generated
```

PR #960, opened 2021-11-17
Add Support for an arm64 iOS Simulator CMAKE build

SahilAshar:


CMakeLists.txt
```diff
@@ -33,7 +33,8 @@ if(APPLE)
     set(CMAKE_OSX_DEPLOYMENT_TARGET "" CACHE STRING "Force unset of the deployment target for iOS" FORCE)
     if((${IOS_PLAT} STREQUAL "iphonesimulator") AND (${IOS_ARCH} STREQUAL "arm64"))
       # iOS arm64 simulator is supported starting BigSur
-      # Unfortunately, CMAKE produces a device binary (not simulator) when providing -miphoneos-version-min flag when building iOS arm64 simulator
+      set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -mios-simulator-version-min=${IOS_DEPLOYMENT_TARGET}")
```
lalitb, line 36: From the comment above, it looks like the simulator build is only supported for BigSur. Should we add condition to build SDK for simulator only if running on correct macOS version?

SahilAshar: Sorry for the late response, I've been OOF. 

I think adding a condition to build it just for BigSur and may not work for all use cases. BigSur adds support for running the iOS simulator on arm64 architectures (because of Apple releasing the M1/arm chip). However, there are also cases where a non-arm64/non-BigSur machine could be running Xcode 13 (which adds Xcode support for building the arm64 simulator binaries), and wants to build binaries for _all_ possible simulator versions, and then pass that binary around. It's a very specific use case, but I think gating the arm64 sim builds to just BigSur could cause compatibility issues.

lalitb:
LGTM. would be good if we can get it reviewed it from @eduardo-camacho or @kindbe too.

lalitb (approved):

CMakeLists.txt
eduardo-camacho: Last time I added simulator support on M1 I tested combination. Any change related to simulator should also have extended testing as well. Please share your testing matrix.

eduardo-camacho (changes requested):
Waiting on Sashil to provide testing matrix used to make sure this change do not brake any combination. Word of caution: do not get sidetracked by lipo apparently reporting correct architecture; you might need to actually build a product on intended architecture and make sure it builds and it's able to run both device and simulator (both from M1 and Intel machines).

SahilAshar:
> Waiting on Sashil to provide testing matrix used to make sure this change do not brake any combination. Word of caution: do not get sidetracked by lipo apparently reporting correct architecture; you might need to actually build a product on intended architecture and make sure it builds and it's able to run both device and simulator (both from M1 and Intel machines).

Our internal build pipelines build the full matrix and run tests on them, and it's been passing with these changes. I'll double check and get you a matrix!

lalitb:
> I'll double check and get you a matrix!

@SahilAshar - Did you have a chance to validate this?

PR #966, opened 2021-11-30
Add Android.bp to support Android platform builds

robertwernsman:
This build file enables Android platform builds to include this telemetry library directly. This is to support using source code releases from this Github repo in Surface Android builds without maintaining a fork just to add a build file.

### DIFF
--- a/Android.bp
+++ b/Android.bp
@@ -1,0 +1,118 @@
+cc_defaults {
+    name: "maesdk_defaults",
+    cflags: [
+        "-DBUILD_SHARED_LIBS=1",
+        "-DUSE_ROOM=1"
+    ],
+    cppflags: [
+        "-fexceptions",
+        "-Wno-missing-field-initializers",
+        "-Wno-non-virtual-dtor",
+        "-Wno-reorder-ctor",
+        "-Wno-unused-const-variable",
+        "-Wno-unused-parameter",
+        "-Wno-unused-variable"
+    ],
+    rtti: true,
+    vendor_available: true
+}
+
+cc_library_shared {
+    name: "libmaesdk",
+    defaults: ["maesdk_defaults"],
+    srcs: [
+        "lib/api/AllowedLevelsCollection.cpp",
+        "lib/api/AuthTokensController.cpp",
+        "lib/api/ContextFieldsProvider.cpp",
+        "lib/api/CorrelationVector.cpp",
+        "lib/api/DataViewerCollection.cpp",
+        "lib/api/ILogConfiguration.cpp",
+        "lib/api/LogConfiguration.cpp",
+        "lib/api/LogManager.cpp",
+        "lib/api/LogManagerFactory.cpp",
+        "lib/api/LogManagerImpl.cpp",
+        "lib/api/LogManagerProvider.cpp",
+        "lib/api/LogSessionData.cpp",
+        "lib/api/Logger.cpp",
+        "lib/api/capi.cpp",
+        "lib/backoff/IBackoff.cpp",
+        "lib/bond/BondSerializer.cpp",
+        "lib/callbacks/DebugSource.cpp",
+        "lib/compression/HttpDeflateCompression.cpp",
+        "lib/decorators/BaseDecorator.cpp",
+        "lib/filter/EventFilterCollection.cpp",
+        "lib/http/HttpClientFactory.cpp",
+        "lib/http/HttpClientManager.cpp",
+        "lib/http/HttpRequestEncoder.cpp",
+        "lib/http/HttpResponseDecoder.cpp",
+        "lib/jni/JniConvertors.cpp",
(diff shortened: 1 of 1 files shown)


Android.bp
```diff
@@ -0,0 +1,118 @@
...
+        "lib/system/TelemetrySystem.cpp",
+        "lib/tpm/DeviceStateHandler.cpp",
+        "lib/tpm/TransmissionPolicyManager.cpp",
+        "lib/tpm/TransmitProfiles.cpp",
+        "lib/utils/FileUtils.cpp",
+        "lib/utils/StringUtils.cpp",
+        "lib/utils/ZlibUtils.cpp",
+        "lib/utils/Utils.cpp",
+        "lib/offline/OfflineStorage_Room.cpp",
+        "lib/http/HttpClient_Android.cpp"
```
sid-dahiya, line 81: How is this list sourced? How do we ensure that someone doesn't miss updating this list when a file is updated? Can this file be validated via Actions/CIs?

sid-dahiya (changes requested):
Awaiting more information on the file list.

Android.bp
robertwernsman: This part of the .bp file is for creating the native library, so we borrowed the entries from what CMake is doing. Specifically, these entries are sourced from the [CMakeList.txt](https://github.com/microsoft/cpp_client_telemetry/blob/327c219d0b0b0f83ed2b8f501d1a3c8c95c12433/lib/android_build/maesdk/src/main/cpp/CMakeLists.txt#L22) file. We also add an additional two entries at the bottom of this list (OfflineStorage_Room and HttpClient_Android). These are conditionally added in the CMake file, but we add them to our list here since Android.bp ([Soong](https://android.googlesource.com/platform/build/soong/) build system) does not support conditionals or control flow.

Since this list is already maintained in that Cmake txt file, it would be great if we could somehow point to that without having to re-create it here in this build file. I don't know of a solution to that at this time.

I need to investigate what options we have for Actions/CI checks.

sid-dahiya: Sounds good. In that case, I'd suggest at least having a comment in the CMakeList.txt to point authors towards updating this file as well. Preferably if we can have CI validation, that'll be best, but if we can alternatively have some script that does a quick validation to ensure the lists are in sync, it will ensure you don't have to go debug weird build breaks you may run into.

I'm satisfied with this answer and am signing off with the understanding you'll investigate the Action/CI and consider adding a comment to CMakeList.txt.

sid-dahiya (approved):
I am signing off with the understanding you'll investigate the Action/CI and consider adding a comment to CMakeList.txt.

lalitb (approved):
Looks good, but preferably if we can add this in the CI to ensure it builds and runs fine ( in the simulator ).

lalitb:
@robertwernsman as discussed in the community meeting - to investigate if the bp file can be generated automatically using cmake extension.

robertwernsman:
After investigating and asking around, it seems that there is no way to create a Android.bp file from a CMake list. And as I mentioned in the meeting, doing a build with the .bp file would require a full AOSP platform build environment which would require significant effort to enable.

Would it be acceptable to check in the .bp file as is, and also add a comment in the Cmakelist.txt file to mention that this duplicate list exists?

Android.bp
```diff
@@ -0,0 +1,118 @@
...
+        "-Wno-unused-variable"
+    ],
+    rtti: true,
+    vendor_available: true
+}
+
+cc_library_shared {
+    name: "libmaesdk",
+    defaults: ["maesdk_defaults"],
+    srcs: [
```
maxgolov, line 23: Is there something in there that can't be included using `**/*` - I don't know, maybe `lib/api/**/*.cpp` would've worked. Totally realize that explicitly listing each item is more explicit than adding all by mask.

Android.bp
```diff
@@ -0,0 +1,118 @@
...
+    ],
+    static_libs: [
+        "libz"
+    ]
+}
+
+android_library {
+    name: "maesdk",
+    srcs: [
+        "lib/android_build/maesdk/src/main/java/**/*.java"
```
maxgolov, line 101: ..like here it's all by mask, just one-liner.

maxgolov (approved):

lalitb:
> Would it be acceptable to check in the .bp file as is, and also add a comment in the Cmakelist.txt file to mention that this duplicate list exists?

@mkoscumb do you see any concern here?

Android.bp
robertwernsman: That might also work. We were trying to replicate how the Cmakelist selectively chooses .cpp files based on the build configuration. I don't think we (myself or the original folks who set up our builds in Surface) ever tested using a wildcard and investigated the library output.

sid-dahiya (approved):

PR #972, opened 2021-12-02
Add option to enable CAPI http for Android

vgerasimovms:


lalitb (approved):
LGTM.  tagging @saurabh-sagrawal, @larvacea if they have comments from the android perspective.

sid-dahiya:
What's the motivation of this change? Why is CAPI based http client needed here?

lalitb:
> What's the motivation of this change? Why is CAPI based http client needed here?

Good question, will wait for @vgerasimovms to reply. The scenario I could think of was trying to use C-API from the android app using JNI. Vitaliy can confirm if the default HTTP client for android is not getting initialized in that case?

vgerasimovms:
> 

in MIP we use CAPI and we re-define task dispatcher and HTTP methods - it is how that works today for MAC, iOS, Linux, Windows
I'm in the process to start using OneDS client for android in the same way how that had done for other platforms. I've faced issue: HTTP client is null for Android. I believe that is the bug because in that case, I see dereferencing of nullptr, what leads to a crash. Do you have a sample of using CAPI for Android (test)?

lalitb:
> Do you have a sample of using CAPI for Android (test)?

I don't think we have it in the repo. But SDK should configure default HTTP Client (based on the platform ) if HAVE_MAT_DEFAULT_HTTP_CLIENT is enabled.

vgerasimovms:
> I don't think we have it in the repo. But SDK should configure default HTTP Client (based on the platform ) if HAVE_MAT_DEFAULT_HTTP_CLIENT is enabled.

in the SDK it tries to use what jni provides, but we don't use java wrappers for calling OneDS - we use c++ calls c++. In that case no default http client present. One way is to use curl, but it is not part of ndk and we will have to build it by ourselves.

lalitb:
> in the SDK it tries to use what jni provides, but we don't use java wrappers for calling OneDS - we use c++ calls c++. In that case no default http client present. One way is to use curl, but it is not part of ndk and we will have to build it by ourselves.

Ok, I think it would help if someone with android expertise ( @larvacea @saurabh-sagrawal  ? )  have a look at this PR.

maxgolov (approved):
You may need to get some prior history from Trevor on this. He did the original implementation of CAPI for MIP SDK.

vgerasimovms:
I've talk to Trevor, and he said #ifdef for Android was introduced because of it wasn't fully supported at that time.

About http client: So, if OneDS client will be compiled without curl and will be run without involving **mat_open_with_params** with overridden http stack it has to be run/initialized through JNI - otherwise NO HTTP client at all?

saurabh-sagrawal:
> > in the SDK it tries to use what jni provides, but we don't use java wrappers for calling OneDS - we use c++ calls c++. In that case no default http client present. One way is to use curl, but it is not part of ndk and we will have to build it by ourselves.
> 
> Ok, I think it would help if someone with android expertise ( @larvacea @saurabh-sagrawal ? ) have a look at this PR.

Actually I'm confused here, what is a CAPI?
Is the intention to use the sdk in android app without any java code to initialize the sdk?

vgerasimovms:
> Actually I'm confused here, what is a CAPI? Is the intention to use the sdk in android app without any java code to initialize the sdk?

We build C++ SDK and create java wrapper where we expose just SDK functionality and prefer don't do that for dependencies (OneDS)

lalitb:
> We build C++ SDK and create java wrapper where we expose just SDK functionality and prefer don't do that for dependencies (OneDS)

I am fine merge this PR ( Max also has approved it ), just curious - the changes  ( in this PR ) are not required if you use `mat_open_with_params()` -> `mat_log()` -> `mat_upload()` capi functions through jni ?

vgerasimovms:
> I am fine merge this PR ( Max also has approved it ), just curious - the changes ( in this PR ) are not required if you use `mat_open_with_params()` -> `mat_log()` -> `mat_upload()` capi functions through jni ?

It still requires changes: **mat_open_with_params** calls **mat_open_core** which ignore http override in case of Android

PR #975, opened 2021-12-08
Try to fix the potential data race in InformatonProviderImpl

PetrGuan:
I'm from Outlook iOS team. Not sure if this is a false positive, so please correct me if I'm wrong. The `Thread Sanitizer` complains the potential data race.

Screen Shot 2021-12-07 at 5 32 54 PM

### DIFF
--- a/lib/pal/InformationProviderImpl.cpp
+++ b/lib/pal/InformationProviderImpl.cpp
@@ -73,22 +73,22 @@ namespace PAL_NS_BEGIN {
             // be deleted.  The current design is that IPropertyChangedCallback is
             // not refcount'ed.  Should we refcount it?
 
-            std::lock_guard<std::mutex> lock(m_lock);
-            if (m_registeredCount > 0)
+            std::vector<IPropertyChangedCallback*> local_callbacks;
             {
-                std::vector<IPropertyChangedCallback*> local_callbacks;
+                std::lock_guard<std::mutex> lock(m_lock);
+                if (m_registeredCount > 0)
                 {
                     local_callbacks.insert(local_callbacks.end(), m_callbacks.begin(), m_callbacks.end());
-                }
+                }    
+            }
 
-                size_t count = local_callbacks.size();
-                for (size_t index = 0; index < count; ++index)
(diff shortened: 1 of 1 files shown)


lib/pal/InformationProviderImpl.cpp
```diff
@@ -73,11 +73,11 @@ namespace PAL_NS_BEGIN {
             // be deleted.  The current design is that IPropertyChangedCallback is
             // not refcount'ed.  Should we refcount it?
 
+            std::lock_guard<std::mutex> lock(m_lock);
```
lalitb, line 76: While this rightly increases the scope of mutex lock, it looks like it's not required to keep lock explicitly during callback function call (`callback::OnChanged`)

PetrGuan: Thanks for the comments! Any suggestions to this? Can we unlock?

lalitb: Don't see any straightforward fix. Either 
 - move the`local_callbacks` vector definition outside the lock guard. or
 - Use explicit lock/unlock instead of lock_guard.

lalitb: In general fine with it. Bringing `local_callbacks` vector outside may have some memory overhead in terms of heap allocation (std::allocator) every time the `InformatonProviderImpl::OnChanged` is called. But this should be negligible as most of the STL implementations have initial reserve capacity as `0` for std::vector. Still would be good if we can have another eye on this.

lalitb (approved):

lalitb (approved):
LGTM. Would be good if someone else too can review on the potential memory overhead of vector initialization.

lib/pal/InformationProviderImpl.cpp
lalitb: We can wait for a couple of days for comments and merge.

lalitb:
I think it should be good to merge, don't see any further comments/concerns.

PetrGuan:
Thanks @lalitb, appreciate your help ! Can you help me merge this? I'm not sure if i have the merge permission.

Issue #973, opened 2021-12-07
HttpClient SendRequestAsync often has partial response


shae-brown:
**Describe your environment.**
Platform: Windows
OS Version - 10.0.22000
SDK Version - 3.5.131.1
Has not been fixed on main branch

**Steps to reproduce.**
Not able to consistently reproduce. I have a TTD trace of a repro.
Based on our telemetry this is occurring often, ~10% of the time.
Seems to happen for some, but not for others.

**What is the expected behavior?**
The full response payload returns

**What is the actual behavior?**
Missing chunks of the payload

**Additional context.**
What is happening in HttpClient_WinInet.cpp

First call to InternetReadFile
- Returns false indicating failure
- Error is ERROR_IO_PENDING
- Number of bytes read is 0
- However, I can see that the buffer contains the beginning of the request. We don't do anything with this since it says 0 bytes read

Second call to InternetReadFile
- Returns true indicating call is okay
- Number of bytes read 1024
- The buffer contains the section of the request that follows AFTER the first chunk that was already read into the buffer

shae-brown:
Two issues

PR #976, opened 2021-12-08
Fix HttpClient SendRequestAsync bugs

shae-brown:
#973 

- "To ensure all data is retrieved, an application must continue to call the InternetReadFile function until the function returns TRUE and the lpdwNumberOfBytesRead parameter equals zero." [InternetReadFile](https://docs.microsoft.com/en-us/windows/win32/api/wininet/nf-wininet-internetreadfile) Updating the while condition to match API documentation
- "The lpdwNumberOfBytesRead parameter will be populated on async completion, so it must exist until INTERNET_STATUS_REQUEST_COMPLETE. The same is true of lpBuffer parameter.." [Async InternetReadFile](https://docs.microsoft.com/en-us/windows/win32/wininet/asynchronous-example-application) Ensuring the byte count & buffer stays valid by making them member variables.

The bug is reproducible with slower network condition. Tested by enabling Fiddler -> Rules -> Performance -> Simulate Modem Speeds
Before: Consistently hit responses that were missing chunks
After: Response body now contains full response

### DIFF
--- a/lib/http/HttpClient_WinInet.cpp
+++ b/lib/http/HttpClient_WinInet.cpp
@@ -34,7 +34,7 @@ class WinInetRequestWrapper
     BYTE                   m_buffer[1024] {0};
     DWORD                  m_bufferUsed {0};
     std::vector<uint8_t>   m_bodyBuffer;
-    bool                   m_responsePending {false};
+    bool                   m_readingData {false};
     bool                   isCallbackCalled {false};
     bool                   isAborted {false};
   public:
@@ -325,12 +325,12 @@ class WinInetRequestWrapper
             // trigger INTERNET_STATUS_REQUEST_COMPLETE again.
 
             m_bodyBuffer.insert(m_bodyBuffer.end(), m_buffer, m_buffer + m_bufferUsed);
-            while (!m_responsePending || m_bufferUsed != 0) {
+            while (!m_readingData || m_bufferUsed != 0) {
                 BOOL bResult = ::InternetReadFile(m_hWinInetRequest, m_buffer, sizeof(m_buffer), &m_bufferUsed);
+                m_readingData = true;
                 if (!bResult) {
                     dwError = GetLastError();
                     if (dwError == ERROR_IO_PENDING) {
-                        m_responsePending = true;
                         // Do not touch anything from this thread anymore.
                         // The buffer passed to InternetReadFile() and the
                         // read count will be filled asynchronously, so they
--- a/lib/http/HttpClient_WinInet.cpp
+++ b/lib/http/HttpClient_WinInet.cpp
@@ -33,7 +33,8 @@ class WinInetRequestWrapper
     SimpleHttpRequest*     m_request;
     BYTE                   m_buffer[1024] {0};
     DWORD                  m_bufferUsed {0};
-    std::vector<uint8_t>   m_bodyBuffer{};
+    std::vector<uint8_t>   m_bodyBuffer {};
+    bool                   m_responsePending {false};
     bool                   isCallbackCalled {false};
     bool                   isAborted {false};
   public:
@@ -325,10 +326,15 @@ class WinInetRequestWrapper
 
             m_bodyBuffer.insert(m_bodyBuffer.end(), m_buffer, m_buffer + m_bufferUsed);
             do {
+                if (m_responsePending && m_bufferUsed == 0) {
+                    LOG_TRACE("InternetReadFile finished async");
+                    break;
+                }
                 BOOL bResult = ::InternetReadFile(m_hWinInetRequest, m_buffer, sizeof(m_buffer), &m_bufferUsed);
                 if (!bResult) {
                     dwError = GetLastError();
                     if (dwError == ERROR_IO_PENDING) {
+                        m_responsePending = true;
                         // Do not touch anything from this thread anymore.
                         // The buffer passed to InternetReadFile() and the
                         // read count will be filled asynchronously, so they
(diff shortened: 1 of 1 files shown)


lib/http/HttpClient_WinInet.cpp
```diff
@@ -32,6 +32,8 @@ class WinInetRequestWrapper
     HINTERNET              m_hWinInetRequest {nullptr};
     SimpleHttpRequest*     m_request;
     BYTE                   m_buffer[1024] {0};
+    DWORD                  m_bufferUsed {0};
+    std::vector<uint8_t>   m_bodyBuffer{};
```
wayvad, line 36: Nit: insert a trailing space after 'm_bodyBuffer' for consistency.

lib/http/HttpClient_WinInet.cpp
```diff
@@ -316,19 +318,13 @@ class WinInetRequestWrapper
...
-        std::vector<uint8_t> & m_bodyBuffer = response->m_body;
-        DWORD m_bufferUsed = 0;
-
         if (dwError == ERROR_SUCCESS) {
             // If looking good so far, try to fetch the response body first.
             // It might potentially be another async operation which will
             // trigger INTERNET_STATUS_REQUEST_COMPLETE again.
 
             m_bodyBuffer.insert(m_bodyBuffer.end(), m_buffer, m_buffer + m_bufferUsed);
             do {
```
wayvad, line 327: Do not need to enter this loop (more accurately: prevent calling InternetReadFile again) if the previous call to IRF returned ERROR_IO_PENDING followed by a zero-byte buffer on completion -- and this function is called on completion of the previous IRF.

lib/http/HttpClient_WinInet.cpp
```diff
@@ -339,16 +335,20 @@ class WinInetRequestWrapper
...
                     break;
                 }
 
                 m_bodyBuffer.insert(m_bodyBuffer.end(), m_buffer, m_buffer + m_bufferUsed);
-            } while (m_bufferUsed == sizeof(m_buffer));
+            } while (m_bufferUsed != 0);
         }
 
+        std::unique_ptr<SimpleHttpResponse> response(new SimpleHttpResponse(m_id));
+        response->m_body = m_bodyBuffer;
```
wayvad, line 350: On error, better to not copy the body buffer at all to prevent returning partial/corrupted body. So perhaps move this line to inside the L353 if block.

lib/http/HttpClient_WinInet.cpp
```diff
@@ -339,16 +335,20 @@ class WinInetRequestWrapper
...
+
         // SUCCESS with no IO_PENDING means we're done with the response body: try to parse the response headers.
         if (dwError == ERROR_SUCCESS) {
```
wayvad, line 353: Do we need to reset the state of the new members after the InternetReadFile operation completes? (is this class designed to be reused again?)

wayvad (changes requested):

lib/http/HttpClient_WinInet.cpp
shae-brown: fixed

lib/http/HttpClient_WinInet.cpp
shae-brown: good call, moved into the if block below

lib/http/HttpClient_WinInet.cpp
mkoscumb: `{}` is unnecessary as std::vector has a default c'tor.

lib/http/HttpClient_WinInet.cpp
shae-brown: good call, moved to if block below

lib/http/HttpClient_WinInet.cpp
shae-brown: No, it is only made for one request.
It is deleted at the end of this call

lib/http/HttpClient_WinInet.cpp
shae-brown: Added the if statement, is this what you mean?

shae-brown: Changed to !m_responsePending || m_bufferUsed != 0 like suggested

mkoscumb (commented):
Seems reasonable, but I've not touched this code in any meaningful manner. If possible I'd like someone on Reiley's team to take a look as well.

lib/http/HttpClient_WinInet.cpp
```diff
@@ -325,11 +325,7 @@ class WinInetRequestWrapper
             // trigger INTERNET_STATUS_REQUEST_COMPLETE again.
 
             m_bodyBuffer.insert(m_bodyBuffer.end(), m_buffer, m_buffer + m_bufferUsed);
-            do {
-                if (m_responsePending && m_bufferUsed == 0) {
-                    LOG_TRACE("InternetReadFile finished async");
-                    break;
-                }
+            while (!m_responsePending || m_bufferUsed != 0) {
```
wayvad, line 328: Need to make sure "m_bufferUsed != 0" is re-evaluated after IRF is called (not after IRF returns pending). So perhaps rename this to something like m_readingData instead?

wayvad (changes requested):

wayvad (approved):
LGTM

lib/http/HttpClient_WinInet.cpp
```diff
@@ -32,6 +32,9 @@ class WinInetRequestWrapper
     HINTERNET              m_hWinInetRequest {nullptr};
     SimpleHttpRequest*     m_request;
     BYTE                   m_buffer[1024] {0};
+    DWORD                  m_bufferUsed {0};
+    std::vector<uint8_t>   m_bodyBuffer;
+    bool                   m_readingData {false};
```
lalitb, line 37: nit - Can we add a comment after these member variables indicating their purpose. would make it easy to understand the logic. 
PR looks good.

lalitb (approved):
LGTM. Thanks for fixing the issue.

sid-dahiya (approved):
This look like a good change. Are there any tests you may be able to add here?

shae-brown:
This file doesn't have any unit testing yet, and looks like it would be difficult to test.
We will be integrating this change into the Teams repo and testing this out end to end

PR #977, opened 2021-12-13
Use msbuild from environment

anod:
windows-2016 is deprecated (https://github.com/actions/virtual-environments/issues/4312), the build is failing when running on windows-2019

### DIFF
--- a/.github/workflows/test-win-latest.yml
+++ b/.github/workflows/test-win-latest.yml
@@ -21,13 +21,13 @@ on:
 
 jobs:
   test:
-    name: Test on Windows ${{ matrix.arch }}-${{ matrix.config }}
+    name: Test on Windows ${{ matrix.arch }}-${{ matrix.build }}
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
         arch: [Win32, x64]
-        config: [Release, Debug]
-        os: [windows-2016]
+        build: [Release, Debug]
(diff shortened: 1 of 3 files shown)


.github/workflows/test-win-latest.yml
```diff
@@ -21,20 +21,25 @@ on:
...
-    name: Test on Windows ${{ matrix.arch }}-${{ matrix.config }}
+    name: Test on Windows ${{ matrix.arch }}-${{ matrix.build }}
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
         arch: [Win32, x64]
-        config: [Release, Debug]
-        os: [windows-2016]
+        build: [Release, Debug]
+        os: [windows-2019]
```
lalitb, line 30: Can you also update the README.md ( https://github.com/microsoft/cpp_client_telemetry#target-platforms, Covered with CI column)? Or do you think we should keep both Windows 2016 and Windows 2019 in CI?  Windows 2016 end of life is Jan 12, 2027, and SDK should ideally continue supporting it.

anod: windows-2016 will be removed in 3 month and jobs will start fail fron January (https://github.com/actions/virtual-environments/issues/4312), support it will require a custom image

another problem which we might hit is cogs...

lalitb (approved):
LGTM.

PR #980, opened 2022-01-03
Natalieparry/fixdllbuild

natalieparry:
Restoring the .lib build for the win32 dll. This is different from the static version of the entire lib and serves as an import library for the dll that is important for c++ projects trying to link the lib. See issue #979 

Validation: Built the samplecpp project against the new import lib/dll. The project ran without reporting any linking failures.

PR #982, opened 2022-01-05
Finish DLL Build Fix

natalieparry:
This is an update to PR #980, which fixes the win32-dll build. In [this comment](https://github.com/microsoft/cpp_client_telemetry/pull/980#pullrequestreview-843123385) @lalitb correctly pointed out that there was an additional win32-lib reference that was now not necessary. It also breaks the build via multiple definitions and therefore must be removed. Thanks for pointing that out! This change will resolve the issue.

Local testing: build-x64Debug.bat script now correctly builds the dll


lalitb (approved):

lalitb:
@natalieparry - thanks for the fix. you would be merging this PR, or else are there any more changes required here?

PR #983, opened 2022-01-10
Fix the data race in DispatchEvent

PetrGuan:
I'm from Outlook Mobile iOS team. While debugging with ThreadSanitizer, found another possible data race. 

Screen Shot 2022-01-10 at 1 09 45 PM

### DIFF
--- a/lib/callbacks/DebugSource.cpp
+++ b/lib/callbacks/DebugSource.cpp
@@ -35,13 +35,15 @@ namespace MAT_NS_BEGIN {
     /// <summary>Microsoft Telemetry SDK invokes this method to dispatch event to client callback</summary>
     bool DebugEventSource::DispatchEvent(DebugEvent evt)
     {
-        seq++;
-        evt.seq = seq;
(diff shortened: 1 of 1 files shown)


lib/callbacks/DebugSource.cpp
```diff
@@ -35,13 +35,15 @@ namespace MAT_NS_BEGIN {
     /// <summary>Microsoft Telemetry SDK invokes this method to dispatch event to client callback</summary>
     bool DebugEventSource::DispatchEvent(DebugEvent evt)
     {
-        seq++;
-        evt.seq = seq;
-        evt.ts = PAL::getUtcSystemTime();
+        int64_t timestamp = PAL::getUtcSystemTime();
```
lalitb, line 38: `evt.ts = PAL::getUtcSystemTime();` needn't be inside critical region?

PetrGuan: Updated, thanks!

lalitb (approved):
LGTM.

PR #985, opened 2022-01-10
Let the windows UWP build work in unpackaged app

yulin-li:
Currently the UWP build tries to get app name/version using `Package::Current` api, which is not available if the process is running in unpacked app (see [this doc](https://docs.microsoft.com/en-us/uwp/api/windows.applicationmodel.package.current?view=winrt-22000)).

I am using `cpp_client_telemetry` to build my UWP dlls, which could run in packaged app or a native win32 environment. So the `Package::Current` would cause crash.

This PR add a checking to check if the process is running in a packaged app.

We can talk internally via Teams/e-mail if you have more concern.


lib/pal/universal/WindowsRuntimeDeviceInformationImpl.cpp
```diff
@@ -120,8 +120,11 @@ namespace PAL_NS_BEGIN {
                         }
                     });
 
-                    token1 = ::Windows::System::Power::PowerManager::BatteryStatusChanged += onPowerSourceChanged;
-                    token2 = ::Windows::System::Power::PowerManager::PowerSupplyStatusChanged += onPowerSourceChanged;
+                    if (MAT::IsRunningInApp())
+                    {
+                        token1 = ::Windows::System::Power::PowerManager::BatteryStatusChanged += onPowerSourceChanged;
+                        token2 = ::Windows::System::Power::PowerManager::PowerSupplyStatusChanged += onPowerSourceChanged;
+                    }
```
lalitb, line 127: Why are these steps not required for unpackaged apps?

lalitb:
@yulin-li - Is this PR still relevant. Can you please reply to the review comment to enable proceed further? Thanks

lib/pal/universal/WindowsRuntimeDeviceInformationImpl.cpp
yulin-li: Thanks for review and sorry for late reply as Chinese New Year holiday.

I didn't find any doc to say `BatteryStatusChanged` cannot work for unpackaged app, but it indeed crash in the deconstructor where we unsubscribe the event `::Windows::System::Power::PowerManager::BatteryStatusChanged -= token1;`

I removed these and it works fine.

lalitb:
> @yulin-li - Is this PR still relevant. Can you please reply to the review comment to enable proceed further? Thanks

@yulin-li - Reminder, if we still want to get these changes merged ?

lalitb (approved):

namrog84:
@yulin-li  FYI this caused win10-dll to no longer compile for me because GetModuleHandle identifier not found :(

yulin-li:
Sorry for the late reply. @namrog84, is this issue still blocking you?

namrog84:
We aren't blocked and everything is fine for us right now.

We have had to make a few custom changes to get working with VS2022, Win11, and do be BinSkim/Component Governance Microsoft Compliance. 

I don't have the specific changes readily available to mention and I can always re-open a bug/issue in future when/if we hit things again, next time we snap to newer version.  So feel free to close and move on.

PR #987, opened 2022-01-13
Try to fix the data race in DeviceStateHandler

PetrGuan:
I'm from Outlook Mobile iOS team, found another data race: 

SUMMARY: ThreadSanitizer: data race (OLMAnalyticsKit:x86_64+0xa90565) in `Microsoft::Applications::Events::DeviceStateHandler::_UpdateDeviceCondition()+0x65`

### DIFF
--- a/lib/tpm/DeviceStateHandler.cpp
+++ b/lib/tpm/DeviceStateHandler.cpp
@@ -26,6 +26,7 @@ using namespace PAL;
 void DeviceStateHandler::Start()
 {
 	// TRACE("_RetrieveAndRegisterForDeviceConditionChange");
+	LOCKGUARD(m_lock);
 
 	m_networkInformation = PAL::GetNetworkInformation();
 	if (m_networkInformation)
@@ -63,6 +64,7 @@ void DeviceStateHandler::Start()
 void DeviceStateHandler::Stop()
 {
 	// TRACE("Enter stop transmission policy manager");
+	LOCKGUARD(m_lock);
 
 	// 4. Stop listening to platform network, power and device info changes
 	if (m_networkInformation)
@@ -91,7 +93,7 @@ void DeviceStateHandler::OnChanged(
 {
     // TRACE("OnChanged: Platform network callback with Name=%s, Val=%s",
  //       propertyName.c_str(), propertyValue.c_str());
-
+    LOCKGUARD(m_lock);
 
     if (propertyName.compare(NETWORK_TYPE) == 0)
     {
@@ -117,6 +119,8 @@ void DeviceStateHandler::OnChanged(
  ******************************************************************************/
  void DeviceStateHandler::_UpdateDeviceCondition()
 {
+     LOCKGUARD(m_lock);
+	 
 #ifdef _WIN32
      if (m_networkInformation)
      {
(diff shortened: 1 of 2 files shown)


PetrGuan:
I'm not sure if this is the right way to fix, since this class is completely not lock protected

lib/tpm/DeviceStateHandler.cpp
```diff
@@ -26,6 +26,7 @@ using namespace PAL;
 void DeviceStateHandler::Start()
 {
 	// TRACE("_RetrieveAndRegisterForDeviceConditionChange");
+	std::lock_guard<std::mutex> lock(m_lock);
```
lalitb, line 29: I generally don't have a good feeling about adding a mutex guard for the `potential` critical region just by going through the ThreadSanitizer report. What makes you think that the methods within this class would get called from multiple threads simultaneously?

PetrGuan: My thought was this `Start` probably won't need to be guarded by the mutex, while the `Stop` and `OnChanged` might be needed, what do you think?

lalitb:
@PetrGuan - This looks to be a false alarm to be. Looking to the three methods in this class:
 - `DeviceStateHandler::Start` and `DeviceStateHandler::Stop` - These two methods are called through TPM startup [here ](https://github.com/microsoft/cpp_client_telemetry/blob/1fd167fa1595853cd423f9d3229915ad0092a09b/lib/tpm/TransmissionPolicyManager.cpp#L54) and shutdown [here](https://github.com/microsoft/cpp_client_telemetry/blob/1fd167fa1595853cd423f9d3229915ad0092a09b/lib/tpm/TransmissionPolicyManager.cpp#L59). Both these flows are called from the main thread when LogManager is initialized or shutdown.

- `DeviceStateHandler::OnChanged` is supposed to be called from platform to notify any network, power, device info changed. and the caller needs to ensure to call it in a thread-safe way as done [here](https://github.com/microsoft/cpp_client_telemetry/blob/c488c5467ac3af04958e8c26522af31828691b54/lib/pal/InformationProviderImpl.cpp#L91).

PetrGuan:
@lalitb Thanks for the explanation! Let me get through to make sure if I understand this correctly.

So according to the TS report the race could be `m_networkCost` or `m_powerSource`. Reading the codes it seems only `Start` and `OnChanged` might lead to this. 

So as you pointed out, the `Start` will always be called from main thread and only once, and by design the `OnChanged` callback should always be called after the `Start`. 

For the `OnChanged` as you pointed out again, it's from `NetworkInformationImpl.mm` 's `SetupNetDetect` func. And the thread-safety is ensured by `dispatch_get_global_queue`.

So to conclude there's no data race, this is a false alarm. 

Is my understanding correct?

lalitb:
> @lalitb Thanks for the explanation! Let me get through to make sure if I understand this correctly.
> 
> So according to the TS report the race could be `m_networkCost` or `m_powerSource`. Reading the codes it seems only `Start` and `OnChanged` might lead to this.
> 
> So as you pointed out, the `Start` will always be called from main thread and only once, and by design the `OnChanged` callback should always be called after the `Start`.
> 
> For the `OnChanged` as you pointed out again, it's from `NetworkInformationImpl.mm` 's `SetupNetDetect` func. And the thread-safety is ensured by `dispatch_get_global_queue`.
> 
> So to conclude there's no data race, this is a false alarm.
> 
> Is my understanding correct?

Yes understanding is correct here.

PetrGuan:
Thanks! I'll close this.

PR #994, opened 2022-02-18
[WIP] Remove google test source drop dependency

sid-dahiya:
The source drop of googletest is _very old_ and are causing weird test behavior. There have been band-aid fixes to move things on specific platforms, but this change is fixing it everywhere.


sid-dahiya:
This was done via #997

PR #995, opened 2022-02-21
Try to fix the potential data race in Statistics

PetrGuan:
Can we remove the usage of m_intervalMs, or if this is a false alarm for the data race?

image


PetrGuan:
@lalitb can you take a look? Thanks!

lalitb (approved):
LGTM

PR #997, opened 2022-03-04
Use latest version of googletest (v.1.11.0) for testing across all platforms.

lalitb:
This change 
 - Deletes existing older version of googletest from the repo. The folder now only contains the vsproj files which are used to build the latest version of googletest. The vsproj files are generic enough to build the new googletest.
 - Googletest submodule upgraded to the new version ( v1.11.0). Platform-specific build scripts updated to clone this version in third-party if not already present.
 - Building googletest for the iPhone simulator requires custom flags in the googletest cmake file which are now added during build time.


sid-dahiya (approved):

googletest/googlemock/msvc/2015/gmock.vcxproj
```diff
@@ -103,7 +103,7 @@
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
     <ClCompile>
       <Optimization>Disabled</Optimization>
-      <AdditionalIncludeDirectories>..\..\include;..\..;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+      <AdditionalIncludeDirectories>..\..\..\..\third_party\googletest\googlemock\include;..\..\..\..\third_party\googletest\googlemock\;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
```
sid-dahiya, line 106: Should we move this file into the test folder? Maybe something like tests/googletest?

sid-dahiya (approved):
One comment about moving googletest vcxproj files, but good to commit if wanted.

googletest/googlemock/msvc/2015/gmock.vcxproj
lalitb: These include files are coming from googletest submodule, which is configured in `<repo>\third_party folder. But we can move vsproj files from `<repo>\googletest` to `<repo>\test`. I kept it as it is, so as not to change their references in code.

sid-dahiya: I think we should move them. it would be weird to have random googletest/msvc 2015 files in a folder when it is only intended for tests.

lalitb: Good point. Will move them.

build-gtest.sh
```diff
@@ -15,53 +15,43 @@ else
...
+  message("-- CMAKE_OSX_SYSROOT       \${CMAKE_OSX_SYSROOT}")
+elseif(\${ARCH} STREQUAL "arm64")
+  set(CMAKE_C_FLAGS "\${CMAKE_C_FLAGS} -arch arm64")
+  set(CMAKE_CXX_FLAGS "\${CMAKE_CXX_FLAGS} -arch arm64")
+  set(CMAKE_SYSTEM_PROCESSOR arm64)
+  set(APPLE True)
+endif()
+EOF
+
+sed -i -e "/^cmake_minimum_required/r $GTEST_PATH/CMakeLists_temp.txt"  $GTEST_PATH/CMakeLists.txt
```
ThomsonTan, line 52: Is this remove `cmake_minimum_required`? If yes, why need it?

googletest/googlemock/msvc/2015/gmock.vcxproj
```diff
@@ -103,7 +103,7 @@
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
     <ClCompile>
       <Optimization>Disabled</Optimization>
-      <AdditionalIncludeDirectories>..\..\include;..\..;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
```
ThomsonTan, line 106: >AdditionalIncludeDirectories [](start = 7, length = 28)

Just curious, could we use cmake to generate the project file form googletest repo under third_party, instead of check in the project files directly?

build-gtest.sh
lalitb: This inserts an additional set of cmake instructions after `cmake_minimum_required`. These instructions are required to build googletest for the ios simulator.

googletest/googlemock/msvc/2015/gmock.vcxproj
lalitb: Good point. I tried that but the googletest build fails for win32. The problem is cmake by default generates solution files either for 32 bit or 64 bit based on generator options. The custom solution files include both. Also, there are a few changes in those files over time to speed up the build and didn't want to lose them.
We can probably look at it separately if it's possible to get rid of it?

ThomsonTan: Consider it in future PR is fine.

ThomsonTan (approved):

tests/functests/FuncTests.vcxproj
```diff
@@ -28,10 +28,10 @@
     </ProjectConfiguration>
   </ItemGroup>
   <ItemGroup>
-    <ProjectReference Include="..\..\googletest\googlemock\msvc\2015\gmock.vcxproj">
+    <ProjectReference Include="..\googletest\googlemock\msvc\2015\gmock.vcxproj">
```
sid-dahiya, line 31: Should we also remove the `msvc\2015` path as well?

sid-dahiya (approved):
Minor comment around removing sub-folders where possible, which you're welcome to do as a follow-up change.

Issue #991, opened 2022-02-01
Loggers with different collector URLs at a time, +(nullable ODWLogger *)loggerWithTenant:(NSString *)tenantToken source:(NSString *)source withConfig:(nonnull ODWLogConfiguration *)config;


nishchith-cp:
**- Describe your environment:** 
iOS
pod 'OneDsCppSdk.iOS', '3.5.200'

**Scenario:**
Want to use the below API to create a new logger with a different configuration. But the API does not allow sending the following dictionary configuration of PG as part of the parameter. Is there any other way to set the dictionary configuration for the API?

```
**Config that needs to be sent:**
    // For privacy guard, set CFG_MAP_COMPAT's CFG_STR_COMPAT_PREFIX to "" to set our custom baseType for events.
    NSDictionary *compatLogConfig = @{TSOneDSConfigurationKey.cfgStrCompatPrefix: @""};
    [logConfiguration setValue:compatLogConfig forKey:TSOneDSConfigurationKey.cfgMapCompat];

**API:**
/*!
 @brief Retrieves a new instance of ODWLogger for logging telemetry events. This function expects that the telemetry logging system has already been initialized.
 @param tenantToken A string that contains the tenant token.
 @param source A string that contains the name of the source of events.
 @param config A Dictionary that contains a custom configuration.
 @return An ODWLogger pointer that points to the logger for the specified tenantToken and source.
 */
+(nullable ODWLogger *)loggerWithTenant:(NSString *)tenantToken
                  source:(NSString *)source
                  withConfig:(nonnull ODWLogConfiguration *)config;

```

nishchith-cp:
Discussed offline with @sid-dahiya 

1. The configuration for PG base type can be passed as an empty string for the TSOneDSConfigurationKey.cfgMapCompat instead of a dictionary.
2. Call `oneDsLogger?.initializePrivacyGuard(with: pgInitConfig)` if needed for logger created using different LogManager. 

3. Call this API to get a logger for a different config only if the config is different. 
```
+(nullable ODWLogger *)loggerWithTenant:(NSString *)tenantToken
                  source:(NSString *)source
                  withConfig:(nonnull ODWLogConfiguration *)config
```

nishchith-cp:
Need some more clarifications for the above API.

1. Test accounts to test the end to end behavior of the below API to get the logger for different config
```
+(nullable ODWLogger *)loggerWithTenant:(NSString *)tenantToken
                  source:(NSString *)source
[...]

PR #1005, opened 2022-03-24
Update multiple log manager test

lalitb:
The multiple log manager functional test is updated to test for 3 different log managers. This is to demonstrate that each of these log managers is configured with a separate configuration (and so separate collector endpoint), and are sending events to configured collector endpoints.  As per #991, the configuration of the latest(i.e, last created) log manager is getting used for all the loggers, which is not what we are observing as per the tests.

FYI - @nishchith-cp


tests/functests/MultipleLogManagersTests.cpp
```diff
@@ -80,6 +102,15 @@ class MultipleLogManagersTests : public ::testing::Test,
         config1["name"] = "Instance2";
         config1["version"] = "1.0.0";
         config1["config"]["host"] = "Instance2";  // host
+
```
eduardo-camacho, line 105: Don't you need to set config2 as well?

lalitb: Good catch. Fixed - all config1, config2, and config3 are set properly now.

eduardo-camacho (approved):

tests/functests/MultipleLogManagersTests.cpp
```diff
@@ -36,13 +36,39 @@
...
-class MultipleLogManagersTests : public ::testing::Test,
-                                 public HttpServer::Callback
+class RequestHandler : public HttpServer::Callback 
+{
+   public:
+    RequestHandler(int id) : count_(0), id_(id){}
+
+    int onHttpRequest(HttpServer::Request const& request, HttpServer::Response& response) override
+    {
+        UNREFERENCED_PARAMETER(response);
```
sid-dahiya, line 46: Personal Style Preference Nit:
Instead of marking as `UNREFERENCED_PARAMETER`, can we just comment the variable names out in line 44? Like `onHttpRequest(HttpServer::Request const& /*request*/, HttpServer::Response& /*response*/)`

tests/functests/MultipleLogManagersTests.cpp
```diff
@@ -36,13 +36,39 @@
...
+        count_++;
+        return 200;
+    }
+
+    size_t GetRequestCount() {
+        return count_;
+    }
+
+   private:
+    size_t count_;
```
sid-dahiya, line 59: Nit: Do we need the `_` suffix? Maybe just name these as `m_count` and `m_id` instead?

sid-dahiya (approved):

tests/functests/MultipleLogManagersTests.cpp
lalitb: Thanks, have commented the unused argument now.

tests/functests/MultipleLogManagersTests.cpp
lalitb: Yes, I got used to follow the Google Style guide for [variable naming](https://google.github.io/styleguide/cppguide.html?msclkid=6e031e1dd15111ec940468bd031e25d6#Variable_Names) - we follow this in OpenTelemetry.
```
Data members of classes, both static and non-static, are named like ordinary nonmember variables, but with a trailing underscore.
```

But realized we use `m_` for members in 1DS. fixed it now.

PR #1006, opened 2022-03-28
Fix build scripts to support CntOS8 and Windows build with VS2019 (any)

vgerasimovms:


reyang:
The change itself looks good to me. I have a question - do we even want to maintain this build flavor? It is not listed here https://github.com/microsoft/cpp_client_telemetry#build-environments.

vgerasimovms:
> The change itself looks good to me. I have a question - do we even want to maintain this build flavor? It is not listed here https://github.com/microsoft/cpp_client_telemetry#build-environments.

What do you mean by build flavor? CentOS8?

reyang:
> What do you mean by build flavor? CentOS8?

Yeah, I guess either we agree with the community AND add it to the supported list AND make it part of the CI; or remove it?

vgerasimovms:
> > What do you mean by build flavor? CentOS8?
> 
> Yeah, I guess either we agree with the community AND add it to the supported list AND make it part of the CI; or remove it?

Unfortunately, we need it, like many others. So, need to keep it

docker/centos8/Dockerfile
```diff
@@ -1,13 +1,17 @@
...
-RUN yum install -y curl libcurl-devel zlib-devel clang python38 pkg-config git bzip2 unzip make wget sudo cmake
-RUN yum --enablerepo=powertools install -y zlib-static gmock gtest
-
+FROM centos:centos8
+
+USER 0
+
+
+RUN sed -i 's/mirrorlist/#mirrorlist/g' /etc/yum.repos.d/CentOS-Linux-* &&\
+    sed -i 's|#baseurl=http://mirror.centos.org|baseurl=http://vault.centos.org|g' /etc/yum.repos.d/CentOS-Linux-*
```
lalitb, line 7: What is wrong with installing packages from `mirror`? Or do we need old/specific versions from `vault`

build-win.ps1
```diff
@@ -5,11 +5,10 @@ param (
   [string]$enableWin10 = "true",
   [string]$enableMini = "true",
   [string]$enableTests = "true",
-  [string]$customProps = ""
+  [string]$customProps = "",
+  [string]$vsDevCmdBat = "C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\Tools\VsDevCmd.bat"
```
sid-dahiya, line 9: If possible, use the `%ProgramFilesx86%` variable.

docker/centos8/Dockerfile
vgerasimovms: it failed to build the docker image - not sure what is the cause of issue

build-win.ps1
vgerasimovms: It is not easy because in build environment it could be empty, so logic could be complicated - not good for default parameter

docker/centos8/Dockerfile
lalitb: Ok probably because CentOS8 has reached end of lifecycle, and the repos are not more available, so need to switch to `vault` for archives. Should be fine.

lalitb (approved):

PR #1009, opened 2022-04-09
Clang static analysis fix for rand()

utsabms:
Calling platform specific UUID function for apple to fix rand()


tests/unittests/PalTests.cpp
```diff
@@ -1,370 +1,16 @@
...
 #include "common/Common.hpp"
-#include "common/MockISemanticContext.hpp"
+#include "pal/PseudoRandomGenerator.hpp"
+#include "Version.hpp"
 
 using namespace testing;
 
 class PalTests : public Test {};
 
-//---
```
sid-dahiya, line 14: Why are all these tests being removed?

tests/unittests/PalTests.cpp
```diff
@@ -1,370 +1,16 @@
-#if 0
 //
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 //
 
 #include "common/Common.hpp"
-#include "common/MockISemanticContext.hpp"
+#include "pal/PseudoRandomGenerator.hpp"
+#include "Version.hpp"
```
sid-dahiya, line 8: You shouldn't be using `Version.hpp` anywhere.

tests/unittests/PalTests.cpp
utsabms: These changes were never run. The whole file was commented out. I recovered some.

sid-dahiya: In that case, there should be an issue tracking the tests not running. as it means there is test coverage gap right now.

tests/unittests/PalTests.cpp
sid-dahiya: Is there a test validating SDK version? If so, this header is fine to have here. If the test is removed, you don't need this.

utsabms: Yep. The last one in PalTests.cpp. Thats why I added it.

tests/unittests/PalTests.cpp
utsabms: Testcases are commented out in lib\pal\PalTests.cpp #1010

utsabms: I have created a bug,

lalitb (changes requested):
Need more clarity on the UUID format - hex strings in upper case or lower case.

lalitb (approved):
Approved as discussed over teams - The format generated is similar to that for Windows so shouldn't break anything upstream. Should be good to publish once tests are passing.

raghavthind2005 (approved):

lib/pal/PAL.cpp
```diff
@@ -300,6 +304,32 @@ namespace PAL_NS_BEGIN {
...
+            uuidStr = CFStringGetCStringPtr(uuidStrRef, kCFStringEncodingASCII);
+        }
+        else
+        {
+            const auto uuidNullTerminatedSize {CFStringGetLength(uuidStrRef) + 1};
+            uuidStr.resize(uuidNullTerminatedSize);
+            CFStringGetCString(uuidStrRef, &uuidStr[0], uuidNullTerminatedSize, kCFStringEncodingASCII);
+        }
+        CFRelease(uuidStrRef);
+	std::transform(uuidStr.begin(), uuidStr.end(), uuidStr.begin(), ::tolower);
```
sid-dahiya, line 331: Might be the viewer, but the indentation looks weird. Mind double checking it?

sid-dahiya (approved):

PR #1013, opened 2022-05-01
Suppress false positive CodeQL errors

t-ransemel:
We noticed that in our mirrored branch codeQL raises some errors and after looking into them decided they were false positives:
1. "Dereference of potentially uninitialized pointer field" in HttpClient_CAPI.cpp - the shared_ptr operation is dereferenced within an if block that checks it isn't null, in addition it is default constructed with nullptr so in the case that it isn't null, it means it was purposely assigned a non garbage value. I added the lgtm comment with justification to suppress the error and changed the initialization of operation for better clarity on its initialized value.
link to the error:

2. "Alert Suppression Without Justification" in json.hpp - The function scan_number() uses goto which raised a codeQL error which was later suppressed with the comment "lgtm [cpp/use-of-goto]", I added justification for this suppression in order to comply with the new CodeQL suppression standard: https://onees.lgtm.microsoft.com/rules/1001259/
link to the error:

### DIFF
--- a/lib/http/HttpClient_CAPI.cpp
+++ b/lib/http/HttpClient_CAPI.cpp
@@ -199,7 +199,7 @@ namespace MAT_NS_BEGIN {
         
         if (operation != nullptr)
         {
-            operation->Cancel();// CodeQL [cpp/uninitializedptrfield] operation is explicitly constructed with with so it will never hold garbage value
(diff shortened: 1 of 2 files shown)


lalitb (approved):

t-ransemel:
@lalitb Can you please merge the PR? I seem to not have write access.

lalitb:
@t-ransemel - you can get write permission from here - https://microsoft.sharepoint.com/teams/1ds.sdk.cpp/SitePages/Home.aspx

PR #1016, opened 2022-05-25
Configuring option for force resetting the osx deployment target

sjb7:
**Issue**

Recently we did the integration of the 1DS library in a common xplat repo and built it for iOS. The common xplat repo contains 1DS as a submodule and adds it as a subdirectory in CMakeList. Post the integration and building it for iOS we observed that the minimum supported OS Version (LC_VERSION_MIN_IPHONEOS) is not getting set in the generated framework. Hence, the build is failing while submitting to test flight.

**RCA**

We observed that 1DS CMakeLists is forcefully setting the CMAKE_OSX_DEPLOYMENT_TARGET as an empty string which is causing the issue (`set(CMAKE_OSX_DEPLOYMENT_TARGET "" CACHE STRING "Force unset of the deployment target for iOS" FORCE)`).

**Solution**

Configuring an option that gives the flexibility on whether the module building 1DS wants this behavior or not.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,12 +25,17 @@ if(APPLE)
 
   # iOS build options
   option(BUILD_IOS           "Build for iOS"               NO)
+  option(FORCE_RESET_OSX_DEPLOYMENT_TARGET "Clear the OSX Deployment Target Set" ON)
 
   if(BUILD_IOS)
     set(TARGET_ARCH "APPLE")
     set(IOS True)
     set(APPLE True)
-    set(CMAKE_OSX_DEPLOYMENT_TARGET "" CACHE STRING "Force unset of the deployment target for iOS" FORCE)
+
+    if(FORCE_RESET_OSX_DEPLOYMENT_TARGET)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -25,12 +25,17 @@ if(APPLE)
 
   # iOS build options
   option(BUILD_IOS           "Build for iOS"               NO)
+  option(FORCE_RESET_OSX_DEPLOYMENT_TARGET "Clear the OSX Deployment Target Set" ON)
```
eduardo-camacho, line 28: Value should be YES, so by default current behavior prevails

eduardo-camacho (changes requested):
Value should be YES, so by default current behavior prevails

CMakeLists.txt
sjb7: Yes, the value ON is equivalent to YES. You can either use ON/OFF or you can use YES/NO.

sid-dahiya: The project standard is to use YES, please use that.

sjb7: I have update that to YES.

sid-dahiya (approved):

eduardo-camacho (approved):

PR #1021, opened 2022-06-09
Fixed the issue of configuration being shared between multiple loggers

nishchith-cp:
**Bug:**
The `ODWLogManager :+(nullable ODWLogger *)loggerWithTenant:(nonnull NSString *)tenantToken source:(nonnull NSString *)source withConfig:(nonnull ODWLogConfiguration *)config 
` always picks the most recent collector URI for all the loggers that have been created using this API.

****Issue Raised with all the details:**** - https://github.com/microsoft/cpp_client_telemetry/issues/991

**Reason:**
`static` configuration object was being used that was being shared between all the loggers created using the above API.

**Fix:** 
Removed the `static` and created an independent configuration instance.
Now the events are going through correct collector URL.
Screenshot 2022-06-06 at 1 35 28 PM


lalitb:
LGTM. would be good to get it reviewed with @eduardo-camacho before merging.

lalitb (approved):

eduardo-camacho (approved):
Thanks Nishchith for this change

PR #1042, opened 2022-07-25
Fix Mac Memory leak issue

shachau:
https://github.com/microsoft/cpp_client_telemetry_modules/issues/176

Multiple users are reporting high memory usage, especially on M1 Macs. We've made some significant gains in the last few months, but we are still running into users seeing 12+GB of usage. I've been in a thread with one such user who has provided leak traces

 The "with aria" traces were collected first, in OneDrive's default configuration. The user enabled malloc_history and let the product run until it reached around 1GB usage, at which point it is my understanding that we crashed (this has to do with malloc tracing and not another bug). After that, I had the user disable Aria, and run the same trace again (the "without aria" trace). No crash this time.

From a comparison of the traces, it is obvious that Aria is responsible for significant memory consumption.  What seems to be happening here (and this is consistent with when Apple looked at it) is that Aria is allocating objects inside an autoreleasepool that is never released. 

What changes:
Just added autorelease pool to functions that were not releasing memory.

### DIFF
--- a/lib/pal/WorkerThread.cpp
+++ b/lib/pal/WorkerThread.cpp
@@ -182,7 +182,7 @@ namespace PAL_NS_BEGIN {
             LOG_INFO("Running thread %u", std::this_thread::get_id());
 
             for (;;) {
-                #ifdef MacOS
+                #ifdef __APPLE__
                 @autoreleasepool
                 #endif
                 {
--- a/lib/pal/WorkerThread.cpp
+++ b/lib/pal/WorkerThread.cpp
@@ -182,74 +182,69 @@ namespace PAL_NS_BEGIN {
             LOG_INFO("Running thread %u", std::this_thread::get_id());
 
             for (;;) {
-                #ifdef __APPLE__
-                @autoreleasepool
-                #endif
+                std::unique_ptr<MAT::Task> item = nullptr;
+                wakeupCount++;
+                unsigned nextTimerInMs = MAX_FUTURE_DELTA_MS;
                 {
-                    std::unique_ptr<MAT::Task> item = nullptr;
-                    wakeupCount++;
-                    unsigned nextTimerInMs = MAX_FUTURE_DELTA_MS;
-                    {
-                        LOCKGUARD(self->m_lock);
+                    LOCKGUARD(self->m_lock);
 
-                        auto now = getMonotonicTimeMs();
-                        if (!self->m_timerQueue.empty()) {
-                            const auto currTargetTime = self->m_timerQueue.front()->TargetTime;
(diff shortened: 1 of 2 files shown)


shachau:
@sid-dahiya  @lalitb  for reference, this is the fix for memory leak issue on Mac that we saw in OneDrive. This fix worked for OneDrive so merging it to the common code.

lib/pal/WorkerThread.cpp
```diff
@@ -182,69 +182,74 @@ namespace PAL_NS_BEGIN {
             LOG_INFO("Running thread %u", std::this_thread::get_id());
 
             for (;;) {
-                std::unique_ptr<MAT::Task> item = nullptr;
-                wakeupCount++;
-                unsigned nextTimerInMs = MAX_FUTURE_DELTA_MS;
+                #ifdef MacOS
```
eduardo-camacho, line 185: This should be applicable to both Mac and iOS

sid-dahiya: `#ifdef APPLE` I believe. Otherwise, you can do or as well with `|` operator.

lalitb: `ifdef __APPLE__` should suffice for both Mac and iOS.

shachau: Fixed the macro

shachau:
Getting error. Why is 
/Users/runner/work/cpp_client_telemetry/cpp_client_telemetry/lib/pal/WorkerThread.cpp:186:17: error: expected expression
                @autoreleasepool

lalitb:
> @lalitb what compiler do we use? GCC or Clang compiler

`Clang` is used for both mac and iOS compilation. But looking into the PR now, it doesn't make sense to add `@autoreleasepool` in the C++ code. It is not a C++ construct and will fail with any C++ compiler. Why do we need to add it in the worker thread?

eduardo-camacho:
> > @lalitb what compiler do we use? GCC or Clang compiler
> 
> `Clang` is used for both mac and iOS compilation. But looking into the PR now, it doesn't make sense to add `@autoreleasepool` in the C++ code. It is not a C++ construct and will fail with any C++ compiler. Why do we need to add it in the worker thread?

Agree with your statement, this block do not contain Obj-C object references, hence @autoreleasepool doesn't apply here

shachau:
Thanks removed the autorelease pool from WorkerThread.cpp

lalitb (approved):
Thanks. LGTM

lalitb:
On lighter front, @autoreleasepool user can ignore all these discussions, he is unknowingly getting added here. :)

PR #1044, opened 2022-08-10
Added support for AAD device auth header

chkilpa1:


lalitb (approved):

lalitb:
Just to add, in separate discussion with Andrew Faust, the Collector is able to successfully parse the AAD device token passed using `AadDeviceToken` HTTP header. So, approving it based on that.

Issue #1049, opened 2022-09-01
Enable HTTP/2.0 support for the SDK.


lalitb:
1ds cpp sdk currently uses HTTP/1.1 to push events to the ingestion service. This issue is open to explore adding support for HTTP/2.0 in the library. HTTP/2.0 is in general more secure and performant as compared to HTTP/1.1.

1DS SDK uses different client libraries to add HTTP support depending on the platform it is built for. Had a quick validation on the HTTP/2.0 support for all the platforms

* Android - The default HTTP library part of Android SDK doesn't have HTTP/2.0 
support. This needs to be replaced with `okHTTP` to add support.
* Apple - HTTP/2.0 is supported from iOS9 onwards.
* WinINet - Fully supports HTTP/2.0
* libcurl - has an external dependency on  Nghttp2  to add the support. 

While there is no immediate plan to support this requirement, unless there is enough traction on this issue, and/or contributors are ready to add the support.

lalitb:
None of the collector URLs configured in 1DS SDK have HTTP/2.0 enabled.

But this can be tested in Japan west : onedscolprdjpw02.japanwest.cloudapp.azure.com

lalitb:
The support is added for:

 Posix/curl:  https://github.com/microsoft/cpp_client_telemetry/pull/1068 
 Android (allow adding different android http implementation) : https://github.com/microsoft/cpp_client_telemetry/pull/1067 ( enable plugging the okhttp3 package)
Windows and iOS is already supported by Native HTTP Libraries

Closing this issue. Will create a separate issue to add okhttp support as external module or addition package.

PR #1055, opened 2022-09-13
Draft - Android - allow supply a different request factory to HttpClient

anod:
Task #1049 Enable HTTP/2.0 support for the SDK.

- Add option to provide a different request factory to HttpClient
   - Add `HttpClientRequest` and `HttpClientRequest.Factory` interfaces
   - Supply a default `AndroidUrlConnection` factory
- Add a new `maesdk-okhttp` module with okhttp implementation to avoid pulling/managing external dependencies

image

TODO:
[  ] Test okhttp: parsing request, cancellation, http/2 connection are being reuse
[  ] Add unit test for okhttp
[  ] Publish okhttp module


lib/android_build/app/build.gradle
```diff
@@ -30,14 +30,19 @@ android {
...
+    compileOptions {
+        sourceCompatibility JavaVersion.VERSION_1_8
+        targetCompatibility JavaVersion.VERSION_1_8
+    }
 }
 
 dependencies {
     implementation fileTree(dir: 'libs', include: ['*.jar'])
     implementation project(':maesdk')
+    implementation project(':maesdk-okhttp')
```
sid-dahiya, line 42: What is `okhttp`?

sid-dahiya:
Why don't we have okhttp implementation be the default implementation? @lalitb ?

anod:
> Why don't we have okhttp implementation be the default implementation? @lalitb ?

need to cosider version management: default will require all clients to update to 4.10, and vise versa in case of new version of okhttp, the library needs to be updated...

sid-dahiya:
> > Why don't we have okhttp implementation be the default implementation? @lalitb ?
> 
> need to cosider version management: default will require all clients to update to 4.10, and vise versa in case of new version of okhttp, the library needs to be updated...

Got it, in that case, maybe this can be the default in the next major SDK version.

lib/android_build/app/build.gradle
```diff
@@ -30,14 +30,19 @@ android {
             version "3.10.2"
         }
     }
+    compileOptions {
+        sourceCompatibility JavaVersion.VERSION_1_8
+        targetCompatibility JavaVersion.VERSION_1_8
+    }
```
lalitb, line 36: Is this dependency coming for okhttp ? As I understand there are small number of android devices on java 6/7, the build will fail for those?

anod: latest android stuido uses java 11 by default, so it failed to compile, this tells it use 1.8 like in maesdk library

lalitb:
@sid-dahiya  Do you know if someone from office telemetry team with Android expertise (Saurabh?) can help review this code?

anod:
> > > Why don't we have okhttp implementation be the default implementation? @lalitb ?
> > 
> > 
> > need to cosider version management: default will require all clients to update to 4.10, and vise versa in case of new version of okhttp, the library needs to be updated...
> 
> Got it, in that case, maybe this can be the default in the next major SDK version.

Yes, this sounds like good plan to give users option to migrate separately and test the implementation, and then make it default

PR #1057, opened 2022-09-14
Fix memory Leaks on iOS

samirnen:
Observing significant leaks on iOS devices when using 1-DS to send telemetry. Making NSURLSession static instead of creating a new instance for every telemetry event resolved the issue of memory leakage. Attaching snapshots from instruments where memory usage is being tracked with a static NSURLSession and a non static one.

Screenshot 2022-09-14 at 14 44 15
Screenshot 2022-09-14 at 14 45 03


lalitb:
pasting here from the 1ds-sdk team's channel

--
Hi team.

We are observing significant memory spiking and leaks with 1-DS on iOS devices. I'm attaching a snapshot from the leaks profile taken from instruments. I have tested with the fixes made in this [Pull Request](https://github.com/microsoft/cpp_client_telemetry/pull/1042). As we can see in the snapshot the memory is spiking quite fast and leaking a significant amount as well. 
 
On further analysis we have found out that for every 1-DS telemetry event we send, we are creating a new URL session. Instead of which if a static URL session is used, the leaks and spiking aren't observed anymore. Attaching a screenshot of the same.
 
Wanted to know if there is any reason behind having unique URL sessions for each telemetry request and if making the URL session static might cause any issues. Here is a link to the [draft PR](https://github.com/microsoft/cpp_client_telemetry/pull/1057/files).
---

lalitb:
Adding @eduardo-camacho as reviewer, as he has written this part of iOS code.

lib/http/HttpClient_Apple.mm
```diff
@@ -50,9 +52,11 @@ void SendAsync(IHttpResponseCallback* callback)
         {
             m_callback = callback;
             NSString* url = [[NSString alloc] initWithUTF8String:m_url.c_str()];
-            NSURLSessionConfiguration* sessionConfig = [NSURLSessionConfiguration defaultSessionConfiguration];
             m_urlRequest = [[NSMutableURLRequest alloc] initWithURL:[NSURL URLWithString:url]];
-            m_session = [NSURLSession sessionWithConfiguration:sessionConfig];
+            if(m_session == nil) {
```
eduardo-camacho, line 56: This can lead to a raise condition

lib/http/HttpClient_Apple.mm
```diff
@@ -50,9 +52,11 @@ void SendAsync(IHttpResponseCallback* callback)
         {
             m_callback = callback;
             NSString* url = [[NSString alloc] initWithUTF8String:m_url.c_str()];
```
eduardo-camacho, line 54: If session is static and url changes overtime during session, this change would events sent to wrong endpoint

eduardo-camacho (changes requested):
I see concurrency and multi-url problems with this change, please take a look.

lib/http/HttpClient_Apple.mm
samirnen: As per the [documentation from apple](https://developer.apple.com/documentation/foundation/urlsession) 

> Your app creates one or more URLSession instances, each of which coordinates a group of related data-transfer tasks. For example, if you’re creating a web browser, your app might create one session per tab or window, or one session for interactive use and another for background downloads. Within each session, your app adds a series of tasks, each of which represents a request for a specific URL (following HTTP redirects, if necessary).

In the current implementation if the URL changes overtime, then a new URL Session Task would be created and even though if we have the same URL session, the events would still be routed to the correct endpoint because URL session can handle multiple URL Session Tasks

lib/http/HttpClient_Apple.mm
samirnen: This is a lazy initialisation. Moving the initialisation to the constructor of HttpRequestApple class.

lib/http/HttpClient_Apple.mm
```diff
@@ -37,6 +39,10 @@
         m_parent(parent)
     {
         m_parent->Add(static_cast<IHttpRequest*>(this));
+        if(m_session == nil) {
```
eduardo-camacho, line 42: I still see a potential race condition here, suggest using dispatch_once

eduardo-camacho (approved):

lib/http/HttpClient_Apple.mm
sid-dahiya: @samirnen can you confirm you have tested this scenario? Getting URL updates wrong will have severe implications for Privacy and Compliance of the data being exhausted.

lib/http/HttpClient_Apple.mm
```diff
@@ -50,10 +57,7 @@ void SendAsync(IHttpResponseCallback* callback)
         {
             m_callback = callback;
             NSString* url = [[NSString alloc] initWithUTF8String:m_url.c_str()];
-            NSURLSessionConfiguration* sessionConfig = [NSURLSessionConfiguration defaultSessionConfiguration];
             m_urlRequest = [[NSMutableURLRequest alloc] initWithURL:[NSURL URLWithString:url]];
-            m_session = [NSURLSession sessionWithConfiguration:sessionConfig];
-            m_session.sessionDescription = url;
```
lalitb, line 56: We don't need to set `sessionDescription` anymore?

lib/http/HttpClient_Apple.mm
```diff
@@ -29,6 +29,9 @@
     return std::string("RESP-") + std::to_string(seq.fetch_add(1));
 }
 
+static dispatch_once_t once;
+static NSURLSession* m_session = nil;
```
lalitb, line 33: Can't these static's be member variables of HttpRequestApple ?

lib/http/HttpClient_Apple.mm
samirnen: The session Description is just a string used for debugging purposes. Previously since we had created new URL session each time we made a network call, we set the description to the URL. But now since it is a static URL Session having a session description set to one URL won't be of much use. Hence removed it.

Removing it will not cause any issues to the workflow. Apple documentation for the same: https://developer.apple.com/documentation/foundation/nsurlsession/1408277-sessiondescription

lib/http/HttpClient_Apple.mm
samirnen: Yes I have tried to do the same. The HttpRequestApple class is defined in .mm file, I was getting a compilation error if I defined the static objects in .mm. A workaround for this is to define the HttpRequestApple class inside .hh header and define the static variables in the header itself. However that exposes the whole HttpRequestApple class to files which import the header. Hence I have created static objects outside of the class.

lalitb: Can't we make it split it in the cc/h file, and ensure the source file is linked when building for iOS?

lib/http/HttpClient_Apple.mm
lalitb: > Is there a way that I can test this out

I don't think this use-case is valid as of now  - as we can't change the endpoint URL once SDK is bootstrapped.

lib/http/HttpClient_Apple.mm
eduardo-camacho: > Can't these static's be member variables of HttpRequestApple ?

Indeed, this static variable could definitively be moved to the function itself as a local static.

eduardo-camacho (changes requested):
static variable could definitively be moved to the function itself as a local static

lib/http/HttpClient_Apple.mm
samirnen: Okay. In that case having a single URL session should not create any issues right? Even if the URL can be changed, it should not cause issues (according to Apple documentation) but that flow cannot be tested right now.

lalitb: yes that's correct.

lib/http/HttpClient_Apple.mm
lalitb: @samirnen  - Do you think this is doable?

samirnen: I tried moving the variable to the function as a local static. But then the memory is leaking same as before. So didn't go ahead with that approach

samirnen: I have a create a new source file for the HTTPRequestApple class and have defined the static variables there

samirnen:
Hey @eduardo-camacho @lalitb can you please re review this PR

lib/http/HttpRequestApple.mm
```diff
@@ -0,0 +1,153 @@
...
+#import <CFNetwork/CFNetwork.h>
+
+#include "HttpClient_Apple.hpp"
+#include "HttpRequestApple.hpp"
+#include "utils/StringUtils.hpp"
+
+namespace MAT_NS_BEGIN {
+
+static dispatch_once_t once;
+static NSURLSession* m_session;
```
lalitb, line 21: nit - `m_` prefix is used for class members, can we name it without the prefix?

lib/http/HttpRequestApple.mm
```diff
@@ -0,0 +1,153 @@
...
+#import <Foundation/Foundation.h>
+#import <CFNetwork/CFNetwork.h>
+
+#include "HttpClient_Apple.hpp"
+#include "HttpRequestApple.hpp"
+#include "utils/StringUtils.hpp"
+
+namespace MAT_NS_BEGIN {
+
+static dispatch_once_t once;
```
lalitb, line 20: These are still global statics, but I understand there are challenges in making it member-variable for this class - https://stackoverflow.com/questions/13856037/can-i-declare-dispatch-once-t-predicate-as-a-member-variable-instead-of-static/13858628#13858628. Also, there are issues in making it static member of the class. Will approve it if it is not practically possible to do either.
Also I understand end to end tests have been done - events are flowing, and there is no memory leak observed in this part of code.

lalitb (approved):

lib/http/HttpRequestApple.hpp
```diff
@@ -0,0 +1,34 @@
+//
```
eduardo-camacho, line 1: This file should be named HttpRequest_Apple.http (similar case with new .mm file) to match current platform specific naming scheme.

eduardo-camacho (changes requested):
Effectively this change is more disruptive than previous version when Lalit commented about the global static, and the global static thing is still present; am I missing something here? If we are going to allow the global statics in this case, then it's just cleaner how it was before splitting files. Would you agree @lalitb?

lalitb:
> Effectively this change is more disruptive than previous version when Lalit commented about the global static, and the global static thing is still present; am I missing something here? If we are going to allow the global statics in this case, then it's just cleaner how it was before splitting files. Would you agree @lalitb?

Yes, I too feel so, the initial approach was much cleaner if we can't get rid of global statics with this approach. Sorry @samirnen if there are too many iterations :)

eduardo-camacho (approved):

PR #1066, opened 2022-10-31
Add Distributed Tracing semantics

maxgolov:
Add Distributed Tracing semantics to the following APIs:
- `ISemanticContext` - allows to set context on all events emitted by `ILogger`
- ability to use `env_dt` aliases to customize TraceContext of individual events running in a given scope
- semantic context projection for C++/CL/CX apps on Windows
- add test that emits two events with DT semantics

PR adds "common" fields:
- `env_dt_traceId`
- `env_dt_spanId`
- `env_dt_traceFlags`
- Part B `parentId`

And the set of corresponding high-level API setters at context-level:
- SetTraceId
- SetSpanId
- SetTraceFlags
- SetParentId
which is styled in the same way as other common context variables.

How to use it:
- events land in Aria-Kusto or Kusto and could be subsequently routed to Geneva via 1DS-Interchange.
- events landing in Geneva have the required `ext.dt` extension semantics, just enough for the Trace Explorer view to work.. Less duration semantics: since 1DS in principle does not require client events to have "duration", start / end / scope. Events are timepoints that carry `timestamp` and `name`. How exactly to remap these, i.e. what would/could/should be considered duration - is out-of-scope of this PR.

How this could be improved further:
- There's no helper for formatting TraceContext. Users may use whatever formatter of their choice, e.g. taking a dependency on OpenTelemetry API for well-formed context of [trace parent header values](https://www.w3.org/TR/trace-context/). In our org scenario we are utilizing existing .NET APIs that already give us the access to values in the right format, so calling into 1DS C++ SDK via projection is trivial. We do not require manual formatting of these values in C++ land.
- Should someone need to learn how to do this formatting in C++ - please refer to the [OpenTelemetry C++ code here](https://github.com/open-telemetry/opentelemetry-cpp/tree/main/api/include/opentelemetry/trace), that illustrates how to format TraceId and SpanId to hexadecimal string values.

Example showing how events look when traced from client to service in one Kusto table:
image

Disregard the column names in the above screenshot, the proper long-term names compatible with our current Distributed Tracing explorer are those mentioned above (`env_dt_*` and Part B `parentId`).

`APITest` verifies the basic functionality. More complex usage examples are not presently a priority for our team, since in our case we use the higher-level .NET `StartActivity` API while interacting with remote service instrumented with OpenTelemetry, and supplement the existing client telemetry events with correlation identifiers that follow Common Schema naming, specifically Common Schema -to- Geneva mapping.

### DIFF
--- a/lib/include/public/CommonFields.h
+++ b/lib/include/public/CommonFields.h
@@ -24,6 +24,15 @@
 #define COMMONFIELDS_DEVICE_CLASS                            "DeviceInfo.Class"
 #define COMMONFIELDS_DEVICE_ORGID                            "DeviceInfo.OrgId"
 
+/* Ref. https://www.w3.org/TR/trace-context/#traceparent-header-field-values
+ - ParentId  - parent of the current span received via remote context propagation.
+ - SpanId    - current SpanId sent in 'traceparent' to remote service.
+ */
+#define COMMONFIELDS_DT_TRACEID                              "env_dt_traceId"
+#define COMMONFIELDS_DT_SPANID                               "env_dt_spanId"
+#define COMMONFIELDS_DT_TRACEFLAGS                           "env_dt_traceFlags"
+#define COMMONFIELDS_DT_PARENTID                             "parentId"
+
 #define COMMONFIELDS_NETWORK_PROVIDER                        "DeviceInfo.NetworkProvider"
 #define COMMONFIELDS_NETWORK_TYPE                            "DeviceInfo.NetworkType"
 #define COMMONFIELDS_NETWORK_COST                            "DeviceInfo.NetworkCost"
--- a/lib/include/public/ISemanticContext.hpp
+++ b/lib/include/public/ISemanticContext.hpp
@@ -109,11 +109,38 @@ namespace MAT_NS_BEGIN
         /// <param name="deviceClass">Device class.</param>
         DECLARE_COMMONFIELD(DeviceClass, COMMONFIELDS_DEVICE_CLASS);
 
-          /// <summary>
+        /// <summary>
         /// Set the device orgId context information of telemetry event.
         /// </summary>
         /// <param name="deviceClass">Device orgId</param>
         DECLARE_COMMONFIELD(DeviceOrgId, COMMONFIELDS_DEVICE_ORGID);
+
+        /// <summary>
+        /// Set the W3C TraceContext TraceId information of telemetry event.
+        /// </summary>
+        /// <param name="traceId">TraceContext TraceId</param>
+        DECLARE_COMMONFIELD(TraceId, COMMONFIELDS_DT_TRACEID);
+
+        /// <summary>
+        /// Set the W3C TraceContext SpanId information of telemetry event.
+        /// </summary>
+        /// <param name="spanId">TraceContext SpanId</param>
+        DECLARE_COMMONFIELD(SpanId, COMMONFIELDS_DT_SPANID);
+
+        /// <summary>
+        /// Set the W3C TraceContext TraceFlags information of telemetry event.
+        /// </summary>
+        /// <param name="TraceFlags">TraceContext TraceFlags</param>
+        virtual void SetTraceFlags(uint8_t traceFlags)
+        {
+            SetCommonField(COMMONFIELDS_DT_TRACEFLAGS, traceFlags);
+        }
+
+        /// <summary>
+        /// Set the remote context (parent) SpanId information of telemetry event.
+        /// </summary>
+        /// <param name="ParentId">ParentId</param>
+        DECLARE_COMMONFIELD(ParentId, COMMONFIELDS_DT_PARENTID);
 
         /// <summary>
         /// Set the network cost context information of telemetry event.
--- a/lib/api/ContextFieldsProvider.cpp
+++ b/lib/api/ContextFieldsProvider.cpp
@@ -244,6 +244,70 @@ namespace MAT_NS_BEGIN
                     record.extDevice[0].deviceClass = iter->second.as_string;
                 }
 
+                /* Distributed tracing fields.
+                 * 
+                 * Part A extension:
+                 * - ext.dt.traceId    -> env_dt_traceId
+                 * - ext.dt.spanId     -> env_dt_spanId
+                 * - ext.dt.traceFlags -> env_dt_traceFlags
+                 *
+                 * Part B mapping:
+                 * - parentId          -> parentId
+                 */
+                iter = m_commonContextFields.find(COMMONFIELDS_DT_TRACEID);
+
+                /* Only run the context mapping if TraceId field is set */
+                if (iter != m_commonContextFields.end())
+                {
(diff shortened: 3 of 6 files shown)


lib/include/public/ISemanticContext.hpp
```diff
@@ -109,12 +109,39 @@ namespace MAT_NS_BEGIN
...
+        /// Set the W3C TraceContext SpanId information of telemetry event.
+        /// </summary>
+        /// <param name="spanId">TraceContext SpanId</param>
+        DECLARE_COMMONFIELD(SpanId, COMMONFIELDS_DT_SPANID);
+
+        /// <summary>
+        /// Set the W3C TraceContext TraceFlags information of telemetry event.
+        /// </summary>
+        /// <param name="TraceFlags">TraceContext TraceFlags</param>
+        virtual void SetTraceFlags(uint8_t traceFlags)
```
fibann, line 134: nit: the current spec only supports one flag (`sampled`) so I wonder if we should rather:

- add a `SetSampled(bool)` method, or
- change this to take a bitfield, or
- provide an enum or a TRACEFLAGS_SAMPLED constant to use here instead of '1'

(the last is probably best since it doesn't constrain the caller)

lib/include/public/CommonFields.h
```diff
@@ -24,6 +24,15 @@
...
 #define COMMONFIELDS_DEVICE_ORGID                            "DeviceInfo.OrgId"
 
+/* Ref. https://www.w3.org/TR/trace-context/#traceparent-header-field-values
+ - ParentId  - parent of the current span received via remote context propagation.
+ - SpanId    - current SpanId sent in 'traceparent' to remote service.
+ */
+#define COMMONFIELDS_DT_TRACEID                              "env_dt_traceId"
+#define COMMONFIELDS_DT_SPANID                               "env_dt_spanId"
+#define COMMONFIELDS_DT_TRACEFLAGS                           "env_dt_traceFlags"
+#define COMMONFIELDS_DT_PARENTID                             "parentId"
```
fibann, line 34: My understanding is that `parentId` is meant to be used in events with Span semantics. Here we are adding it to every single event instead, which looks a bit weird. Are we doing this because supporting explicit Span semantics is more complicated? Would it cause problems if we decide to support explicit Span semantics later?

lib/include/public/ISemanticContext.hpp
maxgolov: W3C spec and Common Schema support `traceFlags`:

image

An integer representation of the [W3C TraceContext trace-flags](https://www.w3.org/TR/trace-context/#trace-flags).

lib/include/public/CommonFields.h
maxgolov: There are two separate logical destinations here:

- first three: **traceId**, **spanId**, **traceFlags** - is our current span, what we would've sent in `traceparent` header.

- last one: **parentId** - is what we receive as our parent, what initiated us to run, what span is our parent. Developer focusing on client telemetry would receive this not as part of `traceparent` sent to the remote end, but as part of telemetry that their own code is emitting. That way we stitch together the caller (our parent) and our call into tree view.. Without this information attached on event - the Trace View explorer UX is not going to stitch the two operations on one time ladder chart.

So `spanid` -- is our current span id, `parentId` -- is our parent remote span id.

maxgolov: >  Here we are adding it to every single event instead

Not every single event. Only for events on specific component logger, that are relevant to specific service interaction. However, if a client controls the initial seed of `traceId`, then the same `traceId` could be set to different loggers - `serviceA-REST-Logger`, `serviceB-REST-Logger`, `serviceC-REST-Logger`. Since they have share the same `traceId`, they all show up in one call tree.

The second example in `APITest` shows how you can selectively append the context directly on just one event, without using `SetContext`.. so both ways are possible, context - all events for specific logger, and local - just one event.

lib/include/public/ISemanticContext.hpp
fibann: Yes I am saying that we could have
```cpp
struct TraceFlags {
  unsigned char sampled: 1,
  unsigned char reserved: 7
};
...
virtual void SetTraceFlags(TraceFlags traceFlags)
...
TraceFlags flags;
flags.sampled = 1;
context.SetTraceFlags(flags);
```
or rather
```cpp
enum TraceFlags {
  Sampled: 1,
};
...
context.SetTraceFlags(TraceFlags::Sampled);
``` 
rather than passing a raw uint8_t.

maxgolov: We could later on add any additional method overloads we want. Indeed, we could add yet another method what you're proposing. I would rather go with an additional method that just does `SetTraceContext`. My goal was to expose the bare min set of flags, as per spec:

- in Common Schema spec this is `int`
- in W3C TraceContext spec this described as `byte` / `uint8_t`

Thus, I followed the W3C TraceContext guidelines in this regard. You can see that other implementations, e.g. [Elastic Node Traceparent package](https://github.com/elastic/node-traceparent#traceparentflags) - did not elaborate into what flags could mean exactly. It's just flags number, no more no less than what the transport layer is expecting to get.

We can add more decoration later. It's add-ons, not the necessity.

fibann: Ok sounds good not to over-specify.

lib/include/public/CommonFields.h
fibann: Discussed more offline. Having a `parentId` for an event is a valid model even if the system doesn't follow the span model (see also [comment in TraceContext standard](https://github.com/w3c/trace-context/blob/main/http_header_format_rationale.md#spanparent-nomenclature)). Supporting explicit Span semantics is sort of overlapping with what OpenTelemetry does and out of scope for now.

fibann (approved):

lalitb:
Most of the attributes/fields defined in the `ISemanticContext` interface are not going to change after bootstrapping of SDK.  Does Distributed Tracing attributes fit in this category - as trace_id/ span_id could change based on the context where Logger::LogEvent() is called. 

Should we instead have an overloaded `Logger::LogEvent` method which accepts trace context. Something similar to how it is done in OpenTelemetry logging api - https://github.com/open-telemetry/opentelemetry-cpp/blob/389b84f4303f728c97dcf8194f76277488e18cf4/api/include/opentelemetry/logs/logger.h#LL59C3-L59C3

lalitb:
@maxgolov - Do you plan to address the concerns raised in this PR, and make it ready for merge. Thanks.

maxgolov:
@lalitb 
I'd like to address some comments from reviewers, i.e.

- adding a method that allows to set both TraceId+SpanId at once
- possibly exposing a method like this to higher-level apps (I need it to work via C API too)

Apologies for keeping it open for so long. It's coming soon.

lalitb:
> are there any plans for this integration?

@maxgolov can confirm, but I believe we are waiting for the comments to be resolved.

PR #1067, opened 2022-11-02
Android: Allow to provide a different HttpClient implementation

anod:
Subset of changes from  #1055 to avoid pulling 3rd party library dependencies into SDK

Example of OkHttp implementation which enablex HTTP/2 support

https://github.com/microsoft/cpp_client_telemetry/blob/b6f41f6e9b974c24902b144ebf271e57ec66b2c5/lib/android_build/maesdk-okhttp/src/main/java/com/microsoft/applications/events/okhttp/OkHttpRequest.java


anod:
@lalitb @sid-dahiya I would like to make Android HTTP client replaceable. Any concerns, suggestions?

This will allow us to use our own OkHttp implementation which supports HTTP/2 for AI integration.

lib/android_build/app/build.gradle
```diff
@@ -30,6 +30,10 @@ android {
             version "3.10.2"
         }
     }
+    compileOptions {
+        sourceCompatibility JavaVersion.VERSION_1_8
+        targetCompatibility JavaVersion.VERSION_1_8
+    }
```
lalitb, line 36: Will this means minimum JAVA SE 8 would be required to build 1ds-sdk for android, even if okhttp is not used as HTTPClient ?

lalitb:
@anod -  I am not sure how much different are APIs for OkHttp and (existing) HttpURLConnection, but do you think we can introduce HTTPClient for OkHttp as a module similar to other HTTP Clients [here](https://github.com/microsoft/cpp_client_telemetry/tree/main/lib/http). So there would be two differnt HTTPClients for Android, and right one would be picked through [HTTPClientFactory ](https://github.com/microsoft/cpp_client_telemetry/blob/main/lib/http/HttpClientFactory.cpp) class as below:

```
#elif defined(ANRDOID AND WITH_HTTP2)
    std::shared_ptr<IHttpClient> HttpClientFactory::Create() {
        LOG_TRACE("Creating HttpClient_Android");
        return HttpClient_OkHttp_Android::GetClientInstance();
    }
#elif defined(ANDROID)
    std::shared_ptr<IHttpClient> HttpClientFactory::Create() {
        LOG_TRACE("Creating HttpClient_Android");
        return HttpClient_Android::GetClientInstance();
    }
```

lib/android_build/app/build.gradle
anod: Current Android Studio uses Java 11, this tells it to use 1.8 which is what it was built with

anod:
> @anod - I am not sure how much different are APIs for OkHttp and (existing) HttpURLConnection, but do you think we can introduce HTTPClient for OkHttp as a module similar to other HTTP Clients [here](https://github.com/microsoft/cpp_client_telemetry/tree/main/lib/http). So there would be two differnt HTTPClients for Android, and right one would be picked through [HTTPClientFactory ](https://github.com/microsoft/cpp_client_telemetry/blob/main/lib/http/HttpClientFactory.cpp) class as below:
> 
> ```
> #elif defined(ANRDOID AND WITH_HTTP2)
>     std::shared_ptr<IHttpClient> HttpClientFactory::Create() {
>         LOG_TRACE("Creating HttpClient_Android");
>         return HttpClient_OkHttp_Android::GetClientInstance();
>     }
> #elif defined(ANDROID)
>     std::shared_ptr<IHttpClient> HttpClientFactory::Create() {
>         LOG_TRACE("Creating HttpClient_Android");
>         return HttpClient_Android::GetClientInstance();
>     }
> ```

The problem is including okhttp as dependency. This will require client app to use same or similar version of okhttp
You can see the examples in the [following article](https://jeroenmols.com/blog/2020/11/11/library-dependencies/)

Usually libraries provide different implementations as a separate libraries, so the client can decide which implementation to use.  Example from retrofit:
image

I can followup with adding a `maesdk-okhttp` module once I test it and confirm the implementation.

lib/android_build/maesdk/src/main/java/com/microsoft/applications/events/HttpClient.java
```diff
@@ -376,7 +394,8 @@ public FutureTask<Boolean> createTask(
       int[] header_index,
       byte[] header_buffer) {
     try {
-      Request r = new Request(this, url, method, body, request_id, header_index, header_buffer);
+      Map<String, String> headers = HttpClient.convertHeaders(header_index, header_buffer);
```
lalitb, line 397: Can we avoid copying request data from byte buffer to map, as this adds an performance overhead(though small) in the hot path of HTTP request processing ?

lalitb:
Sorry for late reply on this. The changes look good in general if we can avoid the intermediate buffer to store the request data. 
Tagging @sid-dahiya , @mkoscumb in they want to look into these changes.

lalitb:
@anod - Just wondering if you would like to resolve and go ahead with this PR ?

anod:
> @anod - Just wondering if you would like to resolve and go ahead with this PR ?

Yes, it will improve performance for us

lalitb (approved):
LGTM. Thanks for the changes. We can keep for couple of days for @microsoft/1ds-c-sdk-approver to review this, and merge end of the week.

PR #1068, opened 2022-11-08
Add HTTP/2 support to HTTPClient_Curl

lalitb:
Small change to add HTTP/2 support to `HTTPClient_curl`. It will fallback to HTTP/1.1 if either of client or server doesn't support HTTP/2.


lib/http/HttpClient_Curl.hpp
```diff
@@ -128,6 +128,8 @@ class CurlHttpOperation {
         // TODO: expose SSL cert verification opts via ILogConfiguration
         curl_easy_setopt(curl, CURLOPT_SSL_VERIFYPEER, 0);      // 1L
         curl_easy_setopt(curl, CURLOPT_SSL_VERIFYHOST, 0);      // 2L
+        // HTTP/2 please, fallback to HTTP/1.1 if not supported
+        curl_easy_setopt(curl, CURLOPT_HTTP_VERSION, CURL_HTTP_VERSION_2_0);
```
ThomsonTan, line 132: The new line of API call says using version 2.0, will it fallback automatically, or any extra parameter needed?

lalitb: Yes, it will fallback automatically to using HTTP/1.1 if 
 - curl is built without the nghttp2 library.
 - the collector is not supporting HTTP/2.0

ThomsonTan (approved):

PR #1070, opened 2022-11-08
Hide misleading exception

ms-iankudinova:
At the moment test server produces
```
Exception: System.IO.EndOfStreamException: Unexpected end of stream reached.
         at Bond.Protocols.Throw.EndOfStreamException()
         at Bond.IO.Safe.InputBuffer.EndOfStream(Int32 count)
         at Bond.IO.Safe.InputBuffer.ReadUInt8()
         at lambda_method(Closure , CompactBinaryReader`1 )
         at Bond.Deserializer`1.Deserialize[T](R reader)
         at Bond.Deserialize`1.From[R](R reader)
         at CommonSchema.Decoder.FromBondToJSONList(Boolean outputCompactJson, Boolean flatExt)
```

Even though [Decoder.cs](https://github.com/microsoft/cpp_client_telemetry/blob/7f36d8933e621f90513f69d78a40529fcc47a088/tools/decoder/Decoder.cs#L33) successfully read data and send successful response.

As I don't know better way to handle / detect that end of stream is reached and it seems that 1DS Test Server can keep processing request successfully, let's handle and log message for this exception (not stack trace). Then user won't see stacktrace and won't think that there is any error problem with Request or 1DS Test Server

### DIFF
--- a/tools/decoder/Decoder.cs
+++ b/tools/decoder/Decoder.cs
@@ -26,6 +26,7 @@ using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using CsProtocol;
 using System.Linq;
+using Fiddler;
 
 namespace CommonSchema
 {
@@ -324,6 +325,10 @@ namespace CommonSchema
                         jsonList.Add(j);
                         outputBuffer.SetLength(0);
                     } while (true);
+                }
+                catch (EndOfStreamException)
+                {
+                    Logger.LogDebug("End of Binary Stream");
                 }
                 catch (Exception ex)
                 {


tools/decoder/Decoder.cs
```diff
@@ -325,6 +326,10 @@ public List<string> FromBondToJSONList(bool outputCompactJson = true, bool flatE
                         outputBuffer.SetLength(0);
                     } while (true);
                 }
+                catch (EndOfStreamException)
```
lalitb, line 329: Is the exception happening while reading the event from input buffer, and there is nothing to be read ? Just wondering if there is better way to avoid this exception.

ms-iankudinova: Because of infinity loop, after data was already read, it is tried to continue and read data from input buffer, but there is nothing to read.
Unfortunately, I couldn't find a way to check that there is nothing to read, and loop should finish.
Open to hear better solution :)

lalitb: I don't know a better way either, but giving it a thought, won't worry much as the decoder component is only used with test-server/fiddler. Should be fine I guess.

lalitb (approved):

PR #1072, opened 2022-11-16
Handle exception thrown PlatformEventHandler deconstructor

j388923r:
Catch exception when accessing Application::Current in WindowsRTPlatformEventHandler deconstructor like in the constructor as both have the same risk of throwing the exception.

Manually tested that it fixes the issue where the exception was thrown on the first deconstruction of a PlatformEventHandler on Windows.


lalitb (approved):

lalitb:
@j388923r - Would you like to merge the changes?

j388923r:
@lalitb Yes, I don't think I have write access to merge, so if you know who could do it for me, that would be great.

PR #1084, opened 2023-01-13
use `find_packages` in cmake files to for shared lib build

yulin-li:


lalitb:
@yulin-li The build script already supports building shared library for 1ds as [here](https://github.com/microsoft/cpp_client_telemetry/blob/974c581854d4f6d72a3f889b5520e0e08ba96d86/build.sh#L49).

So, something like this should work:

```bash
$ export CMAKE_OPTS='-DBUILD_SHARED_LIBS=OFF'
$ ./build-tests.sh release
```

yulin-li:
> @yulin-li The build script already supports building shared library for 1ds as [here](https://github.com/microsoft/cpp_client_telemetry/blob/974c581854d4f6d72a3f889b5520e0e08ba96d86/build.sh#L49).
> 
> So, something like this should work:
> 
> ```shell
> $ export CMAKE_OPTS='-DBUILD_SHARED_LIBS=ON'
> $ ./build-tests.sh release
> ```

oh thanks, let me revert the changes in `build-test.sh`. I think it's still worth to keep the `find_package` updates.

lalitb:
> oh thanks, let me revert the changes in `build-test.sh`. I think it's still worth to keep the `find_package` updates.

This  would still be a problem if we install sqlite3 from (say) apt-get package manager. As sqlite3" or "sqlite3-dev" package installed through it doesn't seem to included a CMake package config file. 

```bash
# dpkg-query -L  libsqlite3-dev:amd64
/.
/usr
/usr/include
/usr/include/sqlite3.h
/usr/include/sqlite3ext.h
/usr/lib
/usr/lib/x86_64-linux-gnu
/usr/lib/x86_64-linux-gnu/libsqlite3.a
/usr/lib/x86_64-linux-gnu/libsqlite3.la
/usr/lib/x86_64-linux-gnu/pkgconfig
/usr/lib/x86_64-linux-gnu/pkgconfig/sqlite3.pc
/usr/share
/usr/share/doc
/usr/share/doc/libsqlite3-dev
/usr/share/doc/libsqlite3-dev/copyright
/usr/lib/x86_64-linux-gnu/libsqlite3.so
/usr/share/doc/libsqlite3-dev/changelog.Debian.gz

# dpkg-query -L   libsqlite3-0:amd64
/.
/usr
/usr/lib
/usr/lib/x86_64-linux-gnu
/usr/lib/x86_64-linux-gnu/libsqlite3.so.0.8.6
/usr/share
/usr/share/doc
/usr/share/doc/libsqlite3-0
/usr/share/doc/libsqlite3-0/README.Debian
/usr/share/doc/libsqlite3-0/changelog.Debian.gz
/usr/share/doc/libsqlite3-0/copyright
/usr/lib/x86_64-linux-gnu/libsqlite3.so.0
```

lalitb:
@yulin-li - let us know if the PR is still valid ?

yulin-li:
> ```shell
> CMAKE_OPTS='-DBUILD_SHARED_LIBS=ON'
> ```

sorry for this very late reply. The camke configurations is actually provided by cmake (https://github.com/Kitware/CMake/blob/master/Modules/FindSQLite3.cmake), not sqlite3 package.

The problem without `find_packages` is, we cannot use sqlite3 installed by package managers, we need to install a duplicated one in `/usr/local/lib/`

yulin-li:
> **[lalitb](/lalitb) ** c

@lalitb this PR is still valid, which allows us to use sqlite3 libs installed by package managers.

yulin-li:
@ThomsonTan I saw you were updating this PR. Could it be reviewed and merge? So we can avoid patching it in our project.

ThomsonTan:
[@yulin-li](https://github.com/yulin-li), is this PR still valid? Does the consuming of local install sqlite3 break other users who may not epect the local install?

yulin-li:
> [@yulin-li](https://github.com/yulin-li), is this PR still valid? Does the consuming of local install sqlite3 break other users who may not epect the local install?

This is still valid. The local sqlite is only required if the user wants to build a shared lib. This won't change the behavior for static lib build.

lib/CMakeLists.txt
```diff
@@ -276,7 +276,9 @@ if(BUILD_SHARED_LIBS STREQUAL "ON")
     # Prefer shared libraries for sqlite3 and zlib
     add_library(sqlite3 SHARED IMPORTED GLOBAL)
     add_library(z SHARED IMPORTED GLOBAL)
-    target_link_libraries(mat PUBLIC sqlite3 PUBLIC z ${LIBS} "${CMAKE_THREAD_LIBS_INIT}" "${CMAKE_DL_LIBS}" "${CMAKE_REQUIRED_LIBRARIES}")
+    find_package(SQLite3 REQUIRED)
```
ThomsonTan, line 279: > find_package(SQLite3 REQUIRED) [](http://example.com/codeflow?start=3&length=31)

Is there any scenario where the call to `find_package(SQLite3 REQUIRED)` is required? Assume the current hard-coded name still works, right?

yulin-li: yes, the current hard-coded name still works. But I think the `find_package` is a preferred way by cmake, and it will report a clearer error if `sqlite3` is not installed

bmehta001:
Thanks, @yulin-li, for contributing this PR. I have since made some efforts to improve the CMake/build files and am now using find_packages for all packages, including SQLite, as of #1478 as well, so this PR has been subsumed. Closing. Please let me know if you have any questions.

PR #1089, opened 2023-01-20
[WIP] Modern lifecycle + Ownership

lalitb:
Included changes
 * Deprecate shared global state (the LogManager instance that is not really an instance).

### DIFF
--- a/lib/include/public/LogManagerBase.hpp
+++ b/lib/include/public/LogManagerBase.hpp
@@ -170,7 +170,7 @@ namespace MAT_NS_BEGIN
         /// <summary>
         /// Concrete instance for servicing all singleton calls
         /// </summary>
-        static ILogManager* instance;
+        static std::unique_ptr<ILogManager> instance;
 
         /// <summary>
         /// Debug event source associated with this singleton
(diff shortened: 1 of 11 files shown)


lib/api/LogManagerImpl.cpp
```diff
@@ -629,7 +629,7 @@ namespace MAT_NS_BEGIN
         return m_logConfiguration;
     }
 
-    ILogger* LogManagerImpl::GetLogger(const std::string& tenantToken, const std::string& source, const std::string& scope)
+    std::shared_ptr<ILogger> LogManagerImpl::GetLogger(const std::string& tenantToken, const std::string& source, const std::string& scope)
     {
         {
             LOCKGUARD(m_lock);
```
absaroj, line 635: testing if i have access

lib/include/public/LogManagerBase.hpp
```diff
@@ -731,7 +731,7 @@ namespace MAT_NS_BEGIN
         static ILogManager* GetInstance() noexcept
```
mkoscumb, line 731: This probably needs to go as well

mkoscumb: Really all the static methods

lalitb: Yes, I am trying to see if we can make LogManager as instance of LogManagerBase, and then make all these methods as non-static. Though LogConfiguration is still a concern here, as it is difficult to create deep copy of it (as it can contain `void *` as the config value).

PR #1097, opened 2023-02-08
Fixing URLs for Collector

aef123:
The current default URLs for Collector the SDK is using are not the preferred URLs. We no longer want teams to use EMEA, NOAM or APAC URLs, because those are geographic boundaries and not data boundaries. We also don't want the "self" URL to be used, since that was never intended to be public. That was intended for internal use. 

The preferred URLs are:

mobile.events.data.microsoft.com
eu-mobile.events.data.microsoft.com
us-mobile.events.data.microsoft.com
au-mobile.events.data.microsoft.com
jp-mobile.events.data.microsoft.com

..

All future geo URLs will follow this same pattern.

### DIFF
--- a/Solutions/win32-cs/Program.cs
+++ b/Solutions/win32-cs/Program.cs
@@ -54,7 +54,7 @@ namespace CLI
                 AutoLogAppSuspend = false,
                 AutoLogUnhandledException = false,
                 OfflineStorage = "offline.storage",
-                // CollectorURL = "https://pipe.int.trafficmanager.net/Collector/3.0/",	// INT collector example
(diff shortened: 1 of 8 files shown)


lalitb (approved):

aef123:
@lalitb, @reyang, I don’t have write permissions so can’t merge this PR. Could either one of you merge it, or grant me permission and I’ll merge it?

reyang:
@aef123 I just merged the PR. If you want write access, join this group https://github.com/orgs/microsoft/teams/client-telemetry-sdk.

PR #1098, opened 2023-02-08
Windows write to log if no debugger attached

tomkaragounis:
Should attempt to write debug info to log on windows if no debugger is available to recieve output

Debug logging will now be off by default


lib/pal/PAL.cpp
```diff
@@ -223,7 +223,22 @@ namespace PAL_NS_BEGIN {
...
             buffer[std::min<size_t>(len + 0, sizeof(buffer) - 2)] = '\n';
             buffer[std::min<size_t>(len + 1, sizeof(buffer) - 1)] = '\0';
-            ::OutputDebugStringA(buffer);
+            if (IsDebuggerPresent())
+            {
+                ::OutputDebugStringA(buffer);
+            }
+            else
+            {
+                // Log to debug log file if enabled
```
maxgolov, line 232: Biggest complaint was usually if something goes wrong (e.g. customer blocks the end-point with firewall), these logs may grow substantially big on disk. Win 10 Storage Sense somewhat alleviates it by cleaning up the old session logs, but still it may get bad, especially on server flavor. This needs to be carefully tested and the logs should def be OFF by default. Like, no logs should ever be created by default unless the app config for debug/dev/test environments explicitly enables it.

tomkaragounis: I have shifted logging to be default off. Is there somewhere I can note the change of behavior for users to be aware of that they will need to add a setting to continue getting logs?
Could you be more specific on the type of testing you would like to see?

maxgolov: Please check with @lalitb during weekly meeting. As long as you can avoid SDK spewing too many warnings into log file (ensuring that normally the file isn't being written too), it's all good.

lalitb: Just to add, there is plan to add the support for log rotation/limiting in near future.

lalitb (approved):
Looks good to me. We can merge this after discussing in next week's community meeting.

PR #1099, opened 2023-02-12
Maxgolov/shared library

maxgolov:
C# projection layer - old work item #4 

It took us a few years to realize that we finally need 1DS C++ SDK bindings for Unity C# code.

Implementation should work well under the following environments:
- Unity WebAssembly IL2CPP (Unity transpiles C# to C++ to WebAssembly)
- Unity Editor with Mono 2.x
- Unity based on NETStandard 2.x+

List of changes:
- allow **build.sh** to accept parameters and cmake options
- clean-up CMakeLists.txt to allow building a shared library (tested on Mac only, need to test on Linux)
- update corresponding packaging scripts to include the shared library output
- update package versions from v3.2 to v3.3
- mod mat.h to pack structure
- draft example that shows how to P/Invoke from .NET Core 3.x into shared library (tested on Mac OS X and Windows). This is a DRAFT, not final. But it is a working draft to upload events to Collector.

Practical reasons for this implementation:

**Reason 1. New ways to express an event for Geneva IFx SDK without using Bond. Main intent is to move IFx customers away from Bond in their code to 1DS-style EventProperties.**

A concept of strongly-typed variant dictionary that can be used to express an event in a very user-friendly, intuitive manner. This is C# code that is pretty much identical to API surface we have in C++11 land and other 1DS SDKs:
```
                var props = new EventProperties() {
                    { "strKey", "value1" },
                    { "intKey", 12345 },
                    { "dblKey", 0.12345 } ,
                    { "guidKey", new Guid("73e21739-9d4e-497d-9c66-8e399a532ec9") }
                };
```

This managed object is implemented as "variant dictionary":
```
Dictionary<string, EventProperty>
```

where _EventProperty_ - allows to contain the standard supported by collector strong types, plus optional Pii Kind tag. That managed container object then gets converted to native structure passed down to 1DS C++ SDK via "ABI stable" C API. The marshaling is not ideal at the moment. I believe we can make it x2 times faster by avoiding memcpy in a few spots.

**Reason 2. Prove that it's a viable approach from performance standpoint that can be used in system services written in C#.**

Benchmarking how fast we can go -- measuring combined time of the following steps:
- emit an event in managed code
- translate it from managed into native representation, including string translations from UTF-16
- serialize it with native code Common Schema 4.x Bond serializer
- save it to disk storage in SQLite DB

Preliminary results of early implementation:
- 9 seconds to process 100,000 managed code events (string, integer, double and guid, I'm a bit cheating with GUID as I've been processing it as string... final properly packed GUID would be much faster)
- Note that every event also gets decorated with all possible CS4.x Part A properties, like system state, device id, timestamps, etc... many others. So it's a "bulky" event, not just 4 raw fields.
- 10,000 events per second can be achieved on an oldish 2015 Macbook Pro with one thread (not a server machine)
- 50,000 events per second on more modern i7-9750H at 2.6GHz with one thread (gaming laptop, lol)
- no memory leaks (ahem).
- no memory fragmentation (ahem-ahem...).

**Reason 3. We can do the rest of "incremental" work to provide a compat layer for Aria C# Server customers on top of our 1DS C++ SDK**. 

We can also deprecate our C++/CX and C++/CLI projection layers. Those two other technologies to generate Windows C# wrapper are "legacy" and "deprecated", we should switch to netcore P/Invoke instead.

**Reason 4. I wanted to learn more about data marshaling from managed to native via P/Invoke** 😄(out of curiosity ... to learn something new)

Perf numbers for a quick test to emit 100000 events in ONE thread in a loop (not multi-threaded, just 1 core, 1 background disk flush thread):
```
...
SDK version: 3.3.0
>>> evt_open...
handle=2933216226
>>> evt_pause...
>>> evt_log...
Elapsed    = 00:00:09.3721759
Event rate = 10669.880833115818 eps
Latency    = 0.093721759 ms
Mem used   = 6208 bytes
Fragmented = -37536 bytes
>>> evt_close...
result=0
```

Usage instructions for building shared library instead of static:

```
bash-3.2$ ./build.sh -h
Usage: build.sh [clean] [noroot] [release] [-h|-?] [-l (static|shared)] [-D CMAKE_OPTION]

options:

 -h | -?             - this help.
 -l [static|shared]  - build static (default) or shared library.
 -D [CMAKE_OPTION]   - additional option to pass to cmake.

cmake options can be passed using CMAKE_OPTS environment variable.
```

To build shared library:
```
./build.sh -l shared
```

or - new way of passing options from script to cmake:
```
CMAKE_OPTS=-DBUILD_SHARED_LIBS=ON ./build.sh
```

Unity wrapper will be maintained elsewhere outside of this repo.

### DIFF
--- a/wrappers/netcore/EventSender.csproj
+++ b/wrappers/netcore/EventSender.csproj
@@ -1,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+  </PropertyGroup>
+  <ItemGroup>
+    <None Include="sdk-config.json" CopyToOutputDirectory="Always" />
+    <None Include="EventSender.xml">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="System.Json" Version="4.7.0" />
+  </ItemGroup>
+</Project>
--- a/wrappers/netcore/Program.cs
+++ b/wrappers/netcore/Program.cs
@@ -1,0 +1,82 @@
+using System;
+using System.Runtime.InteropServices;
+using System.Text;
+using System.IO;
+using System.Json;
+using System.Diagnostics;
+
+using Microsoft.Telemetry.Core;
+
+namespace EventSender
+{
+
+    class Program
+    {
+        static string ReadConfiguration(string filename)
+        {
+            string result = "";
+            using (StreamReader sr = new StreamReader(filename))
+            {
+                result = sr.ReadToEnd();
+            }
+            return result;
+        }
+
+        static void Main(string[] args)
+        {
+            Console.WriteLine("Reading configuration...");
+            string cfg = ReadConfiguration("sdk-config.json");
+
+            // Parse to verify it is valid and print it out..
+            // Parser should throw if config is invalid.
+            JsonObject jsonDoc = (JsonObject)JsonObject.Parse(cfg);
+            Console.WriteLine("SDK configuration:\n{0}", jsonDoc.ToString());
+
+            // Obtain SDK version from native library
+            Console.WriteLine("SDK version: {0}", EventNativeAPI.evt_version());
+
+            // Initialize
+            Console.WriteLine(">>> evt_open...");
+            var handle = EventNativeAPI.evt_open(cfg);
+            Console.WriteLine("handle={0}", handle);
+
+            // Initialize
+            Console.WriteLine(">>> evt_pause...");
+            EventNativeAPI.evt_pause(handle);
+
+            Console.WriteLine(">>> evt_log...");
+
+            long total0 = GC.GetTotalMemory(true);
+            long frag0  = GC.GetGCMemoryInfo().FragmentedBytes;
+            Stopwatch sw = new Stopwatch();
+            sw.Start();
+
+            const int MAX_ITERATIONS = 100000;
+            for (int i = 0; i < MAX_ITERATIONS; i++)
+            {
+                var props = new EventProperties() {
+                    { "strKey", "value1" },
+                    { "intKey", 12345 },
+                    { "dblKey", 0.12345 } ,
+                    { "guidKey", new Guid("73e21739-9d4e-497d-9c66-8e399a532ec9") }
+                };
+                EventNativeAPI.evt_log(handle, ref props);
+            }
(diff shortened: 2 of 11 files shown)


wrappers/netcore/EventSender.csproj
```diff
@@ -0,0 +1,27 @@
...
+        adjust the native libs path accordingly.
+    -->
+    <None Include="$(ProjectDir)\..\..\Solutions\out\Debug\x64\win32-dll\*.*">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+    <!-- Don't show included native libraries in Visual Studio editor -->
+    <Content Include="$(ProjectDir)\..\..\Solutions\out\Debug\x64\win32-dll\*.*" Visible="False" />
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="System.Json" Version="4.7.0" />
```
maxgolov, line 25: While this works for .NET Standard 2.x , unfortunately Newtonsoft works better in Unity. I'm not changing it here since this will be taken care of elsewhere, in the implementation of Unity Package outside of 1DS repository.

wrappers/netcore/Program.cs
```diff
@@ -0,0 +1,133 @@
+using System;
+using System.Runtime.InteropServices;
+using System.Text;
+using System.IO;
+using System.Diagnostics;
+using Microsoft.Telemetry.Core;
+
+#if HAVE_SYSTEM_JSON
```
maxgolov, line 8: Need to define it.

wrappers/netcore/README.md
```diff
@@ -0,0 +1,19 @@
+# .NET Core wrapper example for 1DS C/C++ SDK
+
+Note that this wrapper is incomplete simple reference implementation that illustrates how to use 1DS C API from cross-platform .NET Core application.
```
maxgolov, line 3: I'll change the wording of `incomplete` to `provides the same level of functionality as C API`, i.e. some features are accessible only from C++ layer. This still allows the native code initializer to use `SetContext(...)` efficiently, to populate common known fields, like `session id`, etc.. C# layer becomes `guest` to `host` initialized by native library.

lalitb:
In case the code under `wrappers/netcore` is to demonstrate calling capi from C# code using P/Invoke, should we move this code under `example` directory ?

maxgolov:
| should we move this code under `example` directory?

I can split it into two parts:
- `EventNativeAPI.cs` - belongs in `wrappers` dir. This is nearly identical to what we use in the product.
- `Program.cs` - belongs in `examples` dir.

lalitb:
> I can split it into two parts:

Yes, that will make sense. Thanks.

maxgolov:
Sorry for delay. We integrated it on our end, so I put it on a back-burner.
I'll get back to this.

PR #1101, opened 2023-02-14
replacing rand() with mt19336 PSRG

absaroj:
rand() is causing static security analysis issues, replacing this pseudo random number generator with mt19937().

### DIFF
--- a/lib/pal/PAL.cpp
+++ b/lib/pal/PAL.cpp
@@ -348,11 +348,11 @@ namespace PAL_NS_BEGIN {
 
         static std::once_flag flag;
         std::call_once(flag, [](){
-			std::array<std::mt19937::result_type, std::mt19937::state_size> seed_data;
-			std::random_device rd;
-			std::generate_n(seed_data.data(), seed_data.size(), std::ref(rd));
-			std::seed_seq seq(std::begin(seed_data), std::end(seed_data));
-			random_generator.seed(seq);
+            std::array<std::mt19937::result_type, std::mt19937::state_size> seed_data;
+            std::random_device rd;
+            std::generate_n(seed_data.data(), seed_data.size(), std::ref(rd));
+            std::seed_seq seq(std::begin(seed_data), std::end(seed_data));
+            random_generator.seed(seq);
         });
 
         GUID_t uuid;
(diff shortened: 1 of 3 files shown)


eduardo-camacho (approved):

lib/pal/PAL.cpp
```diff
@@ -343,19 +343,24 @@ namespace PAL_NS_BEGIN {
 	std::transform(uuidStr.begin(), uuidStr.end(), uuidStr.begin(), ::tolower);
         return uuidStr;
 #else
+        static std::mt19937 random_generator;
```
maxgolov, line 346: Can you check with Visual Studio debug tracepoint on Windows how often this function gets called on a simple test?

Is it significantly heavier than the old one?

lalitb: std::mt19937 can be heavy. In OpenTelemetry, we use [XorShift](https://en.wikipedia.org/wiki/Xorshift) shift register generator - https://github.com/open-telemetry/opentelemetry-cpp/blob/v1.8.2/sdk/src/common/fast_random_number_generator.h. But better to benchmark.

maxgolov: My thoughts are that this spot is NOT used for any encryption-related activities. It would've been risky if we could predict the GUIDs randomly generated, but at massive scale - we haven't yet seen a duplicate GUID, even if the function itself is not "random enough". There are several suggestions on stackoverflow how this could be improved further.

I am not against the safer/nicer/cleaner solution, but the static code analysis warning in this spot is "false positive".

absaroj: Just ran a simple test program to compare the performance of both of the PRNGs (rand() and mt19337) and it seems there is perf impact in latter's case. It seems mt19937 is 4 times slower than rand().
Agree with your comment about rand() fulfilling the requirement here. I think supressing the warning generated here would be the better approach.

absaroj: Also, @maxgolov can you please share the suggestion link you mentioned on the stackoverflow. I can try looking at those as well to see if they fit here.

absaroj:
Abandoning the PR based on the comments provided about perf regressions caused by the new PRNG. I would explore some directives to suppress the error generated by the tool and keep using the current rand() PRNG.

Issue #1095, opened 2023-02-02
Modernize android build 


anod:
The current android build uses an outdated Gradle, cmake and java build dependencies and requires outdated IDE Android Studio 4 to run the project 

Suggested to upgrade
1. Gradle and android Gradle plugin
2. Compile SDK version
3. Cmake
4. Test project dependcies 
4. Java 17

Can we upgrade it? Is there any additional guidance?

lalitb:
@anod It would be good to upgrade the android build toolset we are using. I am not Java/Android expert, will this anyway change the minimum sdk version we correctly support, as I know few teams still needing support for Android 4.4 (sdk level 19) ?

anod:
Thanks @lalitb I think there is no need to update minimum version. I will test and comeback with more details.
it might require to upgrade cmake, is there limitation for cmake version capabilities.

lalitb:
> it might require to upgrade cmake, is there limitation for cmake version capabilities.

Which cmake version to upgrade to ? Ideally it would be good to support 3.1.0 as minimum version, and conditionally include any feature required from higher version:

```
cmake_minimum_required(VERSION 3.1.0)
if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.12")
  ...
endif()
```

maxgolov:
Can I ask so that when we implement this, we can specify what versions of Android deps to use via env vars. My team needs this layer to be compatible with certain versions of NDK and SDK.. I wonder if we can use something like this:

```
export SDKAPILevelOverride=android-28
export NDKAPILevelOverride=latest
```
or what not. Similar how it's handled in Unreal Engine. It's a [moving target for us](https://developer.oculus.com/documentation/native/android/mobile-studio-setup-android/#install-additional-packages-and-tools)

anod:
For SDK level there is an option to overlay default config with custom `tools.gradle.user` file
I think app which depends on the lib upgrades a library defined target sdk, lowering target sdk can cause unpredictable behavior

Regarding NDK, There is [a default NDK](https://developer.android.com/studio/projects/install-ndk) mapping between Gradle plugin and NDK 
NDK version can be overridden via `ANDROID_NDK_VERSION` env variable

Regarding `cmake`, current version is 3.10, but there is a warning
```
CMake Warning at /Users/algavris/Library/Android/sdk/ndk/23.1.7779620/build/cmake/android-legacy.toolchain.cmake:416 (message):
[...]

PR #1103, opened 2023-02-24
Upgrade android gradle plugin [Java 11 build]

anod:
Upgrade Android Gradle Plugin to support for Android Studio 2022.1.1

#1095 

**Breaking build change: use Java 11 to build the project**

Upgrade  android gradle plugin 4.0.1 -> 7.3.1
Upgrade gradle 6.1.1 -> 7.5
 - Upgrade default NDK for gradle 7.5: 21.1.6352462 -> 23.1.7779620
Upgrade room 2.2.6 -> 2.4.3
Update compileSdkVersion 29 -> 32


sid-dahiya (changes requested):
This PR is updating a lot of the core Android versioning and can lead to app teams to run into issue as there are teams that need to support older Android versions. Is there any explicit requirement to move to these SDK versions?

anod:
> This PR is updating a lot of the core Android versioning and can lead to app teams to run into issue as there are teams that need to support older Android versions. Is there any explicit requirement to move to these SDK versions?

Is there a specific library that can cause a problem? 
The change just provides a new defaults and doesn't update minimum requirements,
also configuration allows to override existing values for custom build via `tools.gradle.user` file

sid-dahiya (approved):

sid-dahiya:
> > This PR is updating a lot of the core Android versioning and can lead to app teams to run into issue as there are teams that need to support older Android versions. Is there any explicit requirement to move to these SDK versions?
> 
> Is there a specific library that can cause a problem? The change just provides a new defaults and doesn't update minimum requirements, also configuration allows to override existing values for custom build via `tools.gradle.user` file

My bad! I misread the changes are for defaults.

lalitb:
Seems CI build is failing for latest gradle version -
```No matching variant of com.android.tools.build:gradle:7.4.1```

Do we need to install this in CI pipeline explicitly?

**Comment by anod** [comment]
> Seems CI build is failing for latest gradle version - `No matching variant of com.android.tools.build:gradle:7.4.1`
> 
> Do we need to install this in CI pipeline explicitly?

Looking into that... usually not needed unless pipeline not overriding something

**Comment by anod** [comment]
@lalitb I think I figred out the build error. 

New gradle plugin requires Java 11, but android sdk tools instaled via choc are outdated. 

There is an action to install newer SDK tools (https://github.com/marketplace/actions/setup-android-sdk-tools) 

Is it ok to use the action or I new to download new version manually?

**Comment by anod** [comment]
> @lalitb I think I figred out the build error. 
> 
> New gradle plugin requires Java 11, but android sdk tools instaled via choc are outdated. 
> 
> There is an action to install newer SDK tools (https://github.com/marketplace/actions/setup-android-sdk-tools) 
> 
> Is it ok to use the action or I new to download new version manually?
> 

On second thought probably we don't want to upgrade java yet, this needs to be communicated, so I will lower grade build plugin to previous version

**Comment by lalitb** [comment]
> On second thought probably we don't want to upgrade java yet, this needs to be communicated, so I will lower grade build plugin to previous version

**Comment by anod** [comment]
Looks like Java 11 is required for running new build, I will prepare the PR to use java 11 and once we agreed on timelines will publish

File: .github/workflows/build-android.yml [hunk 1]
```diff
@@ -23,39 +23,54 @@ jobs:
...
       with:
-        submodules: 'true'
-      continue-on-error: true
-    - name: Chocolatey
+        depth: 1
+        submodules: 'false'
+    - name: Update submodules
+      run: |
+        git submodule sync
+        git config --global submodule.lib/modules.update none
```
anod, line 33: @lalitb  FYI - For some reason I had to exclude manually update of lib/module, otherwise other submodules weren't checked out

anod:
@lalitb @sid-dahiya Can you please take a look on the latest iteration, I would like to complete it next week

mkoscumb (approved):
LGTM

.github/workflows/build-android.yml
```diff
@@ -23,39 +23,54 @@ jobs:
...
+        git -c protocol.version=2 submodule update --init --force --depth=1
+    - name: Setup Java
+      uses: actions/setup-java@v3
+      with:
+        distribution: 'adopt'
+        java-version: '11'
+    - name: Remove default github maven configuration
+      # Workaround for: 'Unable to decrypt local Maven settings credentials'
```
ThomsonTan, line 41: Is this workaround caused by some bug?

anod: Looks like settings.xml included in the image has some format that not readable  by gradle

https://github.com/mark-vieira/gradle-maven-settings-plugin/issues/15#issuecomment-404548907

lalitb (approved):

PR #1104, opened 2023-02-27
Introduce option to checkpoint database on Flush()

dawidk-msft:
The way SQLite is initialized (synchronous=NORMAL config) introduces potential issues with durability of changes/updates in some corner cases (power loss, ungraceful process termination/exit, etc).

iOS Outlook logs telemetry in scenarios where if some critical validation/check failed and cannot be handled process will self-terminate. In between logging the event and terminating we call Flush() to avoid lost events. If we checkpoint the DB before returning from the Flush() call app can safely assume no events will be lost. This change introduces an option that enables such behavior.

Alternative solution would be to use the default configuration (with synchronous=FULL) which doesn't have that problem but has overhead of filesystem syncing for each transaction which makes it a bad idea from the perf point of view.

### DIFF
--- a/lib/include/public/ILogConfiguration.hpp
+++ b/lib/include/public/ILogConfiguration.hpp
@@ -145,6 +145,11 @@ namespace MAT_NS_BEGIN
     static constexpr const char* const CFG_INT_RAM_QUEUE_BUFFERS = "maxDBFlushQueues";
 
     /// <summary>
+    /// SQLite DB will be checkpointed when flushing.
+    /// </summary>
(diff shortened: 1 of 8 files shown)


lalitb:
While the changes are behind config flag, so should be good in general. Just wanted to check, if you have seen the DB corruption / data loss issue in power failure scenario. The synchronous=NORMAL mode should be in general resilient, with very small window of WAL logs not getting synced to physical disk since last checkpoint. Haven't seen this window getting hit, so just curious.

lalitb (approved):
LGTM. The PR was discussed in today's community meeting, and above query was clarified.

PR #1105, opened 2023-02-27
Trigger callback when DDV communication fails

desmart18:


lalitb:
@desmart18  - Is this issue still valid, as private modules repo is already updated

PR #1107, opened 2023-03-03
Add StampEventIKeyForConcerns option in Java layer

sid-dahiya:
Some teams want to be able to always stamp the associated tenant with events as they have events for different tenants flowing through.


lib/include/public/Version.hpp
```diff
@@ -18,7 +18,7 @@ namespace MAT_NS_BEGIN {
 uint64_t const Version =
     ((uint64_t)3 << 48) |
     ((uint64_t)7 << 32) |
-    ((uint64_t)46 << 16) |
+    ((uint64_t)62 << 16) |
```
lalitb, line 21: nit - do you plan to do new release with this version, or else you want me to do a release?

lalitb (approved):

lib/include/public/Version.hpp
sid-dahiya: Yes, I can do a release.

PR #1110, opened 2023-03-08
Integrate Event Tracking Mechanism in 1DS

shivpratap1992:
**Description**: 
Currently there is no way to track what event sent from MIP SDK to 1DS are actually being emitted out.

**Fix**: 
MIP SDK will populate tracking ID in CV while calling evt_log() method under event properties argument in MIP SDK side and write it in MIP log. Eventually, we will add this CV (tracking ID) in http request header while calling MIP http delegate in 1DS.

**Design Doc**: https://microsoft.sharepoint.com/:w:/t/ceip/EVeouFG5ZWBItrZbmppxOxgBSP9dhzDa3vFhVjXJmefw1w


lib/http/HttpRequestEncoder.cpp
```diff
@@ -45,6 +45,7 @@ namespace MAT_NS_BEGIN {
         ctx->httpRequest->SetUrl(m_config.GetCollectorUrl());
 
         ctx->httpRequest->GetHeaders().set("Expect", "100-continue");
+        ctx->httpRequest->GetHeaders().set("CV", ctx->cvStr);
```
lalitb, line 48: This looks to be non-standard header content, not part of contract with Collector.

shivpratap1992: we will remove this MIP Http delegate before sending to collector

lib/system/Contexts.hpp
```diff
@@ -34,7 +34,7 @@ namespace MAT_NS_BEGIN {
 
         IncomingEventContext(std::string const& id, std::string const& tenantToken, EventLatency latency, EventPersistence persistence, ::CsProtocol::Record* source)
             : source(source),
-            record{ id, tenantToken, latency, persistence },
+            record{ id, tenantToken, latency, persistence, source->cV },
```
rybest, line 37: >source [](http://example.com/codeflow?start=59&length=6)

Potential crash - don't use source without checking it for null.

lib/http/HttpRequestEncoder.cpp
rybest, line 48: >set [](http://example.com/codeflow?start=39&length=3)

Can we make this assignment based upon some compiler define so we can configure it only for MIP?

rybest (changes requested):
:clock1:

lib/http/HttpRequestEncoder.cpp
lalitb: And how will this affect the non-windows clients ?

lib/system/Contexts.hpp
shivpratap1992: Thanks. Missed it somehow. let me fix it

lib/http/HttpRequestEncoder.cpp
shivpratap1992: Do we need it. asking since we will be using it for MIP only. Or you mean for scenario, going forward we may put this change in global 1DS repo

shivpratap1992: Currently this will not be checked in for non-mip clients. It will be used for our specific scenarios.

Meanwhile Lalit it will be great if 1DS can expose LogManager for future perspective in dll

shivpratap1992: Made required changes

shivpratap1992: Made required changes

lib/packager/Packager.cpp
```diff
@@ -48,6 +48,7 @@ namespace MAT_NS_BEGIN {
             LOG_TRACE("Adding event %s:%s, size %u bytes",
                 tenantTokenToId(record.tenantToken).c_str(), record.id.c_str(), static_cast<unsigned>(record.blob.size()));
 
+            ctx->cvStr = record.cvStr;
```
maxgolov, line 51: Why is it `cvStr` ? Correlation Vector technology is no longer recommended for use. This is supposed to be W3C TraceContext going forward. I think even naming this thing using a "legacy tech" name isn't great..

maxgolov: https://www.w3.org/TR/trace-context/

lib/http/HttpRequestEncoder.cpp
maxgolov: > going forward we may put this change in global 1DS repo

I think adding this header would be wrong. This is not the tech we should be using going forward.

maxgolov:
Can you explain what you are trying to achieve. It'd be best for the PR to have some non-empty description.

lib/http/HttpRequestEncoder.cpp
lalitb: It should be fine as long as these changes are confined to a separate branch, for debugging purpose, and not added to the `main` branch.

rybest (approved):
:shipit:

lib/packager/Packager.cpp
rybest: +1. If this code ever does get integrated back into the main branch, it would be good to use agreed upon standards. Since this is a private branch, I'm fine with this for now, but it would be best to update it to conform to the standard.

rybest:
+1

---
In reply to: [1462917246](https://github.com/microsoft/cpp_client_telemetry/pull/1110#issuecomment-1462917246) [](http://example.com/codeflow?ancestors=1462917246,1462917246)

rybest (changes requested):
:clock1:

lib/packager/Packager.cpp
shivpratap1992: Thanks @maxgolov let me have a look. Sure Ryan we will update it as required.

Updated description with design doc, @maxgolov you can refer.  Basically, this required for MIP as per customer issue diagnostics.

Just mentioning the option we already tried and stuck  later:

- we tried integrating the client telemetry(1DS) Static library into the MIP code for integrating debug event listener from the MIP code

but it's generating a Runtime library issue for MIP build since MIP code uses CXXFlags MDd for using dll versions. 
Changing these flags in the MIP code may require our partners to also update these CXXFlags as well, hence this change will be very sensitive for us and risky. We are looking into it parallelly.

- 1DS Static library size is also concerning ~150 MB whereas dynamic library size if ~4 MB.

- On the MIP side DoSendHttp method, we tried to decode the request body sent by one DS in the attached callback. But that also showed a linker issue since the symbol was not exposed.


Please let us know if there is some easy way to achieve like mentioned below or some other:

- Exposing required function from client telemetry dll version.
- Passing custom logger delegate and unique id to 1DS.

maxgolov (approved):

lib/http/HttpRequestEncoder.cpp
```diff
@@ -45,6 +45,11 @@ namespace MAT_NS_BEGIN {
         ctx->httpRequest->SetUrl(m_config.GetCollectorUrl());
 
         ctx->httpRequest->GetHeaders().set("Expect", "100-continue");
+
+#ifdef HAVE_MAT_EVT_TRACKINGID
+        ctx->httpRequest->GetHeaders().set("Trace-Id", ctx->traceId);
```
lalitb, line 50: As per the w3c trace context specs, the trace_id should be part of HTTP header "traceparent" - 
https://www.w3.org/TR/trace-context/#traceparent-header

Header["traceparent"] =  `version`-`trace-id`-`parent-id`-`trace-flags`

I don't have strong opinion for now, as we are not sending it across the wire.

docs/building-custom-SKU.md
```diff
@@ -26,6 +26,7 @@ Build recipe must contain the following preprocessor definitions:
 | HAVE_MAT_ZLIB | on | Use zlib for HTTP requests compression. This option must always be turned on for any high-volume telemetry project |
 | HAVE_MAT_LOGGING | on | Enable internal SDK tracing / debug logging |
 | HAVE_MAT_WIN_LOG | off | Will log statements to disk on windows if trace enabled and HAVE_MAT_LOGGING defined |
+| HAVE_MAT_EVT_TRACKINGID | off | Disable audit event tracking by adding event id (CV) to http request header on windows |
```
lalitb, line 29: How about using HAVE_MAT_EVT_TRACEID macro ?

nit : `HAVE_MAT_EVT_TRACEID` - Enable event tracking by adding trace-id to http request header on Windows

lib/include/public/IOfflineStorage.hpp
```diff
@@ -33,12 +33,13 @@ namespace MAT_NS_BEGIN {
         StorageBlob     blob;
         int             retryCount = 0;
         int64_t         reservedUntil = 0;
+        std::string     traceId;
```
lalitb, line 36: Can we have this definition under macro, this will ensure that `StorageRecord` size will not change for non windows platform, if we plan to merge these changes to main.

lib/include/public/IOfflineStorage.hpp
```diff
@@ -33,12 +33,13 @@ namespace MAT_NS_BEGIN {
...
 
         StorageRecord()
         {}
 
-        StorageRecord(std::string const& id, std::string const& tenantToken, EventLatency latency, EventPersistence persistence)
-            : id(id), tenantToken(tenantToken), latency(latency), persistence(persistence)
+        StorageRecord(std::string const& id, std::string const& tenantToken, EventLatency latency, EventPersistence persistence, std::string traceId)
+            : id(id), tenantToken(tenantToken), latency(latency), persistence(persistence), traceId(traceId)
```
lalitb, line 42: Have separate constructor for `StorageRecord` with traceId ( and behind macro).

lib/include/public/IOfflineStorage.hpp
shivpratap1992: done

lib/include/public/IOfflineStorage.hpp
shivpratap1992: done

lalitb (approved):

docs/building-custom-SKU.md
shivpratap1992: done

lib/http/HttpRequestEncoder.cpp
shivpratap1992: Currently we need traceId only, hence logging that only. However, let me know if this change is must.

lib/system/Contexts.hpp
```diff
@@ -88,6 +97,7 @@ namespace MAT_NS_BEGIN {
         unsigned                             maxUploadSize = 0;
         EventLatency                         latency = EventLatency_Unspecified;
         std::map<std::string, size_t>        packageIds;
+        std::string                          traceId;
```
lalitb, line 100: nit - this can be behind macro too.

shivpratap1992: sure. will take care while merging in 1DS main repo

lib/http/HttpRequestEncoder.cpp
```diff
@@ -45,6 +45,11 @@ namespace MAT_NS_BEGIN {
         ctx->httpRequest->SetUrl(m_config.GetCollectorUrl());
 
         ctx->httpRequest->GetHeaders().set("Expect", "100-continue");
+
+#ifdef HAVE_MAT_EVT_TRACEID 
+        ctx->httpRequest->GetHeaders().set("Trace-Id", ctx->traceId);
```
aks3090, line 50: Can we add a simple test for this with macro define to ensure we get this field in header(future proofing change)

shivpratap1992: Ack. will address this comment in PR for main 1DS repo

lib/system/Contexts.hpp
```diff
@@ -32,12 +32,21 @@ namespace MAT_NS_BEGIN {
         {
         }
 
+#ifdef HAVE_MAT_EVT_TRACEID   
+        IncomingEventContext(std::string const& id, std::string const& tenantToken, EventLatency latency, EventPersistence persistence, ::CsProtocol::Record* source)
+            : source(source),
+            record{ id, tenantToken, latency, persistence, (source != nullptr) ? source->cV : "" },
+	    policyBitFlags(0)
```
rybest, line 39: >	 [](http://example.com/codeflow?start=0&length=1)

nit: consistent spacing (tabs for spaces)

rybest (approved):
:shipit:

PR #1115, opened 2023-03-15
[OneDS] Integrate Event Tracking Mechanism in OneDS

shivpratap1992:
**Description:**
Currently there is no way to track what event sent from application like MIP SDK to 1DS are actually being emitted out.

**Fix:**
Application like MIP SDK will populate tracking ID in CV while calling evt_log() method under event properties argument in MIP SDK side and write event id in MIP log before sending. Eventually, we will add this CV (tracking ID) in http request header while calling MIP http delegate in 1DS.

Design Doc: https://microsoft.sharepoint.com/:w:/t/ceip/EVeouFG5ZWBItrZbmppxOxgBSP9dhzDa3vFhVjXJmefw1w

**Flight/Setting:**

Entire code is under a defined macro, until this HAVE_MAT_EVT_TRACEID  macro is not defined, new code won't take effect.


tests/functests/BondDecoderTests.cpp
```diff
@@ -17,8 +17,7 @@
 #include "LogManager.hpp"
 
 #include <atomic>
-#include <cassert>
-
+#include <cass
```
lalitb, line 20: Probably a typo?

shivpratap1992: removed

docs/building-custom-SKU.md
```diff
@@ -26,6 +26,7 @@ Build recipe must contain the following preprocessor definitions:
 | HAVE_MAT_ZLIB | on | Use zlib for HTTP requests compression. This option must always be turned on for any high-volume telemetry project |
 | HAVE_MAT_LOGGING | on | Enable internal SDK tracing / debug logging |
 | HAVE_MAT_WIN_LOG | off | Will log statements to disk on windows if trace enabled and HAVE_MAT_LOGGING defined |
+| HAVE_MAT_EVT_TRACEID  | off | Enable event tracking by adding trace-id to http request header on Windows |
```
lalitb, line 29: Can you bit add more context here  -  this is for debugging purpose, and not recommended to be enabled in production code. The collector doesn't parse/read this header. As of now, this is meant to be used through the capi, where the http-send handler is supposed to remove this header from the event data  before sending it to collector.

shivpratap1992: Sure

lib/system/Contexts.hpp
```diff
@@ -88,6 +97,7 @@ namespace MAT_NS_BEGIN {
         unsigned                             maxUploadSize = 0;
         EventLatency                         latency = EventLatency_Unspecified;
         std::map<std::string, size_t>        packageIds;
+        std::string                          traceId;
```
lalitb, line 100: Can we move this under HAVE_MAT_EVT_TRACEID macro ?

lalitb (approved):
LGTM.

PR #1116, opened 2023-03-17
adding check after utf8string conversion before using it for locale code

absaroj:
In some rare cases, when UTF8String fails the conversions from NSString (due to low memory), it returns NULL, which if passed to std::string(), results in undefined behavior.

Adding null check before the std::string and if null then using "und" as unknown language code (as per iso 639).


wrappers/obj-c/ODWLogManager.mm
```diff
@@ -161,7 +161,17 @@ +(nullable ILogger *)initializeLogManager:(nonnull NSString *)tenantToken withCo
...
+
+            if (localeCode != nullptr)
+            {
+                strUserLocale = std::string(localeCode);
+            }
+            else
+            {
+                strUserLocale = "und";
+            }
+
```
lalitb, line 174: instead of using raw pointer, can we add size check if existing code as below:
```cpp
strUserLocale = std::string([[locale languageCode] UTF8String]);
if (strUserLocale.size() == 0) {
    strUserLocale = "und";
}
```

lalitb (approved):
LGTM. Thanks

PR #1118, opened 2023-03-23
Android, upgrade room and test deps, cleanup not required

anod:
Subset of #1103

~Upgrade room 2.2.6 -> 2.4.3 (latest 2.5.1)~ (requires java upgrade)
Upgrade espresso-core
Clean not needed libraries
Move HttpClientRequestTest to the library project, so it will run during build

### DIFF
--- a/lib/android_build/app/src/main/java/com/microsoft/applications/events/maesdktest/MaeUnitLogger.java
+++ b/lib/android_build/app/src/main/java/com/microsoft/applications/events/maesdktest/MaeUnitLogger.java
@@ -4,9 +4,6 @@
 //
 package com.microsoft.applications.events.maesdktest;
 
-import androidx.annotation.Keep;
-
-@Keep
 public abstract class MaeUnitLogger {
     abstract void log_failure(String filename, int line, String summary);
 }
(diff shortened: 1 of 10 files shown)


lib/android_build/app/src/main/java/com/microsoft/applications/events/maesdktest/MaeUnitLogger.java
```diff
@@ -4,9 +4,6 @@
 //
 package com.microsoft.applications.events.maesdktest;
 
-import androidx.annotation.Keep;
```
maxgolov, line 7: Wasn't this necessary to avoid the obfuscator from stripping/changing the method signature?

maxgolov (approved):
LGTM - except the `Keep` part, please elaborate on this one.

lib/android_build/app/src/main/java/com/microsoft/applications/events/maesdktest/MaeUnitLogger.java
anod: This is app that executes tests, is not part of maesdk

lalitb (approved):

PR #1122, opened 2023-03-30
C API improvements. Host-Guest improvements. Ability to update context via C API.

maxgolov:
_cc: @kindbe since we used to work originally on this feature, especially the `scope` parameter._

**Description**

There is a very detailed long-overdue markdown document that explains the Host-Guest feature, use-cases, purpose of `scope` (for sandboxing / isolation). And a separate file with fairly comprehensive Functional Tests (TADA!).

What it gives - TL;DR

Our team uses 1DS C API to pass telemetry contexts around in our hybrid app that is written in Unity C#, C/C++ and JavaScript (WebAssembly). Eventing works great. What we are missing is ability to `consolidate telemetry contexts` across languages. 1DS itself has the `SetContext` API and `Semantic Context` concept. Although these APIs were not exposed to C API. And did not have ability to intake parameters expressed in canonic Common Schema notation. Ex. you could not pass `ext.device.localId` or `ext.os.ver` to `SetContext`. Our team also uses exclusively the modern Common Schema. While old-style `COMMONFIELD` aliases could be appropriate for teams that export to A**a-Kusto, in our case we prefer our developers to operate on classic Common Schema events. The way you see them showing in Diagnostic Data Viewer application. Referencing old-style aliases causes confusion, questions, and unnecessary effort to remember things our devs would hardly ever need.

What's included:

- design document.
- comprehensive set of tests for Host-Guest functionality.
- two new C APIs: `evt_set_logmanager_context(s)` and `evt_set_logger_context(s)`. These APis are somewhat smarter than `SetContext` since they'd automagically detect a Common Schema Part A property if it starts with `ext.`. And remap it to corresponding place in Bond `CsRecord`. This would also be a gateway to eventual adoption of JSON notation: more and more customers move away from A**a-Kusto to their own private Kusto clusters - due to privacy, compliance, GDPR and EUDB requirements, that could be more easily satisfied with a private cluster. All these customers would prefer Common Schema notation `ext.app.localid` instead of `AppInfo.Id`. Also this matches precisely what you see on wire.

This PR also contains a stable C# ABI from my other older PR. I'll refactor the other one to use it as a C# usage example instead. API has been tested and proven to work on Windows, Mac, and Android ARM64 devices.


examples/cpp/SampleCppMini/main.cpp
```diff
@@ -31,7 +31,7 @@ void test_cpp_api(const char * token, int ticketType, const char *ticket)
 
     if (ticket != nullptr)
     {
-        const char *ticketNames[7] =
+        const char *ticketNames[8] =
```
maxgolov, line 34: This is actually someone else's bug. Visual Studio 2022 claims that it's a gross violation to reserve smaller array than the number of items you're listing.. cleaning it up here.

lib/api/LogManagerImpl.cpp
```diff
@@ -623,7 +623,6 @@ namespace MAT_NS_BEGIN
         LOG_INFO("SetContext");
         EventProperty prop(value, piiKind);
         m_context.SetCustomField(name, prop);
-        m_context.SetCustomField(name, prop);
```
maxgolov, line 626: This was also a glitch.. Probably someone merged the same line twice. Fixing it here.

maxgolov:
Hold on for a moment. I'll refresh it with the shutdown fix.

lib/include/public/ctmacros.hpp
```diff
@@ -127,4 +127,24 @@
 #define EVTSDK_LIBABI_CDECL MATSDK_LIBABI_CDECL
 #define EVTSDK_SPEC         MATSDK_SPEC
 
+/* Implement struct packing for stable FFI C API */
+#ifdef __clang__
+# define MATSDK_PACKED_STRUCT __attribute__((packed))
```
stephenatwork, line 132: This has the side effect of setting struct alignment to 1 which is probably not intended here? https://gcc.godbolt.org/z/jMK4xzbK8

The structs seem like they have been laid out so that natural alignment works, so I would suggest removing these macros unless there's a known issue. static_assert on size/offsets may give some safety?

maxgolov: @stephenatwork - there is (was) a known issue, at least on Mac. Padding depends on compiler optimization level. Current code has been tested on Quest (ARM64) and Win-x64, but I agree that adding static asserts for checking the offset and size totally makes sense.

Test app performs ingestion via C API, reconstructing into C++ EventProperties, then performs for-each - compare that there's a match. I can add a few props that'd be of size larger than 256 bytes / 65536 bytes, etc. to the test as well.

Let me add static asserts for size and offset checks.

stephenatwork: > Padding depends on compiler optimization level
🤯
OK if there are weird compilers to deal with then adding an explicit alignment attribute can avoid all kinds of awful shuffling (especially on platforms like ARM where unaligned loads are bad). So maybe `__attribute__((packed,align(4)))` would do it?

maxgolov: > (especially on platforms like ARM where unaligned loads are bad)

You'd actually fit more props in 64 byte data cache line with packing. Most loads and stores for strings - are pointers to aligned memory prepared by C# anyways:

- Perf is better because you can fit more props (e.g. 8 numbers) in [one data cache line](https://developer.arm.com/documentation/ddi0500/e/level-1-memory-system/about-the-l1-memory-system#:~:text=Data%20side%20cache%20line%20length%20of%2064%20bytes.)

- Perf is unaffected for string props: since once you got a pointer to string - you do loads and stores on aligned memory. C# calling into C on Win-x64 and Oculus would store the actual strings aligned.

I can address a potential issue for 32-bit. C# code already makes an assumption about the packed struct layout. I can add padding for 32-bit, so that the layout and sizing is identical. I'll push a change that adds a static assert check for padding/sizes.

maxgolov: Regarding Pii tags on individual props:
- This used to be a cool feature supported by the old Kusto path, with values obfuscated before being exported to DB.
- We moved on to a different routing model where this old feature is deprecated. There is an alternate technique to tag the whole events table as sensitive.

maxgolov: I kept the order for the following reasons:

- This already works for us on ARM64 and Win-x64, and we don't need more.

- Same order of fields as for older C API / ABI (though different padding/alignment). This allows to easily `#ifdef` and fallback to old C API. Trying to appease the customers who for some reason need to retain the older `v3.1` C ABI struct layout.

- Added an extra TC that verifies padding, alignment, struct size are consistent on 32-bit and 64-bit. Added support for Win32 32-bit / x86 (just in case, although my team doesn't require it). One IL assembly can now load either 32-bit or 64-bit version of the Client Telemetry shared lib, with both flavors exposing the same C API, basically with the same struct ABI across different platforms and compilers. Pointers to `const char *` and `void *`, of course, remain platform-dependent. But 32-bit version gets 32-bit padding in between, whereas 64-bit version occupies the full 64-bit for the pointers with no padding.

- Unaligned read problem is superficial: it won't actually occur on Intel and ARM. It does not affect perf either - because you fit more in fewer data cache lines. When either C/C++ or C# allocs an array of `evt_prop` structs on their heap, structs are `64(name)+32(type)+64(value)+32(piiKind)`-sized, packed back-to-back. More of them fits that way in a smaller-sized contiguous region of memory. Pointer to that region itself is aligned. Pointers to the actual props, e.g. strings that get allocated separately - are also all aligned. IF we ever need to support `RISC` with this... Well, there is no C# support for that. If someone would like to continue using SDK on `RISC`, then for those customers I added an option to revert back to original struct definition with a compiler define `HAVE_MAT_ABI_V3_1_0` . That way the padding decisions are left to compiler imagination. No C# support in that case though.

Given that it works on Android ARM64 and Win-x64 (for our org in the Mesh apps). And none of the CI checks of this PR stumbled upon the struct layout, including Apple mobile device tests and Android - all passing. Plus size-checks - all passed, I think we good.

stephenatwork: SGTM. Back compat plan is nice to have too.

On unaligned loads, maybe all platforms we care about handle them OK now. But have a look at
https://gcc.godbolt.org/z/6jWY6EP76 The runtime alignment doesn't matter so much - the compiler must emit code for the least aligned version it knows about. For some platforms it's only a little (or no) extra code. For some it can be a lot! (Gamedev made me super sensitive to code size)

maxgolov: Sounds good. Resolving this conversation. We can revisit later, if ever needed be.

Issue #1128, opened 2023-04-13
Expose setTicketToken() for Java for Android apps to set AAD(other tokens)


sonalpoddarmsft:
For Android apps who wants to set AAD token( or other tokens) on the java layer, there is no API currently to achieve that.

The C++ API -  `LogManager->GetAuthTokensController()->SetTicketToken(TicketType_AAD, value)` is the right approach to set the AAD Id. This will set the AAD value in HTTP header, and consumed by 1ds collector. 
But this is not exposed to the java layer.

The Java API - semanticContext.setTicket(TicketType.AAD, value) exists but never sends the token via HTTP request

Need a solution for Android apps to set token directly from java layer

bmehta001:
Implemented by #1131 (`setTicketToken exposed to jni and java layer for Android apps to set tokens like AAD`). `setTicketToken` is now available end-to-end in the Android Java layer:
- `ILogManager.java:26` — `public Status setTicketToken(TicketType type, final String tokenValue);`
- `LogManager.java:507` — public API bridging to `nativeSetIntTicketToken`
- `LogManagerProvider.java:92` — instance API, plus the JNI `nativeSetIntTicketToken` bindings

Android apps can set AAD (and other) ticket tokens through this API. Closing as completed.

PR #1131, opened 2023-04-14
setTicketToken exposed to jni and java layer for Android apps to set tokens like AAD

sonalpoddarmsft:
Issue: https://github.com/microsoft/cpp_client_telemetry/issues/1128


lib/include/public/LogManagerBase.hpp
```diff
@@ -585,6 +585,14 @@ namespace MAT_NS_BEGIN
         static IAuthTokensController* GetAuthTokensController()
             LM_SAFE_CALL_PTR(GetAuthTokensController);
 
+        ///Sets the ticket token with a value
```
ThomsonTan, line 588: Put the comment into `<summary></summary>` node.

sonalpoddarmsft: Done

lalitb (approved):
LGTM. The CI failure seems to be transient error due to network timeout, running again should fix it.

PR #1132, opened 2023-04-18
[MIP SDK][RHEL9] Dockerfile for  Rocky Linux 9 support

shivpratap1992:


docker/rockylinux9/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM mipsdkcontainer.azurecr.io/rockylinux9
+
+## Common packages for linux build environment
+RUN yum install -y yum-utils && \
+yum-config-manager --enable crb && \
+yum update -y && \
+yum -y group install "Development Tools" && \
```
dgbeliveau, line 7: It looks like a lot of this stuff is already installed on the base image: https://github.com/AzureAD/mip-sdk-for-cpp/pull/3443/files#diff-8d220225500880f9c1256710af99ef181556ced5628fddcb832e92c2b201f47e you don't need to install it again

docker/rockylinux9/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM mipsdkcontainer.azurecr.io/rockylinux9
```
lalitb, line 1: Is this registry publicly accessible? If not, we shouldn't be adding the setup here.

shivpratap1992: It is going to be checked in MIP SDK repo.

For global OneDs I will raise separate PR.

lalitb: My concern is - if this registry is Microsoft internal, we are exposing a name of private/internal registry name "mipsdkcontainer.azurecr.io" in public repo (even though it is separate branch).  Why do this need to be added here?

shivpratap1992: I see, But it seems other docker files  MIP SDK-specific OneDS branch also have a similar entries like for centos, ubuntu

FROM mipsdkcontainer.azurecr.io/centos8
FROM mipsdkcontainer.azurecr.io/ubuntu2004

FROM mipsdkcontainer.azurecr.io/library/centos:7

shivpratap1992: All these flavors have similar entry

image

lalitb: ok was just a suggestion. I don't have any issue as long as it is safe to expose these registries, and you are the better judge of that :)

shivpratap1992: Just checked an Authentication is required if some accessing outside. Also, this aims for a user branch...not the main OneDS branch....so I think its less risky, that may be the reason Vitally made these changes

dgbeliveau (approved):

PR #1138, opened 2023-04-24
Add Properties class for Swift wrappers

absaroj:
Swift Wrappers:
Adding Swift wrappers to 1DSSDK. Since C++ code can only be accessed from Swift via Obj-C or C, I am taking Obj-C with bridging header.

Wrapping ODWEventProperties.mm class for Swift. Contains other supporting files:
- local build script to build swift wrapper files only
- cmake list file to generate the xcode project files
- bridging header to access obj-c code from swift
- main swift file to check the call to added wrapper


wrappers/swift/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
+CMAKE_MINIMUM_REQUIRED(VERSION 3.16)
```
eduardo-camacho, line 1: This version is higher than root level and Obj-C wrapper level versions, why?

wrappers/swift/EventProperties.swift
```diff
@@ -0,0 +1,285 @@
+//
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+//
+
+/**
+Represents Event's properties.
+*/
+public class EventProperties {
```
eduardo-camacho, line 9: Please check what's then naming convention we use for Swift in APEX/Apple and adhere

eduardo-camacho: Found it, it's the 3 letters of the project, in this case it might be SDW

wrappers/swift/main.swift
```diff
@@ -0,0 +1,10 @@
+//
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+//
+
+let props = EventProperties(name:"TestEvent")
+props.setProperty("PropName", withValue: ["Type":"SwiftWrappers"])
+props.setProperty("PropWithPII", withInt64Value: Int64(30), withPiiKind: ODWPiiKind.distinguishedName)
```
eduardo-camacho, line 8: Are consumers going to use Obj-C constants/enums, or are you going to setup new ones for Swift?

absaroj: I am intending to let consumers use the obj-c enums instead of creating new ones in swift

wrappers/swift/CMakeLists.txt
absaroj: This was done when swift generators were failing to be generated and there was one suggestion to change it to this version.
But i think with xcode generators, this is not required. i can change it back to similar to obj-c.

There is below warning of deprecation though:

_CMake Deprecation Warning at CMakeLists.txt:1 (CMAKE_MINIMUM_REQUIRED):
  Compatibility with CMake < 2.8.12 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions._

wrappers/swift/EventProperties.swift
absaroj: thanks for sharing it. is there a doc/wiki for the conventions?

eduardo-camacho: Looking at other examples in Office codebase, it seems Swift naming guidance is different from Obj-C, so you might leave class names as you have it now

eduardo-camacho (approved):

wrappers/swift/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
+CMAKE_MINIMUM_REQUIRED(VERSION 2.8.2)
+
+PROJECT(Swift_Wrapper)
+
+message("${CMAKE_CURRENT_SOURCE_DIR}")
+
+enable_language(Swift)
```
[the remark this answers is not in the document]
absaroj: Yes. Thanks for pointing out. I think earlier I did use that version but changed it back to lower later.
Updated now.

wrappers/swift/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+
+# TODO: add logic to build dependencies
```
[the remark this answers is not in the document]
absaroj: Since, this is camke file only for the swift wrappers code and it has dependecy on objc, it would require objc built lib to be linked in this make files to generate the proper proj files.
I have added the dependencies and they seem to be working. I would be raising those in PR once this one is in.

wrappers/swift/SDWEventProperties.swift
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+    /**
+    Sets a string property for an event.
+
+    - Parameters:
+        - name: Name of the property
+        - value: Value of the property.
+    */
+    public func setProperty(_ name: String, withValue value: Any) {
```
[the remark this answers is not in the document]
absaroj: I have tried to expose the public methods based on the corresponding header file in objc wrapper, in this case:
ODWEventPropperties.h

wrappers/swift/SDWEventProperties.swift
```diff
@@ -0,0 +1,285 @@
...
+
+    /**
+    Sets a date property for an event with PiiTags.
+
+    - Parameters:
+        - name: Name of the property.
+        - value: A Date that contains the property value.
+        - piiKind: The kind of Personal Identifieable Information (PII), one from the ::ODWPiiKind enum values.
+    */
+    public func setProperty(_ name: String, withDateValue value: Date, withPiiKind piiKind: ODWPiiKind) {
```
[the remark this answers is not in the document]
absaroj: Thanks for the point. 
Yes, we have to access c++ via objc from swift.

Though I haven't noticed any interoperability issues while building and running the code. However there are minor changes when it comes to using the enum declared in objc from swift, such as Enum prefix name is removed while using. But since they are minor, i thought of keep using the objc enums rather than creating their duplicate in swift.
also, i have create a task to investigate more about the different compiler and version compatibility between swift-objc and update the source files accordingly.

lalitb (approved):
Looks Good, with some nit comments. Thanks for the PR.

wrappers/swift/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+get_property(dirs DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY INCLUDE_DIRECTORIES)
+foreach(dir ${dirs})
+    message(STATUS "dir='${dir}'")
+endforeach()
+
+# Add source files to the target
+set(PLATFORM_FILES main.swift SDWEventProperties.swift)
+set(CMAKE_Swift_FLAGS "${CMAKE_Swift_FLAGS} -import-objc-header Swift_Wrapper-Bridging-Header.h")
+add_executable(swift_sample ${PLATFORM_FILES})
+target_link_libraries(swift_sample)
```
ThomsonTan, line 25: This doesn't add any extra property, is it necessary?

absaroj: are you asking about the last line? it i used to create the final exe by linking all the libs.

PR #1141, opened 2023-05-03
adding config class to swift wrappers

absaroj:


wrappers/swift/SDWLogConfiguration.swift
```diff
@@ -0,0 +1,208 @@
...
+public class LogConfiguration {
+    private var odwLogConfiguration: ODWLogConfiguration
+
+    public init(withConfig config: ODWLogConfiguration) {
+        self.odwLogConfiguration = config
+    }
+
+    /// Return a copy of the default configuration
+    public static func getLogConfigurationCopy() -> LogConfiguration? {
+        if let odwLogConfiguration = ODWLogConfiguration.getCopy() {
```
lalitb, line 16: How does this copy works ? If it is creating a shallow copy of [_wrappedConfiguration](https://github.com/microsoft/cpp_client_telemetry/blob/5a511c3e1e05d5c59fcfd025b4c507376705b998/wrappers/obj-c/ODWLogConfiguration.mm#L295), this mayn't be right.

wrappers/swift/SDWLogConfiguration.swift
```diff
@@ -0,0 +1,214 @@
...
+public class LogConfiguration {
+    private var odwLogConfiguration: ODWLogConfiguration
+
+    /**
+    Constructs `self` with provided config as `ODWLogConfiguration` object.
+
+    - Parameters:
+        - config: `ODWLogConfiguration` object which would be wrapped to.
+    */
+    public init(config: ODWLogConfiguration) {
```
eduardo-camacho, line 16: Do this need to be public? I assume this is only used by getLogConfigurationCopy() ?

wrappers/swift/Swift_Wrapper-Bridging-Header.h
```diff
@@ -8,3 +8,6 @@
 #import <Foundation/Foundation.h>
 
 #import "../obj-c/ODWEventProperties.h"
+#import "../obj-c/ODWLogConfiguration.h"
+#import "../obj-c/ODWLogger.h"
```
eduardo-camacho, line 12: The two .h files in lines 12 and 13 are not part of this PR, so compile might break

wrappers/swift/CMakeLists.txt
```diff
@@ -1,25 +1,54 @@
...
     message(STATUS "dir='${dir}'")
 endforeach()
 
 # Add source files to the target
-set(PLATFORM_FILES main.swift SDWEventProperties.swift)
-set(CMAKE_Swift_FLAGS "${CMAKE_Swift_FLAGS} -import-objc-header Swift_Wrapper-Bridging-Header.h")
-add_executable(swift_sample ${PLATFORM_FILES})
-target_link_libraries(swift_sample)
+file(GLOB SWIFT_FILES "${CMAKE_CURRENT_SOURCE_DIR}/*.swift")
+#set(PLATFORM_FILES main.swift SDWEventProperties.swift SDWLogger.swift SDWLogManager.swift)
```
eduardo-camacho, line 45: Since class name is not starting with SDW, let's keep consistency and actually name files EventProperties.swift, Logger.swift and LogManager.swift.
BTW, Logger and LogManager are not part of this PR, so build might fail?

wrappers/swift/SDWLogConfiguration.swift
absaroj: Just tried to follow the similar strategy as what ObjC was following with cpp, as objc is exposing a constructor with cpp LogConfiguration so i kept it public.

wrappers/swift/SDWLogConfiguration.swift
lalitb: Sorry to unresolve it, can you please reply to this comment, as I don't see any change done for it.

absaroj: I thought i did reply to this but perhaps it was not saved.
Its actually creating a new object in objc by passing cpp configurations and returning the newly created object.
i have tried to do the similar thing here with swift.

wrappers/swift/CMakeLists.txt
eduardo-camacho, line 45: You might need to rename here

absaroj: The comment is not needed. Removed it.

wrappers/swift/CMakeLists.txt
```diff
@@ -1,25 +1,53 @@
...
 enable_language(Swift)
 
-# TODO: add logic to build dependencies
+# Set flags and paths
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_Swift_FLAGS "${CMAKE_Swift_FLAGS} -import-objc-header Swift_Wrapper-Bridging-Header.h")
+set(INSTALLED_LIB_PATH "/usr/local/lib")
+
+# Add system libs
+find_library(SYSTEM_CONFIGURATION SystemConfiguration)
```
ThomsonTan, line 22: Should the SystemConfiguration dependency be marked as required?

wrappers/swift/LogConfiguration.swift
```diff
@@ -0,0 +1,214 @@
...
+
+    - Parameters:
+        - config: `ODWLogConfiguration` object which would be wrapped to.
+    */
+    public init(config: ODWLogConfiguration) {
+        self.odwLogConfiguration = config
+    }
+
+    /// Return a copy of the default configuration
+    public static func getLogConfigurationCopy() -> LogConfiguration? {
```
mannie, line 21: What are the scenarios where we aren't able get a copy of the configuration? It may be worth throwing an error if we're unable to retrieve a copy; this is the development paradigm that Swift is moving towards.

wrappers/swift/LogConfiguration.swift
```diff
@@ -0,0 +1,214 @@
...
+        }
+    }
+
+    /**
+    Sets the "URI" of the event collector.
+
+    - Parameters:
+        - collectorUri: `String` for event collector uri.
+    */
+    public static func setEventCollectorUri(_ collectorUri: String!) {
```
mannie, line 35: Consider using computed properties here:
```swift
    static var eventCollectorUri: String {
        get {
            ODWLogConfiguration.eventCollectorUri()
        }
        set {
            ODWLogConfiguration.setEventCollectorUri(newValue)
        }
    }
```

wrappers/swift/LogConfiguration.swift
absaroj: I don't foresee any scenarios here where it might fail as the objc is creating new object by passing cpp log configuration. Also, objc is not doing any exception handling on its side.

wrappers/swift/SDWLogConfiguration.swift
absaroj: resolving this comment. please reopen if any concerns.

wrappers/swift/LogConfiguration.swift
absaroj: Resolving it. Please reopen if any concern.

wrappers/swift/LogConfiguration.swift
```diff
@@ -0,0 +1,214 @@
...
+public class LogConfiguration {
+    private var odwLogConfiguration: ODWLogConfiguration
+
+    /**
+    Constructs `self` with provided config as `ODWLogConfiguration` object.
+
+    - Parameters:
+        - config: `ODWLogConfiguration` object which would be wrapped to.
+    */
+    public init(config: ODWLogConfiguration) {
```
mannie, line 16: Do we envisage clients using this? If not, we should make this private.

wrappers/swift/LogConfiguration.swift
```diff
@@ -0,0 +1,214 @@
+//
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+//
+
+/// Class wrapping `ODWLogConfiguration` ObjC class object, representing configuration related to events.
+public class LogConfiguration {
```
mannie, line 7: Do we expect clients to subclass this? If not, we should mark this as `final` to prevent unexpected behaviors.

wrappers/swift/CMakeLists.txt
```diff
@@ -1,25 +1,53 @@
...
 
 # Add source files to the target
-set(PLATFORM_FILES main.swift SDWEventProperties.swift)
-set(CMAKE_Swift_FLAGS "${CMAKE_Swift_FLAGS} -import-objc-header Swift_Wrapper-Bridging-Header.h")
-add_executable(swift_sample ${PLATFORM_FILES})
-target_link_libraries(swift_sample)
+file(GLOB SWIFT_FILES "${CMAKE_CURRENT_SOURCE_DIR}/*.swift")
+add_executable(swift_sample ${SWIFT_FILES})
+
+target_link_libraries(swift_sample
```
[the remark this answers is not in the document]
absaroj: The output is binary file linking the static lib libmat.a and other required libs to be run by main.swift. This is local cmakelist file used by local script to check the if all swift wrapper code is being compiled and built properly.
main.swift contains test code to test if classes can be used properly, to be run locally on debug machine.

wrappers/swift/LogConfiguration.swift
absaroj: It might get used with other swift wrappers hence keeping it public for now.

mannie: You mean used by other classes in the same package or outside of the package? If inside the package, `internal` should be sufficient.

wrappers/swift/CMakeLists.txt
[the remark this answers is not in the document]
absaroj: Yeah, This is local script which is responsible for just verifying the wrappers code and build for local machine.
I am going to make the global script changes after I submit these PRs, which would have the logic for package manager.

mannie (approved):

eduardo-camacho (approved):

lalitb:
merging to unblock. The build is failing for the old MacOS runner image, and not related to this PR.

PR #1143, opened 2023-05-04
Adding swift wrapper for logmanager and logger classes over Objc

absaroj:
Adding swift wrapper for logmanager and logger classes over Objc .


wrappers/swift/SDWLogManager.swift
```diff
@@ -0,0 +1,182 @@
+//
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+//
+
+/// Wrapper over ODWLogManager which manages the telemetry logging system.
+public class LogManager {
+
+    /**
+    Initializes the telemetry logging system with the speicified tenant token and custom configuration.
```
eduardo-camacho, line 10: typo "speicified"

wrappers/swift/SDWLogManager.swift
```diff
@@ -0,0 +1,182 @@
...
+        }
+    }
+
+    /**
+    Retrieves a new instance of `Logger` for logging telemetry events. Initializes the telemetry logging system,
+    if not already initialized, with the default configuration using the specified tenant token and source.
+
+    - Parameters:
+        - tenantToken: A `String` that contains the tenant token.
+        - withSource: A `String` that contains the name of the source of events.
```
eduardo-camacho, line 33: I'm not quite sure if "with..." applies as well to Swift, like it does with Obj-C, please check Swift parameter naming convention.

wrappers/swift/SDWLogManager.swift
```diff
@@ -0,0 +1,182 @@
...
+    */
+    public static func loggerForSource(source: String!) -> Logger? {
+        if let odwLogger = ODWLogManager.logger(forSource: source) {
+            return Logger(logger: odwLogger)
+        } else {
+            return nil
+        }
+    }
+
+    /// Attempts to send any pending telemetry events that are currently cached either in memory, or on disk.
```
eduardo-camacho, line 86: Functions above have documentation in the form of /*** .. */ while this is just // . Please standardize.

wrappers/swift/SDWLogManager.swift
```diff
@@ -0,0 +1,182 @@
...
+    public static func uploadNow() {
+        ODWLogManager.uploadNow()
+    }
+
+    /// Flushes pending telemetry events from memory to disk (to reduce possible data loss) and returns the flush operation's status.
+    public static func flush() throws -> ODWStatus {
+        do {
+            return ODWLogManager.flush()
+        } catch let exception as NSException {
+            if (LogConfiguration.surfaceCppExceptions()) {
```
eduardo-camacho, line 97: Isn't this bubbling-up/eating behavior already enforced by Obj-C layer? If that's the case, then I don't see why we need to add it here also

wrappers/swift/SDWLogger.swift
```diff
@@ -0,0 +1,237 @@
...
+        self.odwLogger.setContextWithName(name, boolValue: value, piiKind: piiKind)
+    }
+
+    /**
+    Adds (or overrides) a property of context for the telemetry logging system.
+    Context information applies to event generated by the CPP Logger instance, unless it is overwritten on a particular event property.
+
+    - Parameters:
+        - name: A `String` that contains the name of the property.
+        - withDateValue: A `Bool` that contains the property of the value.
```
eduardo-camacho, line 173: This is not a bool

wrappers/swift/SDWLogger.swift
```diff
@@ -0,0 +1,237 @@
...
+enum SDKError: Error {
+    case ONEDSSDKError(reason: String)
+}
+
+/// Wrapper class around ObjC Logger class `ODWLogger` used to events.
+public class Logger {
+    private var odwLogger: ODWLogger;
+
+    /// Constructs `Logger` with `ODWLogger` object which would be wrapped around.
+    public init(logger: ODWLogger) {
```
eduardo-camacho, line 16: Could this not be public?, to avoid passing Obj-C wrapper object, breaking abstraction

wrappers/swift/SDWLogger.swift
eduardo-camacho: Similar case in methods below

wrappers/swift/SDWLogManager.swift
absaroj: I found two ways of adding the comments to the functions as per the conventions on line.
/** */ and ///.
/// I have used only places where comments are concise and don't need the parameters and returns specified separately, rest of the places /** */ is being used.

wrappers/swift/SDWLogManager.swift
absaroj: Yes. Similar logic is being used at objc layer while wrapping cpp exceptions. I think we can remove this one and let the objc exceptions be handled by app

wrappers/swift/SDWLogManager.swift
absaroj: I did found couple of similar names at https://www.swift.org/documentation/api-design-guidelines/. 
so continued using it.
Mostly it emphasized on having good readability while api call.

wrappers/swift/SDWLogger.swift
absaroj: ODWLogger is being init with cpp class logger.
-(instancetype)initWithILogger:(ILogger*)logger

So construction of this class needs ODWLogger (or Ilogger).

wrappers/swift/LogManager.swift
```diff
@@ -0,0 +1,160 @@
...
+    if not already initialized, with the default configuration using the specified tenant token and source.
+
+    - Parameters:
+        - tenantToken: A `String` that contains the tenant token.
+        - withSource: A `String` that contains the name of the source of events.
+
+    - Returns:
+        - A `Logger` instance that points to the logger for the specified tenant token and source.
+    */
+    public static func loggerWithTenant(tenantToken: String!, withSource source: String! = "") -> Logger? {
```
mannie, line 38: Do the parameters on these functions need to be optionals?

wrappers/swift/Logger.swift
```diff
@@ -0,0 +1,226 @@
...
+
+    // MARK: Basic LogEvent methods
+
+    /**
+    Logs a custom event with a specified name.
+
+    - Parameters:
+        - name: A `String` that contains the name of the event.
+    */
+    public func logEventWithName(_ name: String) {
```
mannie, line 23: It may be worth considering something more Swift here: `logEvent(name:)`

wrappers/swift/Logger.swift
```diff
@@ -0,0 +1,226 @@
...
+        self.odwLogger.logEvent(withName: name)
+    }
+
+    /**
+    Logs a custom event with a specified set of properties.
+
+    - Parameters:
+        - eventProperties: The custome event properties, as `EventProperties`.
+    */
+    public func logEventWithProperties(_ properties: EventProperties) {
```
mannie, line 33: It may be worth considering something more Swifty here: `logEvent(properties:)`

wrappers/swift/Logger.swift
```diff
@@ -0,0 +1,226 @@
...
+        - eventProperties: Properties of the failure event, encapsulated within an `EventProperties`.
+    */
+    public func logSessionWithState(_ state: ODWSessionState, eventProperties properties: EventProperties) {
+        self.odwLogger.logSession(with: state, eventProperties: properties.getODWEventPropertiesObject())
+    }
+
+    /**
+    Initializes and gets an instance of Privacy Guard.
+    */
+    public func initializePrivacyGuardWithPrivacyGuardInitConfig(_ initConfigObject: PrivacyGuardInitConfig) {
```
mannie, line 130: It may be worth considering something more Swifty here: `apply(config:)`

wrappers/swift/Logger.swift
```diff
@@ -0,0 +1,226 @@
...
+    /**
+    Adds (or overrides) a property of context for the telemetry logging system.
+    Context information applies to event generated by the CPP Logger instance, unless it is overwritten on a particular event property.
+
+    - Parameters:
+        - name: A `String` that contains the name of the property.
+        - withStringValue: A `String` that contains the property of the value.
+        - withPiiKind: The kind of "Personal Identifiable Information (PII)", as one of the `ODWPiiKind` enum values.
+    */
+    public func setContextWithName(_ name: String!, withStringValue value: String!, withPiiKind piiKind: ODWPiiKind = ODWPiiKind.none) {
```
mannie, line 145: Do the parameters to theses functions need to be optional?

wrappers/swift/LogManager.swift
```diff
@@ -0,0 +1,160 @@
...
+    Initializes the telemetry logging system with the specified tenant token and custom configuration.
+
+    - Parameters:
+        - tenantToken: A `String` that contains the tenant token.
+        - config: A `Dictionary` that contains a custom configuration.
+
+    - Returns:
+        - A `Logger` instance.
+    */
+    public static func initForTenant(tenantToken: String!, config: [AnyHashable: Any]? = nil) -> Logger? {
```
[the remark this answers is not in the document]
absaroj: LogManager class contains all the static methods to control the logging. It also has method to init the Logger and return it. Logger is being created in its class with init.
I have kept it similar to how the design is in Objc classes.

wrappers/swift/LogManager.swift
absaroj: No. Its a mistake. I have corrected it.
Thanks for pointing it out.

lalitb (approved):

wrappers/swift/LogManager.swift
```diff
@@ -0,0 +1,160 @@
...
+
+    - Parameters:
+        - tenantToken: A `String` that contains the tenant token.
+        - withSource: A `String` that contains the name of the source of events.
+        - config: A `Dictionary` that contains a custom configuration.
+
+    - Returns:
+        - A `Logger` instance that points to the logger for the specified tenant token and source.
+    */
+    public static func loggerWithTenant(tenantToken: String, withSource source: String, withConfig config: ODWLogConfiguration!) -> Logger?{
```
mannie, line 58: Let's have this be something more Swifty, please; can we condense these into one function?

```
public static func logger(tenantToken: String?, source: String? = nil, config: ODWLogConfiguration? = nil) throws -> Logger {
    // instantiate and check for value 
    // if we don't have a valid logger, throw OneDSError.unableToCreateLogger
}
```

wrappers/swift/LogManager.swift
```diff
@@ -0,0 +1,160 @@
+//
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+//
+
+/// Wrapper over ODWLogManager which manages the telemetry logging system.
+public class LogManager {
```
mannie, line 7: Do we expect subclasses of this class? If not, we should mark this as `final` to prevent unexpected behaviors if someone chooses to subclass.

wrappers/swift/Logger.swift
```diff
@@ -0,0 +1,226 @@
+//
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+//
+
+/// Wrapper class around ObjC Logger class `ODWLogger` used to events.
+public class Logger {
```
mannie, line 7: Do we expect subclasses of this class? If not, we should mark this as `final` to prevent unexpected behaviors if someone chooses to subclass.

wrappers/swift/Logger.swift
```diff
@@ -0,0 +1,226 @@
...
+    private var odwLogger: ODWLogger;
```
mannie, line 8: nit: we don't need that semi-colon

wrappers/swift/Logger.swift
```diff
@@ -0,0 +1,226 @@
...
+    // MARK: Basic LogEvent methods
+
+    /**
+    Logs a custom event with a specified name.
+
+    - Parameters:
+        - name: A `String` that contains the name of the event.
+    */
+    public func logEvent(name: String) {
+        self.odwLogger.logEvent(withName: name)
```
mannie, line 24: nit: we don't need this `self.` nor the others in the functions that follow

wrappers/swift/SDWLogManager.swift
absaroj: Though i just checked cpp and they seemed to be used ///. So I think we can change it to /// throughout for the consistency.
As that is just about the comments formatting, will take that as separate change.

wrappers/swift/LogManager.swift
absaroj: Thought of doing it but since the  logic to instantiate the logger is bit different (along with error handling) on objc side for these two methods, I kept two separate methods in swift as well.
Condensing these would require a nil check and separate calls to these objc methods so i kept them separate to keep them cleaner.

wrappers/swift/Logger.swift
absaroj: kept it intentionally as convention while referencing the instance vars.

wrappers/swift/Logger.swift
absaroj: Don't see it now but in future it might get subclassed as it has lots of instance methods. So keeping it non-final.

mannie: The problem is that these classes don't seem designed for subclassing since they are modifying class properties instead of instance properties. The state management can get pretty complicated.

wrappers/swift/Logger.swift
mannie: I understand keeping the convention. It's just not the Swift convention. In Swift, when we do `self.` it's a pattern signify value capturing through block :)

wrappers/swift/Logger.swift
absaroj: Not sure if I understand the comment. Can you please explain the class properties vs instance properties? As I see this class is indeed being instantiated and modifying via its instance methods.

mannie: While this class is modifying its internal state via instance variables, there isn't a defined contract on how other classes should subclass this. As a subclasser, will I be required to call super at some point in my implementation? Do I do that at the start or at the end of my implementation? If I override on function, do I need to override the others?

mannie (approved):

ThomsonTan:
There seems some CI failures for iOS build.

eduardo-camacho (approved):

lalitb:
Merging to unblock. iOS build failure is not related to this PR.

PR #1144, opened 2023-05-05
adding privacy and semantic wrappers

absaroj:


wrappers/swift/SDWSemanticContext.swift
```diff
@@ -0,0 +1,131 @@
...
+
+    - Parameters:
+        - userTimeZone: User's time zone relative to UTC, in ISO 8601 time zone format.
+    */
+    public func setUserTimeZone(_ userTimeZone: String!) {
+        self.odwSemanticContext.setUserTimeZone(userTimeZone)
+    }
+
+    /**
+    Specifies a unique user advertizing id to be included with every event.
```
eduardo-camacho, line 78: typo "advertizing"

wrappers/swift/SDWSemanticContext.swift
```diff
@@ -0,0 +1,131 @@
...
+
+    - Parameters:
+        - userAdvertixingID: A `String` that contains the unique user advertising identifier.
+    */
+    public func setUserAdvertisingID(_ userAdvertixingID: String!) {
```
eduardo-camacho, line 83: typo

eduardo-camacho (approved):
Please make sure to get review from Lalit, Sid and Mannie

wrappers/swift/SemanticContext.swift
```diff
@@ -0,0 +1,131 @@
...
+
+    /**
+    Sets the experiment tag (experiment configuration) context information for telemetry events.
+
+    - Note: This method removes any previously stored experiment IDs that were set using `setAppExperimentEtag`.
+
+    - Parameters:
+        - eTag: A `String` that contains the ETag which is a hash of the set of experiments.
+    */
+    public func setAppExperimentETag(_ eTag: String!) {
```
mannie, line 116: Do the parameters in the functions in this file need to be optionals?

wrappers/swift/PrivacyGuard.swift
```diff
@@ -0,0 +1,60 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+//
+
+/// Wrapper to `ODWPrivacyGuard` representing Privacy Guard Hook.
+public class PrivacyGuard {
+    private var odwPrivacyGuard: ODWPrivacyGuard
+
+    /// Constructs object with `ODWPrivacyGuard` object which is to be wrapped around.
+    public init(privacyGuard: ODWPrivacyGuard) {
```
lalitb, line 11: I am surely missing something here, but why do we need this constructor, and also the private instance variable in this class, when all the methods in the class are static. Also, the obj-c class being wrapped around also has only class methods.

wrappers/swift/PrivacyGuard.swift
```diff
@@ -0,0 +1,60 @@
+//
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+//
+
+/// Wrapper to `ODWPrivacyGuard` representing Privacy Guard Hook.
+public class PrivacyGuard {
```
mannie, line 7: Do we expect subclasses of this class? If not, we should mark this as `final` to prevent unexpected behaviors if someone chooses to subclass.

wrappers/swift/PrivacyGuardInitConfig.swift
```diff
@@ -0,0 +1,23 @@
+//
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+//
+
+public class PrivacyGuardInitConfig {
```
mannie, line 6: Do we expect subclasses of this class? If not, we should mark this as `final` to prevent unexpected behaviors if someone chooses to subclass.

wrappers/swift/PrivacyGuardInitConfig.swift
```diff
@@ -0,0 +1,23 @@
...
+    private var odwPrivacyGuardInitConfig: ODWPrivacyGuardInitConfig
```
mannie, line 7: This value is never changed; we should make this a constant instead.

wrappers/swift/PrivacyGuardInitConfig.swift
```diff
@@ -0,0 +1,23 @@
...
+
+    - Parameters:
+        - privacyGuardInitConfig: `ODWPrivacyGuardInitConfig` object which would be wrapped around.
+    */
+    public init(privacyGuardInitConfig: ODWPrivacyGuardInitConfig) {
+        self.odwPrivacyGuardInitConfig = privacyGuardInitConfig
+    }
+
+    /// Returns the `ODWPrivacyGuardInitConfig` representing the corresponding ObjC object.
+    public func getODWPrivacyGuardInitConfig() -> ODWPrivacyGuardInitConfig {
```
mannie, line 20: If we make the underlying variable a public constant, we can remove this function. That's the Swift way of implementing accessors.

wrappers/swift/SemanticContext.swift
```diff
@@ -0,0 +1,131 @@
+//
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+//
+
+/// Wrapper over `ODWSemanticContext` class that manages the inclusion of semantic context values on logged events.
+public class SemanticContext {
```
mannie, line 7: Do we expect subclasses of this class? If not, we should mark this as `final` to prevent unexpected behaviors if someone chooses to subclass.

wrappers/swift/SemanticContext.swift
```diff
@@ -0,0 +1,131 @@
...
+    private var odwSemanticContext: ODWSemanticContext
```
mannie, line 8: `ODWSemanticContext` is a reference type right? If so, we should change this to `private var context: ODWSemanticContext`

wrappers/swift/SemanticContext.swift
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    /**
+    Set the device identifier context information of the telemetry event.
+
+    - Parameters:
+        - deviceID: A unique device identifier, retrieved programmatically where possible and is app/platform speicific.
+    */
+    public func setDeviceID(_ deviceID: String) {
+        self.odwSemanticContext.setDeviceId(deviceID)
```
mannie, line 64: nit: these `self.`s are not needed

wrappers/swift/PrivacyGuard.swift
absaroj: I think you are right here. Currently the constructor for this is not needed as all methods are static. removed it.

wrappers/swift/PrivacyGuardInitConfig.swift
absaroj: It does have some instance method so it might be potentially subclassed in future. so keep it non-final.

wrappers/swift/SemanticContext.swift
absaroj: It has instance methods so potentially might be some advantage to subclassing in future. keeping it non-final.

wrappers/swift/SemanticContext.swift
absaroj: i kept the variable name to reflect that it is objc class instance called by the wrapper.

lalitb (approved):

mannie (approved):

lalitb:
force merge to unblock. The CI is failing for older MacOS runners, not related to the change.

PR #1148, opened 2023-05-09
adding swift wrapper for DiagnosticDataViewer

absaroj:


eduardo-camacho (approved):
Please get review from Lalit and Sid as well

lalitb (approved):
LGTM. Thanks. Is there any plan to add basic tests in [main.swift](https://github.com/microsoft/cpp_client_telemetry/blob/main/wrappers/swift/main.swift) for basic tests on this wrapper ?

wrappers/swift/DiagnosticDataViewer.swift
```diff
@@ -0,0 +1,85 @@
...
+        return ODWDiagnosticDataViewer.currentEndpoint()
+    }
+
+    /**
+    Sets callback for OnDisableNotification event.
+
+    - Parameters:
+        - callback: Code to execute when OnDisableNotification event occurs.
+    */
+    public static func register(OnDisableNotification callback: @escaping () -> Void) {
```
mannie, line 82: Something like `onDisableNotification(run:)` would yield more readable Swift code at the usage site.

absaroj: Thanks for the suggestion. Updated the api.

wrappers/swift/DiagnosticDataViewer.swift
```diff
@@ -0,0 +1,85 @@
+//
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+//
+
+/// Wrapper class over `ODWDiagnosticDataViewer` representing Diagnostic Data Viewer Hook.
+public class DiagnosticDataViewer {
```
mannie, line 7: Do we expect subclasses of this class? If not, we should mark this as `final` to prevent unexpected behaviors if someone chooses to subclass.

wrappers/swift/DiagnosticDataViewer.swift
```diff
@@ -0,0 +1,85 @@
...
+
+    /// MARK: Initialization Methods
+
+    /**
+    Initializes Data Viewer with a specified machine identifier.
+
+    - Parameters:
+        - machineIdentifier: `String` that contains the machine identifier.
+    */
+    public static func initViewerWithMachineIdentifier(machineIdentifier: String) {
```
mannie, line 17: Can we rename this to something like `instantiateViewer(machineIdentifier:)`, please?

wrappers/swift/DiagnosticDataViewer.swift
```diff
@@ -0,0 +1,85 @@
...
+            - Note: This value can be null.
+
+    Execution is dispatched to queue with default priority.
+    */
+    public static func disableViewer(completionWithResult completion: @escaping (Bool) -> Void) {
+        ODWDiagnosticDataViewer.disableViewer(completion)
+    }
+
+    /// Disables Data Viewer.
+    public static func disableViewer() -> Bool {
```
mannie, line 61: The Swifty way of writing getters and setters would be as computed variables; it would be nice for us to leverage that pattern here and in the other PRs :)

absaroj: can you please explain this one? as i see, only viewEnabled can be extracted to computed properties and rest of the viewer methods would need to be as methods.

mannie: My mistake. I meant to put this one function lower.

```
    public static var viewerEnabled: Bool {
         ODWDiagnosticDataViewer.viewerEnabled()
     }

     public static var currentEndpoint: String {
         return ODWDiagnosticDataViewer.currentEndpoint()
     }
```

absaroj: I see, but there are other related methods with these properties which can't be combined into setter computed properties, such as disableViewer.
Converting the just one part of the property into computed ones does not seem correct to me.

mannie: `disableViewer` is a function performing a task, as opposed to a variable holding state. Variables, whether computed or stored, tend to communicate the storage of state in Swift, which functions communicate the performance of an action.

mannie (approved):

eduardo-camacho (approved):

lalitb:
Merging to unblock. The iOS build is stuck not related to this PR.

PR #1151, opened 2023-05-17
Update dependencies

ms-iankudinova:
Update Bond dependencies for the 1DS Test Telemetry Server, because current one is too old and causes complience issue.


lalitb:
@ms-iankudinova Have you validated the test server using new bond version?

ms-iankudinova:
I tried it locally and we are using in it in out integration tests (get the source from the another branch, where I already changed version of dependency).
One thing I haven't tried / test is change in tools/fiddler/OneDSInspector/packages.config

lalitb (approved):

maxgolov (approved):

lalitb:
@ms-iankudinova Would you like to merge this PR ?

PR #1163, opened 2023-06-08
Fix C4777 warning in Pal.cpp, static_cast the return of ::GetCurrentThreadId to unsigned int

mkoscumb:
Fix C4777 warning in Pal.cpp, static_cast the return of ::GetCurrentThreadId to unsigned int


lalitb (approved):
LGTM. Thanks for the fix.

ThomsonTan (approved):

PR #1165, opened 2023-06-08
Convert time from eventPropertiesto ticks

### LINKED ISSUE #1108: [Android] EventInfo.Time is set to year 0001 despite TimeStamp being set correctly on eventproperties
**OneDrive Android** 
We are on [3.7.5.1 ](https://github.com/microsoft/cpp_client_telemetry/releases/tag/v3.7.5.1). We recently updated this version from 3.5.63, which is a massive update worth about two years of changes.

**Steps to reproduce.**
We are sending EventProperties to Logger with the timestamps correctly set to current time in milliseconds. This is what it looks like on our end. The second highlighted line shows `TimeStamp` correctly set with a millisecond time.

I expect that the value passed into `EventProperties.setTimestamp(long)` should be the time reflected in EventInfo.Time

`EventInfo.Time` is incorrect when the data is emitted and picked up in Aria. EventInfo.Time is showing a date in the year 0001. This causes our telemetry tools to reject the data.

This is what the data looks like when I receive it in Aria. I notice that there are two properties marked `TimeStamp`, with the inner `TimeStamp` value being set to a negative number Perhaps this is why EventInfo.Time is incorrect, although I do not know why this happens.

ThomsonTan:
Fixes #1108


lalitb (approved):
Thanks for the fix,

PR #1166, opened 2023-06-09
Adding changes for local build script and cmakelist to create exe using swift package

absaroj:
Changes in local build script and cmakelist file to build swift wrappers package for local sample main.swift file.

Below tasks are done by the cmakelist file:

1. Locate all the required static libs.
2. Run swift build to build swift wrappers package.
3. Combine all object files from swift wrappers to make a static lib.
4. Invoke swiftc compiler to compile the sample swift code while including the modules (wrappers and objc)
and linking the static libs to generate the executable.


absaroj:
Changes in local build script and cmakelist file to build swift wrappers package for local sample main.swift file.

Below tasks are done by the cmakelist file:

# 1. Locate all the required static libs.
# 2. Run swift build to build swift wrappers package.
# 3. Combine all object files from swift wrappers to make a static lib.
# 4. Invoke swiftc compiler to compile the sample swift code while including the modules (wrappers and objc)
#    and linking the static libs to generate the executable.

lalitb:
Can we integrate these build steps with root CMakeLists.txt, and let it build whenever the iOS build is triggered ? This will also let this build trigger from CI.

absaroj:
> Can we integrate these build steps with root CMakeLists.txt, and let it build whenever the iOS build is triggered ? This will also let this build trigger from CI.

absaroj:
> Can we integrate these build steps with root CMakeLists.txt, and let it build whenever the iOS build is triggered ? This will also let this build trigger from CI.

Yeah. I have added the changes to compile the swift wrappers as part of the global build. That is in separate PR.

eduardo-camacho (approved):

wrappers/swift/Sample/main.swift
```diff
@@ -3,8 +3,10 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
+import ObjCModule // Required if Enums defined in ObjC headers are being used
```
mannie, line 6: The enums in `ObjCModule` will almost always be used by the consumers, right? These enums should be in the Swift wrapper; as consumers of the API, we should only have to `import OneDSSwift` and all the relevant dependencies should be pulled in.

wrappers/swift/Sample/main.swift
```diff
@@ -3,8 +3,10 @@
...
+import OneDSSwift // Swift Package containing the wrappers over ObjC
+
 let props = EventProperties(name:"TestEvent")
 props.setProperty("PropName", withValue: ["Type":"SwiftWrappers"])
 props.setProperty("PropWithPII", withInt64Value: Int64(30), withPiiKind: ODWPiiKind.distinguishedName)
-print (props.properties)
-print (props.piiTags)
+print(props.properties())
```
mannie, line 12: Can we have the sample project use the package manifest, please? It would be nice to see how projects will integrate this Swift wrapper. Do we need to move the sample project one level up?

absaroj: I will take that as a separate task and create a project, in examples folder. Planning to keep this swift file as a standalone example of usage.

wrappers/swift/Sample/main.swift
absaroj: I was trying to avoid creating duplicate enums in swift and then do the conversions for them from swift to objc.

absaroj: i have created alias for the types required to be exported in the package and now they would be available without the objc import.

mannie (approved):

lalitb:
Force merge it. The CI failure is not related to this PR.

PR #1167, opened 2023-06-09
Fix variable shadowing warning in LogManagerBase.hpp

mkoscumb:
The `LogManagerBase::instance` variable (line 173) is shadowed by a local variable in LogManagerBase::WaitPause. Rename the local.


lib/include/public/LogManagerBase.hpp
```diff
@@ -716,10 +716,11 @@ namespace MAT_NS_BEGIN
 
         static void WaitPause()
         {
-            auto instance = GetInstance();
-            if (instance) {
+            auto managerInstance = GetInstance();
+            if (managerInstance)
+            {
```
mkoscumb, line 721: Looks like the clang-tidy rules changed this. LMK if you want that reverted.

lalitb (approved):

lib/include/public/LogManagerBase.hpp
lalitb: I think this should be fine.

PR #1168, opened 2023-06-11
Do not call Settings.Secure.ANDROID_ID

anod:
There is a requirement to avoid calling Settings.Secure.ANDROID_ID before user consent,
Add configuration to HttpClient to avoid calling setDeviceInfo in constructor


lalitb (approved):
LGTM.

PR #1170, opened 2023-06-12
Add some const qualifiers identified by static analysis

mkoscumb:


lalitb (approved):

lib/decorators/BaseDecorator.cpp
```diff
@@ -47,7 +47,7 @@ namespace MAT_NS_BEGIN {
 #else
         record.extSdk[0].libVer = PAL::getSdkVersion();
 #endif
-        auto sessionData = m_owner.GetLogSessionData();
+        const auto* sessionData = m_owner.GetLogSessionData();
```
ThomsonTan, line 50: Does this changing the type of `sesstionData` (removing pointer), and is this expected?

ThomsonTan: Ok, seems the variable type is not change, it is just required to be a pointer.

ThomsonTan (approved):
:shipit:

lib/decorators/BaseDecorator.cpp
mkoscumb: Correct, this has no change to the type but it's cv qualifiers.

PR #1175, opened 2023-06-21
adding sample app changes for the swift wrappers

absaroj:


mannie (approved):
For confirmation, this sample project is going to be moved to a different path later, right?

eduardo-camacho (approved):

absaroj:
> For confirmation, this sample project is going to be moved to a different path later, right?

No. This file standalone swift file example to showcase the swift wrappers package usage.
I am creating another example xcodeproject which will import swift package and use the wrappers in "examples" folder.

lalitb (approved):

PR #1176, opened 2023-06-22
Hiding all ObjC types behind Swift wrappers

absaroj:
Public Swift Wrapper API should not be exposing any of the ObjC types to make sure clients are not dependent on ObjC, in case in future we want to bypass ObjC layer and access Cpp apis directly from swift (announced in wwdc23).

- hiding all objc types behind swift and making calls go through swift
- exposed objc types are done via a typealias
- getters and setters are exposed using computed properties
- all methods require objc types in params in swift wrappers are made private

data storage is happening still at the objc layer, but that can be brought back to swift layer later if required (when we plan to bypass objc) due to clients not having objc dependency.


mannie (approved):

wrappers/swift/Sources/OneDSSwift/LogConfiguration.swift
```diff
@@ -7,15 +7,15 @@ import ObjCModule
 
 /// Class wrapping `ODWLogConfiguration` ObjC class object, representing configuration related to events.
 public final class LogConfiguration {
-    private var odwLogConfiguration: ODWLogConfiguration
+    var odwLogConfiguration: ODWLogConfiguration
```
eduardo-camacho, line 10: Why is this no longer private?

eduardo-camacho (approved):

lalitb (approved):

PR #1177, opened 2023-06-23
Replace PAL's to_string with std::to_string.

mkoscumb:
Not exactly sure why this method exists. I'd like to replace the toString methods across the board (except for our own types), but that's probably a downstream change.


lalitb (approved):
Can't see from history why these `toString` methods were added, but the changes look good to internally use `std::to_string()` here.

PR #1183, opened 2023-06-29
Adding sample swift package importing swift wrappers package

absaroj:


examples/swift/SamplePackage/Package.swift
```diff
@@ -0,0 +1,29 @@
...
+let package = Package(
+    name: "SamplePackage",
+    products: [
+        // Products define the executables and libraries a package produces, making them visible to other packages.
+        .library(
+            name: "SamplePackage",
+            targets: ["SamplePackage"]),
+    ],
+    dependencies: [
+        .package(path: "../../../wrappers/swift"),
```
mannie, line 15: In our logging library, it would therefore be sufficient to checkout the source (in say a submodule) and point this to the correct location, right?

mannie (approved):

lalitb (approved):

PR #1186, opened 2023-07-11
Flush before FlushAndTeardown in docs + examples

n8ta-msft:
Document this unexpected api requirement


examples/cpp/SampleCpp/main.cpp
```diff
@@ -387,6 +388,8 @@ int main()
     // normal	- 4 sec
     // low		- 8 sec
 
+    printf("LogManager::Flush\n");
```
lalitb, line 391: Is this `printf` intended?

lib/include/public/ILogManager.hpp
```diff
@@ -49,6 +49,9 @@ namespace MAT_NS_BEGIN
     {
        public:
         /// <summary>
+        /// WARNING: If you are using offline storage you need to call Flush before FlushAndTearDown.
+        /// https://github.com/microsoft/cpp_client_telemetry/blob/main/docs/use-after-free.md#flushandteardown
+        ///
```
lalitb, line 54: While you are here, can you change the below comment too as `FlushAndTearDown` doesn't necessarily flush to persistent storage? I think just "tear-down the telemetry logging system" should be enough.

examples/cpp/SampleCpp/main.cpp
n8ta-msft: I was just matching what was there for `    printf("LogManager::FlushAndTeardown\n");` I can remove it if you'd like

lib/include/public/ILogManager.hpp
n8ta-msft: Do you want me to modify the Flush() comment or remove the "Flushes any pending telemetry events in memory to disk" part of this comment?

lalitb (approved):

ThomsonTan (approved):
:shipit:

examples/cpp/SampleCppMini/main.cpp
```diff
@@ -88,6 +88,7 @@ void test_cpp_api(const char * token, int ticketType, const char *ticket)
     logger->LogEvent(detailed_event);
 
     // Shutdown
+    LogManager::Flush();
```
maxgolov, line 91: There's really something broken in the SDK if it requires `Flush` before `FlushAndTeardown`. The intent of `FlushAndTeardown` to perform `Flush`, wait for (configurable) N seconds, then `Teardown`.

maxgolov:
> Document this unexpected api requirement

It's a bug.

examples/cpp/SampleCppMini/main.cpp
lalitb: Yes, that was what I thought too. But I don't see that's happening in FlushAndTeardown by looking into the code. Also we let this get documented here - https://github.com/microsoft/cpp_client_telemetry/blob/main/docs/use-after-free.md#flushandteardown

``
If one wishes to use this backstop to quiesce the SDK, the application should call Flush before FlushAndTeardown. This looks redundant (doesn't FlushAndTeardown actually flush, you might ask), but despite the method name FlushAndTeardown, the call to Flush is required to ensure persistence before the PauseActivity quiesce.
``

n8ta-msft:
> > Document this unexpected api requirement
> 
> It's not unexpected requirement, it looks like a bug to me that was probably introduced around October 2021+

Glad to hear it's a bug, feel free to close my PR. Let me know if you'd like me to file an issue.

examples/cpp/SampleCppMini/main.cpp
maxgolov: I think this document was actually contributed at the same time when things got broken.. 🤣 I have a theory that this got broken by the zombie-logger change. My team uses a slightly older version of SDK, and we do not see the issue.

lalitb:
> Glad to hear it's a bug, feel free to close my PR. Let me know if you'd like me to file an issue.

Thanks @n8ta-msft. Closing this. Feel free to file  a bug for better tracking :)

PR #1188, opened 2023-07-27
Fix: Issue 1185 - make HttpClient_Android singleton

abuhsayem:
Make HttpClient_Android a singleton object so that multiple instances are not created.

Two HttpClient_Android objects are created (one is created by 1DS SDK and another by UTel SDK using HttpClient.java’s native/JNI createClientInstance() function in constructor). The Android Designer app immediately sends an event ‘Boot’ upon start that is probably causing the race condition and hence these two objects. The events are forwarded from HttpClient_Android to HttpClient.java to send to collector, then using JNI dispatchCallback() function is called of HttpClient_Android (the right object is not called and hence the debug message ‘Maximum number of HTTP requests reached’).
Making the HttpClient_Android class a singleton object fixes this issue (my change is to make it singleton only).

### DIFF
--- a/lib/http/HttpClient_Android.cpp
+++ b/lib/http/HttpClient_Android.cpp
@@ -12,6 +12,7 @@
 namespace MAT_NS_BEGIN
 {
     constexpr static auto Tag = "HttpClient_Android";
+    std::once_flag flagClient;
 
     HttpClient_Android::HttpRequest::~HttpRequest() noexcept
     {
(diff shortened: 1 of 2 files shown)


lalitb:
@abuhansatsayem Thanks for the PR. Can you please add some details on the reason for change. Looks like this is trying to fix the race condition where application is creating multiple HttpClient_Android instance ?

abuhsayem:
@lalitb @anod 
Please review the PR at your earliest convenience. It is a blocker for 'Designer Android' Mobile Private Preview using our 'UTel SDK'

abuhsayem:
@lalitb yes, I will join the community meeting tomorrow

anod:
it sounds that having single instance of java http client is a current design limitation. 
Since it is possible to create multiple LogManagers it should be possible to have a separate instances of Http client. So every log manager instance should have a single http client, but not a singleton.

abuhsayem:
Highlights from '1DS Community Teams Meeting' on Tuesday, August 1st, 2023  
* Making 'HttpClient_Android' singleton is not a viable solution
* 'HttpClient_Android' has a static object 's_client', as a result only one DefaultEventSink can be active 
   static std::shared_ptr<HttpClient_Android> s_client;
*  'HttpClient_Android' function 'GetClientInstance()' returns an static object as well
   static std::shared_ptr<HttpClient_Android> GetClientInstance();
* Propose 'Designer Android App' to use one and not two 'TelemetryService' objects from 'UTel SDK'. Thus, having only one DefaultEventSink active at anytime.

abuhsayem:
Closing/abandoning the PR since its not a viable solution

PR #1191, opened 2023-08-04
EUDB compliance recommendations and example

maxgolov:
Certain Microsoft applications are in scope to be European Union Data boundary (EUDB) Compliant.

This FAQ for 1DS SDK addresses one of the 8 EUDB requirements:

- Data Residency which requires all EU data accessed, processed, displayed/viewed with certain applications must reside within EU. 1DS Collector itself ensures that the data routed to regional collection end-point is maintained in EU. Products must perform determination whether or not a specific installation / seat within a Commercial / Enterprise setting belongs to EU or not. This determination process is out-of-scope for 1DS SDK itself, and may be performed via another systems/APIs, such as ECS, OneSettings, MSGraph, etc.

The goal of this PR is:

- to provide an example how to configure 1DS C++ SDK to respect a regional URL.
- how to perform the asynchronous update of the URL at runtime.
- additional safety check to ensure that the SDK could be started with an empty URL: collection disabled until the URL is known. Although `PauseTransmission` called on app start is generally safe to use, as default transmission timer is at least 1 second, unfortunately it does not reliably cover the "immediate" priority events, which could be emitted by application early at boot. Thus, the need for more reliable mechanism that allows to initialize in a fully "offline" state, functionally similar to "paused" state, without necessity of calling into `PauseTransmission`. This is solving a potential race condition between the main thread calling `PauseTransmission` and a background thread emitting a critical realtime priority event.

Out-of-scope:

- There is no guidance how to actually obtain the EUDB URL. This is out-of-scope for 1DS client SDK.
- There is no guidance what option to prefer: having 2 instances (one for US, one for EU); or having 1 instance that gets adjusted to use the regional EUDB URL at runtime. The choice to use this or that approach is up to concrete product developers.

### DIFF
--- a/docs/EUDB-compliance.md
+++ b/docs/EUDB-compliance.md
@@ -1,0 +1,65 @@
+# EUDB Guidance for 1DS C++ SDK
+
+In order to satisfy the Microsoft commitment to ensuring Privacy and Compliance, specifically
+EUDB compliance with respect to EU 'Schrems II' decision, it is imperative for certain
+commercial products to perform EUDB URL upload determination during application launch.
+
+1DS Collector service accepts data from all One Observability client SDKs. By default, traffic
+simply flows to whichever region can best handle the traffic. This approach works well for
+system required metadata. However some client scenarios require that data is sent to a specific
+geographic location only.
+
+1DS C++ SDK supports ability to specify / adjust the upload URL at runtime.
+
+Two approaches could be applied to implement EUDB-compliant data upload.
+
+## Option 1: Create two instances of 1DS C++ SDK - one for US collector, another for EU collector
+
+See [Multiple Log Managers Example](https://github.com/microsoft/cpp_client_telemetry/tree/main/examples/cpp/SampleCppLogManagers)
+that illustrates how to create multiple instances, each acting as a separate vertical pillar with
+their own data collection URL. Two instances `LogManagerUS` and `LogManagerEU` may be configured
+each with their own data collection URL, for example:
+
+- For US customers: `https://us-mobile.events.data.microsoft.com/OneCollector/1.0/`
+- For EU customers: `https://eu-mobile.events.data.microsoft.com/OneCollector/1.0/`
+
+Depending on data requirements and outcome of dynamic EUDB determination, i.e. organization /
+M365 Commercial Tenant is located in EU, the app decides to use `LogManagerEU` instance for
+telemetry. Default `LogManager` instance can still be used for region-agnostic "global"
+collection of required system diagnostics data. Remember to use the proper compliant instance
+depending on event type.
+
+## Option 2: Autodetect the corresponding data collection URL on app start
+
+EventSender example has been modified to illustrate the concept:
+
+- Application starts.
+
+- `LogManager::Initialize(...)` is called with `ILogConfiguration[CFG_STR_COLLECTOR_URL]` set to
+empty value `""`. This configuration instructs the SDK to run in offline mode. All data gets
+logged to offline storage and not uploaded. This setting has the same effect as running in
+paused state. Key difference is that irrespective of upload timer cadence - even for immediate
+priority events, 1DS SDK never attempts to trigger the upload. This spetial configuration option
+is safer than simply issuing `PauseTransmission` on app start.
+
+Then application must perform asynchronous EUDB URL detection in its own asynchronous task /
+thread. URL detection process is asynchronous and may take significant amount of time from hundred
+milliseconds to seconds. In order to avoid affecting application launch startup performance,
+application may perform other startup and logging actions concurrently. All events get logged
+in offline cache.
+
+- As part of the configuration update process - application calls `LogManager::PauseTransmission()`
+done to ensure exclusive access to uploader configuration.
+
+- Once the EUDB URL is obtained from remote configuration provisioning service (ECS, MSGraph,
+OneSettings, etc.), or read cached value from local app configuration storage, the value is supplied
+to 1DS SDK:
+
+`ILogConfiguration[CFG_STR_COLLECTOR_URL] = eudb_url`
+
+Note that 1DS SDK itself does not provide a feature to store the cached URL value. It is up to the
+product owners to decide what caching mechanism they would like to use: registry, ECS cache, Unity
+player settings, mobile app settings provider, etc.
+
+- Finally the app code could call `LogManager::ResumeTransmission()` - to apply the new configuration
+settings and enable the data upload to compliant destination.
--- a/docs/EUDB-compliance.md
+++ b/docs/EUDB-compliance.md
@@ -39,7 +39,7 @@ EventSender example has been modified to illustrate the concept:
 empty value `""`. This configuration instructs the SDK to run in offline mode. All data gets
 logged to offline storage and not uploaded. This setting has the same effect as running in
 paused state. Key difference is that irrespective of upload timer cadence - even for immediate
-priority events, 1DS SDK never attempts to trigger the upload. This spetial configuration option
+priority events, 1DS SDK never attempts to trigger the upload. This special configuration option
 is safer than simply issuing `PauseTransmission` on app start.
 
 Then application must perform asynchronous EUDB URL detection in its own asynchronous task /
--- a/docs/EUDB-compliance.md
+++ b/docs/EUDB-compliance.md
@@ -42,7 +42,7 @@ paused state. Key difference is that irrespective of upload timer cadence - even
 priority events, 1DS SDK never attempts to trigger the upload. This special configuration option
 is safer than simply issuing `PauseTransmission` on app start.
 
-Then application must perform asynchronous EUDB URL detection in its own asynchronous task /
+Then application must perform asynchronous EUDB URL detection once in its own asynchronous task /
 thread. URL detection process is asynchronous and may take significant amount of time from hundred
 milliseconds to seconds. In order to avoid affecting application launch startup performance,
 application may perform other startup and logging actions concurrently. All events get logged
@@ -56,6 +56,9 @@ OneSettings, etc.), or read cached value from local app configuration storage, t
 to 1DS SDK:
 
 `ILogConfiguration[CFG_STR_COLLECTOR_URL] = eudb_url`
+
+This assignment of URL is done once during application start. Application does not need to change the
+data collection URL after that.
 
 Note that 1DS SDK itself does not provide a feature to store the cached URL value. It is up to the
 product owners to decide what caching mechanism they would like to use: registry, ECS cache, Unity
--- a/examples/cpp/EventSender/EventSender.cpp
+++ b/examples/cpp/EventSender/EventSender.cpp
@@ -79,6 +79,8 @@ const char* defaultConfig = static_cast<const char *> JSON_CONFIG
 // is performed asynchronously and could take a few seconds. EUDB URL for Enterprise applications may be cached
 // in app-specific configuration storage. 1DS SDK does not provide a feature to cache the data collection URL used for
 // a previous session.
+//
+// Note that this function to determine the URL is called once, early at boot.
 std::string GetEudbCollectorUrl()
 {
     const auto randSeed = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch()).count();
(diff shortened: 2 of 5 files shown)


docs/EUDB-compliance.md
```diff
@@ -0,0 +1,65 @@
...
+- For US customers: `https://us-mobile.events.data.microsoft.com/OneCollector/1.0/`
+- For EU customers: `https://eu-mobile.events.data.microsoft.com/OneCollector/1.0/`
+
+Depending on data requirements and outcome of dynamic EUDB determination, i.e. organization /
+M365 Commercial Tenant is located in EU, the app decides to use `LogManagerEU` instance for
+telemetry. Default `LogManager` instance can still be used for region-agnostic "global"
+collection of required system diagnostics data. Remember to use the proper compliant instance
+depending on event type.
+
+## Option 2: Autodetect the corresponding data collection URL on app start
```
lalitb, line 32: Is this logic applicable during application start, or else can something below done while event transmission is happening:

 `url_change_detected` -> `PauseTransmission` -> `update Logconfiguration::url` -> `ResumeTransmission` 

Or we can't recommend this as there is the possibility of events getting leaked to incorrect endpoint.

examples/cpp/EventSender/EventSender.cpp
```diff
@@ -71,6 +72,38 @@ const char* defaultConfig = static_cast<const char *> JSON_CONFIG
...
 );
 
+// Mock function that performs random selection of destination URL. 1DS SDK does not define how the app needs to perform
+// the region determination. Products should use MSGraph API, OCPS, or other remote config provisioning sources, such as
+// ECS: https://learn.microsoft.com/en-us/deployedge/edge-configuration-and-experiments - in order to identify what 1DS
+// collector to use for specific Enterprise or Consumer end-user telemetry uploads. Note that the EUDB URL determination
+// is performed asynchronously and could take a few seconds. EUDB URL for Enterprise applications may be cached
+// in app-specific configuration storage. 1DS SDK does not provide a feature to cache the data collection URL used for
+// a previous session.
+std::string GetEudbCollectorUrl()
```
lalitb, line 82: Can this example be generic enough, without naming, and mentioning the EuDB terms in comment and variable names?

docs/EUDB-compliance.md
```diff
@@ -0,0 +1,65 @@
...
+each with their own data collection URL, for example:
+
+- For US customers: `https://us-mobile.events.data.microsoft.com/OneCollector/1.0/`
+- For EU customers: `https://eu-mobile.events.data.microsoft.com/OneCollector/1.0/`
+
+Depending on data requirements and outcome of dynamic EUDB determination, i.e. organization /
+M365 Commercial Tenant is located in EU, the app decides to use `LogManagerEU` instance for
+telemetry. Default `LogManager` instance can still be used for region-agnostic "global"
+collection of required system diagnostics data. Remember to use the proper compliant instance
+depending on event type.
```
lalitb, line 30: It may be worth mentioning that creating multiple LogManagers can have performance overhead related to memory and CPU usage. Hence, it's recommended that applications perform local testing to ensure the configuration is appropriate and matches their requirements.

docs/EUDB-compliance.md
```diff
@@ -0,0 +1,65 @@
...
+In order to satisfy the Microsoft commitment to ensuring Privacy and Compliance, specifically
+EUDB compliance with respect to EU 'Schrems II' decision, it is imperative for certain
+commercial products to perform EUDB URL upload determination during application launch.
+
+1DS Collector service accepts data from all One Observability client SDKs. By default, traffic
+simply flows to whichever region can best handle the traffic. This approach works well for
+system required metadata. However some client scenarios require that data is sent to a specific
+geographic location only.
+
+1DS C++ SDK supports ability to specify / adjust the upload URL at runtime.
```
lalitb, line 12: Should it be at runtime, or only once during startup from `""` to (say) `url1`? I do suspect that changing at runtime from (say) `url1` to `url2` can result in events leaking to incorrect endpoints.

lib/tpm/TransmissionPolicyManager.cpp
```diff
@@ -107,6 +107,11 @@ namespace MAT_NS_BEGIN {
             return;
         }
         LOCKGUARD(m_scheduledUploadMutex);
+        if (!m_config.IsCollectorUrlSet())
+        {
+            LOG_TRACE("Collector URL is not set, no upload.");
+            return;
```
lalitb, line 113: Will it change the existing behavior of the telemetry? As of now, most of the applications won't be setting the URL assuming that the default COLLECTOR_URL_PROD URL would be used. Now, we won't do any upload if the URL is not set.

maxgolov: No, it doesn't change the current behavior. It was my intent to avoid any functional changes per se.

If customer simply doesn't specify the URL in their configuration, then default IRuntimeConfig contains the global end-point from here : https://github.com/microsoft/cpp_client_telemetry/blob/c7bb501a159341c885c294f7d3a0301f6e8768ba/lib/api/LogConfiguration.cpp#L28C1-L29C1
Everything works as before.

Only if customer obtains the default configuration using `LogManager::GetLogConfiguration()`, and then intentionally and explicitly adjusts it to set the URL to empty string, then this new safety check skips the upload. What would've happened without this safety check: upload thread starts, and HTTP client would fail - since the URL is invalid.

So it doesn't affect the current feature operation in any way. It only improves safety for the case when the URL has been explicitly set to none.

docs/EUDB-compliance.md
maxgolov: Currently there's no protection that would copy the URL on start. The uploader will always "reread" the config today. Expectation here is that the product owners don't try to shoot themselves in the foot. "At runtime" here - means that we initialize, all logging functionality works, and URL is set at some point shortly after (hundreds of milliseconds to a few seconds)... What would happen if URL cannot be determined, i.e. app is offfline, or the corresponding remote service is unreachable? That is left to the product owners to decide. Currently if they set the URL to none, and it is kept that way, that's, perhaps, the most compliant action: do not attempt to upload any telemetry until you reliably know what regional collector you need to upload it to.

Whether or not someone is going to change the URL repeatedly at runtime more than once - that'd be morally wrong to do so. Example illustrates that the determination has to be performed once. Users belong to org, org belongs to fixed region. There is no scenario where user would roam between different environments or regions in my product. In case if users roam between environments (but not for EUDB), that's the option where multiple LogManagers , `LogManagerX` and `LogManagerY` can be used simultaneously.

docs/EUDB-compliance.md
maxgolov: I think in the other prominent product - there are practical reasons why they'd run with multiple instances (e.g. switching between different environments, or using a separate collector URL for customer content data). I don't want to enforce this or that approach. Basically, both approaches: either two instances - each separate collector URL; OR one instance with collector URL determined at runtime - both valid approaches. It really depends on product requirements what approach a particular engineering team is going to take.

examples/cpp/EventSender/EventSender.cpp
maxgolov: I can explain why I kept the narrow focus on EUDB scenario only: if we generalize this to another case, where users are switching environments for non-EUDB reason (private vs public, and/or system metadata vs customer content collection), then it may be more appropriate for that other non-EUDB scenario to use the multiple log managers approach instead of dynamically changing the URL.. Then we'd have to explain and expand on why/when you'd use one-vs-two, separate vertical pillars, separate storage files, etc. It is out-of-scope for me at the moment. I'd like to solve only the EUDB issue.

docs/EUDB-compliance.md
maxgolov: I can elaborate on this further in the document if you like

- `initialize` with no URL - functionally equivalent to "paused" state.
- `pause` - to commit changes to uploader configuration in thread-safe manner.
- `detect` the url ONCE, after the URL determination is complete. It's once only, typically shortly after app start, and/or immediately after the app comes online, and once user is logged in. Set the URL in configuration.
- `resume` - for new uploader configuration to take effect.

There's no intention to run a periodic URL detection. It's done once, typically after user is logged in, OR when the url for a given application (seat) is known based on some other configuration cache.

docs/EUDB-compliance.md
maxgolov: Current code comment before my change - says `Since this method is called for every event, you can change the URI dynamically.`, which I think is semantically similar to `ability to specify / adjust the upload URL at runtime`.

https://github.com/microsoft/cpp_client_telemetry/blob/c7bb501a159341c885c294f7d3a0301f6e8768ba/lib/api/IRuntimeConfig.hpp#L27

```
         /// <summary>
        /// Gets the URI of the collector (where telemetry events are sent).
        /// </summary>
        /// <remarks>
        /// <b>Note:</b> Since this method is called for every event, 
        /// you can change the URI dynamically.
        /// </remarks>
        /// <returns>A string that contains the collector URI.</returns>
```

lib/tpm/TransmissionPolicyManager.cpp
```diff
@@ -107,6 +107,11 @@ namespace MAT_NS_BEGIN {
             return;
         }
         LOCKGUARD(m_scheduledUploadMutex);
+        if (!m_config.IsCollectorUrlSet())
```
ThomsonTan, line 110: >m_config.IsCollectorUrlSet() [](http://example.com/codeflow?start=13&length=28)

Should this be guarded by the above m_scheduledUploadMutex?

examples/cpp/EventSender/EventSender.cpp
```diff
@@ -71,6 +72,38 @@ const char* defaultConfig = static_cast<const char *> JSON_CONFIG
...
+{
+    const auto randSeed = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch()).count();
+    srand(static_cast<unsigned>(randSeed));
```
ThomsonTan, line 85: >srand [](http://example.com/codeflow?start=4&length=5)

Call `srand` once instead of every time when`GetEudbCollectorUrl` is called?

maxgolov: Technically it's called once only. This is just a way to randomize the URL selection. Products must implement the actual logic to deterministically obtain the correct URL.

lib/tpm/TransmissionPolicyManager.cpp
maxgolov: Not necessarily. This mutex used to be on top of the method, guarding the rest of checks.

`PauseGuard` then for some reason was added later - above this mutex:

image

If you have a strong preference, I can move the check above the mutex. It won't change anything functionality, nor from perf standpoint since the method call is made not more than once in 1 - 2 - 4 - 8 seconds.

docs/EUDB-compliance.md
lalitb: Thanks, @magolov, should we explicitly mention that the URL should be changed only one during startup, and NOT afterward?

docs/EUDB-compliance.md
lalitb: Thanks for clarifying. So to summarise " While 1ds sdk supports changing endpoint URL at runtime, the application users should be aware of its consequences.

lib/tpm/TransmissionPolicyManager.cpp
lalitb: Thank for explaining, this clarifies.

lib/tpm/TransmissionPolicyManager.cpp
lalitb: Agree with @ThomsonTan, We can move it out of the lock if it really doesn't require a lock irrespective of how minimal savings we have. Agree that is called from the upload thread, and not the hot path of event generation so not called very frequently, but it is still better to remove it if it doesn't belong there.

lalitb:
Thanks for documenting, and adding a fix to pause the transmission during the start. It would be good to explicitly document that for EuDB scenarios, the URL should be updated only once during startup, and not afterwards. And be aware of the consequences of doing it.

lalitb (approved):

lib/tpm/TransmissionPolicyManager.cpp
maxgolov: Sounds good. I'll move it out of the lock.

docs/EUDB-compliance.md
maxgolov: Sounds good. I'll add an explanation for this in the markdown.

maxgolov: Added an additional explanation that the URL is determined and set once only, during app start.

lib/tpm/TransmissionPolicyManager.cpp
maxgolov: Resolved as suggested.

examples/cpp/EventSender/EventSender.cpp
maxgolov: I added a code comment to explain why / when.

PR #1198, opened 2023-08-31
Expose FlushAndTearDown API in C-API 

shivpratap1992:


lib/include/public/mat.h
```diff
@@ -59,7 +59,8 @@ extern "C" {
         EVT_OP_FLUSH = 0x0000000A,
         EVT_OP_VERSION = 0x0000000B,
         EVT_OP_OPEN_WITH_PARAMS = 0x0000000C,
-        EVT_OP_MAX = EVT_OP_OPEN_WITH_PARAMS + 1
+        EVT_OP_MAX = EVT_OP_OPEN_WITH_PARAMS + 1,
+        EVT_OP_FLUSHANDTEARDOWN = EVT_OP_MAX + 1
```
lalitb, line 63: Change the above two lines to:
```cpp
EVT_OP_FLUSHANDTEARDOWN = 0x0000000D,
EVT_OP_MAX = EVT_OP_FLUSHANDTEARDOWN + 1
```

lib/include/public/mat.h
```diff
@@ -576,6 +577,22 @@ extern "C" {
         ctx.handle = handle;
         return evt_api_call(&ctx);
     }
+
+    /** <summary>
+     * Provide a hint to telemetry system to attempt force-upload of events
+     * based on provided tear down interval before shutdown. This API does not
+     * guarantee the upload.
+     * </summary>
```
lalitb, line 585: Change the summary to:
`Flush any pending telemetry events in memory to disk, attempt upload of events if tear down interval is configured, and eventually tear down the telemetry logging system.`

lalitb:
Can you also add a unit test at:
 https://github.com/microsoft/cpp_client_telemetry/blob/6129741748f9627d765f315a9f83561c70366739/tests/functests/APITest.cpp#L619 
Basically calling this API just before [evt_close](https://github.com/microsoft/cpp_client_telemetry/blob/6129741748f9627d765f315a9f83561c70366739/tests/functests/APITest.cpp#L728) atleast to ensure there is no crash from the API.

lalitb (approved):
Thanks.

maxgolov:
@shivpratap1992 @lalitb

I think this fix is doing something wrong : `evt_close` does ref-counting and performs `FlushAndTeardown` when necessary, when the last instance is gone. In host-guest scenarios, when two entities are anchored to the same `LogManager` instance - allowing one of them to explicitly destroy the `LogManager` via `FlushAndTeardown` may lead to catastrophic results. For example, in case if guest calls the teardown C API.

Was there an issue with `evt_close` not properly invoking the teardown seqeuence?

PR #1217, opened 2023-10-03
Add target .net 6 for the 1DS Test Server

ms-iankudinova:
In our e2e tests, when we are using 1DS Test Server, pipeline uses .net6 (we have been targeting custom branch with the change `user/iankudinova/update_server_to_net6`).
I want to add the support .net6 to the main branch.
In case at some other project / plcases .net core 3 is used for the Test telemetry server I kept both version.

So Users will have to specify --framework version, when they run server.
Existed scripts: 
- `run.cmd`
- `run.sh`
were updated to support --framework option and use .net core 3 as a default option

### DIFF
--- a/tools/server/run.sh
+++ b/tools/server/run.sh
@@ -1,3 +1,39 @@
 #!/bin/sh
-dotnet run
+
+############################################################
+# Help                                                     #
+############################################################
+Help()
+{
+   # Display Help
+   echo "Run 1DS Test Telemetry Server with specified runtime versions (.net core 3 or .net6)."
+   echo
+   echo "Syntax: run.sh [-f,--framework|-h, --help]"
+   echo "options:"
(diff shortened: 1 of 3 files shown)


tools/server/run.sh
```diff
@@ -1,3 +1,39 @@
...
+while test $# -gt 0;
+do
+  case "$1" in
+    -h|--help) Help;;
+    -f|--framework)       
+      shift
+      if test $# -gt 0; then
+        framework=$1
+      else
+        echo "framework is not specified. Server will be run with .netcoreapp3.1"
```
ThomsonTan, line 30: >.netcoreapp3.1 [](http://example.com/codeflow?start=66&length=14)

Reference variable `framework` here?

ms-iankudinova: DOne -> https://github.com/microsoft/cpp_client_telemetry/blob/2c5c3d61cdab297ea336bb7e9dc60d661591d0d7/tools/server/run.sh#L30

maxgolov (approved):

PR #1222, opened 2023-11-07
Upgrade zlib

shivpratap1992:


lalitb (changes requested):
This is not the right way to upgrade zlib. You need to get the latest/required version from zlib repo, and merge it with the intel optimization patches.

PR #1225, opened 2023-11-23
POC: Rust Wrapper for 1DS C++ SDK

CDFriend:
## What is this?

Adds a Rust "wrapper" to the 1DS C++ SDK, which will allow the SDK to be imported into Rust projects as a crate. The crate interacts with the C API of the SDK (instead of C++) as this will ensure a stable binary interface between the Rust layer and C++ layer of the project.

## Project Structure

Crates:
* cpp-client-telemetry-sys: crate directly exposing the C API of the 1DS SDK to Rust. 
* oneds-telemetry: contains some abstractions over the 1DS SDK and allows messages to be sent to the SDK from the log crate.
* telemetry-sample: example executable using the oneds-telemetry crate.

## Limitations

* Currently only `i64` and `string` telemetry fields are supported.
* Currently we're only linking with the Windows-x64 static library - additional work on the build script will be required to add support for dynamic linking and other architectures/operating systems.


CDFriend:
@microsoft-github-policy-service agree company="Microsoft"

wrappers/rust/oneds-telemetry/src/conversion.rs
```diff
@@ -0,0 +1,49 @@
...
+use std::mem;
+use std::os::raw::c_char;
+
+/**
+ * This will leak memory
+ */
+pub fn to_leaky_c_str(value: &str) -> *const i8 {
+    let c_str = Box::new(CString::new(value.clone()).unwrap());
+    let ptr = c_str.as_ptr() as *const i8;
+    mem::forget(c_str);
```
sivadeilra, line 12: Can this be `Box::into_raw(c_str) as *const i8` instead?

wrappers/rust/oneds-telemetry/src/lib.rs
```diff
@@ -0,0 +1,161 @@
...
+
+#[derive(Clone)]
+pub struct LogManager {
+    is_started: bool,
+    configuration: CString,
+    log_handle: Option<i64>,
+}
+
+/// The global `Tracer` provider singleton.
+static GLOBAL_LOG_MANAGER: Lazy<RwLock<LogManager>> = Lazy::new(|| RwLock::new(LogManager::new()));
```
sivadeilra, line 29: Is there a reason this uses `Lazy` instead of just typing this field as `RwLock<LogManager>`?

wrappers/rust/oneds-telemetry/src/types.rs
```diff
@@ -0,0 +1,182 @@
...
+    pub fn as_ptr(&self) -> *const c_char {
+        self.bytes.as_ptr() as *const c_char
+    }
+
+    pub fn as_bytes(&self) -> Vec<u8> {
+        self.bytes.clone()
+    }
+
+    fn to_str(&self) -> Result<&str, std::str::Utf8Error> {
+        unsafe { CStr::from_ptr(self.as_ptr()).to_str() }
```
sivadeilra, line 36: Is there a reason this is not just `core::str::from_utf8(&self.bytes)`?  What merits using unsafe code here?

wrappers/rust/oneds-telemetry/src/types.rs
```diff
@@ -0,0 +1,182 @@
...
+    }
+}
+
+pub static NAME_PROPERTY: Lazy<StringProperty> = Lazy::new(|| StringProperty::new("name"));
```
sivadeilra, line 40: Can these fields be converted to `StringProperty`, instead of using `Lazy`?  You may be able to changes the `bytes` field of `StringProperty` to use `Cow<'static, [u8]>` (where `Cow` is from `std::borrow::Cow`), so that you can initialize it in a `const fn` context.

wrappers/rust/oneds-telemetry/src/types.rs
```diff
@@ -0,0 +1,182 @@
...
+
+use cpp_client_telemetry_sys::{evt_prop, evt_prop_v, evt_prop_t_TYPE_STRING, evt_prop_t_TYPE_TIME, evt_prop_t_TYPE_INT64};
+
+#[derive(Clone, Debug)]
+pub struct StringProperty {
+    bytes: Vec<u8>,
+}
+
+impl StringProperty {
+    pub fn new(value: &str) -> Self {
```
sivadeilra, line 19: This function appears to allocate three copies of the string.  Can this be reduced to one allocation?

```
pub fn new(value: &str) -> Self {
  let mut bytes = Vec::with_capacity(value.len() + 1);
  bytes.extend_from_slice(value);
  // last element is implicitly zero
  Self { bytes }
}
```

wrappers/rust/oneds-telemetry/src/types.rs
```diff
@@ -0,0 +1,182 @@
...
+pub struct TelemetryItem {
+    name: StringProperty,
+    data: TelemetryPropertyBag,
+    time: Option<DateTime<Utc>>,
+}
+
+impl TelemetryItem {
+    pub fn new(name: &'static str) -> TelemetryItem {
+        TelemetryItem {
+            name: StringProperty::new(name),
```
sivadeilra, line 62: If you use the suggestion of using `Cow<'static, [u8]>` for the type of the field within `StringProperty`, and you require that the caller of this function provide a `name` value that is already NUL terminated, then this function can completely avoid allocating and copying the string.

wrappers/rust/oneds-telemetry/src/types.rs
```diff
@@ -0,0 +1,182 @@
...
+        properties.push(evt_prop {
+            name: NAME_PROPERTY.as_ptr(),
+            type_: evt_prop_t_TYPE_STRING,
+            value: evt_prop_v {
+                as_string: self.name.as_ptr(),
+            },
+            piiKind: 0,
+        });
+
+        if self.time != Option::None {
```
sivadeilra, line 92: Use this instead:

```
if let Some(time) = self.time {
  ...
  let timestamp_i64 = time.timestamp();
}
```

This unifies the "is it Some?" check with getting the value inside the Some.

wrappers/rust/cpp-client-telemetry-sys/src/lib.rs
```diff
@@ -0,0 +1,90 @@
...
+        handle: *handle,
+        data: std::ptr::null_mut(),
+        result: 0,
+        size: 0,
+    });
+
+    evt_api_call_wrapper(context).0
+}
+
+pub fn evt_log(handle: &evt_handle_t, data: &mut [evt_prop]) -> evt_status_t {
```
sivadeilra, line 77: This function must have an unsafe signature, i.e. `pub unsafe fn evt_log`, because the values inside `evt_prop` use raw pointers and the implementation of this function dereferences those raw pointers.  As written now, this function could be used to violate type safety.

A better solution could ensure that `evt_prop` was using type-safe types.  You could have a type which represents an owned pointer to a C-style string, that has the same representation as the underlying `evt_*` structure.  That would greatly reduce your exposure to unsafe code semantics.

If you can point out where the C headers are that define those structures, I can help write the safe Rust projections for these.

wrappers/rust/cpp-client-telemetry-sys/src/lib.rs
```diff
@@ -0,0 +1,90 @@
+#![allow(non_upper_case_globals)]
+#![allow(non_camel_case_types)]
+#![allow(non_snake_case)]
+include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
+
+use std::ffi::*;
+
+fn evt_api_call_wrapper(evt_context: Box<evt_context_t>) -> (evt_status_t, Box<evt_context_t>) {
```
sivadeilra, line 8: Why does this take ownership of the `Box<evt_context_t>` and then return it?  My naive reading of this would be that `evt_context: &evt_context_t` would be the right type.

wrappers/rust/cpp-client-telemetry-sys/src/lib.rs
```diff
@@ -0,0 +1,90 @@
...
+        data: std::ptr::null_mut(),
+        result: 0,
+        size: 0,
+    });
+
+    let (result, _) = evt_api_call_wrapper(context);
+    result
+}
+
+pub fn evt_upload(handle: &evt_handle_t) -> evt_status_t {
```
sivadeilra, line 53: Because you are directly exposing `evt_handle_t` outside your crate, and this type has unsafe pointers in it, these functions are actually unsafe to call.  They allow a safe caller to violate type safety.

You can fix this in one of several ways:
1. Don't expose these functions outside this crate.
2. Don't expose `evt_context_t`, and instead define an `EvtContext` wrapper type which contains `evt_context_t` but does not expose it outside of the crate.
3. Mark these functions `unsafe`.

wrappers/rust/oneds-telemetry/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+name = "oneds-telemetry"
+version = "0.1.0"
+edition = "2021"
+publish = false
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+chrono = { version = "0.4.30" }
+log = { version = "0.4.20" }
+once_cell = "1.12.0"
```
sivadeilra, line 11: Is there a reason to use `once_cell`, now that `Once` and `OnceLock` are part of the Rust standard library?

wrappers/rust/oneds-telemetry/src/lib.rs
```diff
@@ -0,0 +1,161 @@
...
+    evt_log,
+    evt_flush,
+    evt_upload,
+    evt_prop
+};
+
+pub mod appender;
+pub mod types;
+
+#[derive(Clone)]
```
sivadeilra, line 21: Is `log_handle` a pointer or a handle that was received from the underlying C API?  If so, then I don't think this Clone call is correct or safe.

wrappers/rust/oneds-telemetry/src/types.rs
```diff
@@ -0,0 +1,182 @@
...
+pub static TIME_PROPERTY: Lazy<StringProperty> = Lazy::new(|| StringProperty::new("time"));
+pub static EMPTY_STRING: Lazy<StringProperty> = Lazy::new(|| StringProperty::new(""));
+
+#[derive(Clone, Debug)]
+pub enum TelemetryProperty {
+    String(StringProperty),
+    Int(i64),
+}
+
+pub type TelemetryPropertyBag = HashMap<&'static str, TelemetryProperty>;
```
sivadeilra, line 50: It's a matter of taste, but I generally prefer to simply see `HashMap<&str, TelemetryProperty>` instead of a type alias.

wrappers/rust/oneds-telemetry/src/types.rs
```diff
@@ -0,0 +1,182 @@
...
+    data: TelemetryPropertyBag,
+    time: Option<DateTime<Utc>>,
+}
+
+impl TelemetryItem {
+    pub fn new(name: &'static str) -> TelemetryItem {
+        TelemetryItem {
+            name: StringProperty::new(name),
+            data: TelemetryPropertyBag::new(),
+            time: Option::None,
```
sivadeilra, line 64: Use `None` instead of `Option::None`.

lalitb:
@CDFriend This looks to be a good addition. when you have time, can you please make it ready for review :)

msft-tsharp:
> @CDFriend This looks to be a good addition. when you have time, can you please make it ready for review :)

@latlib - I can work on resolving the feedback.

lalitb:
> > @CDFriend This looks to be a good addition. when you have time, can you please make it ready for review :)
> 
> @lalitb - I can work on resolving the feedback.

@msft-tsharp Sorry for late reply. Do you think you can still work on resolving it ?

PR #1230, opened 2024-01-05
Fix an Apple break in Office around complex c'tors running at binary load time.

mkoscumb:
Fix is to move the complex c'tors into a Test Facet, so that they're initialized and destroyed only when the Test is run.

### DIFF
--- a/tests/unittests/MemoryStorageTests.cpp
+++ b/tests/unittests/MemoryStorageTests.cpp
@@ -60,21 +60,6 @@ class TestObserver : public IOfflineStorageObserver
     }
 };
 
-/// <summary>
-/// No-op NULL object pattern log manager
-/// </summary>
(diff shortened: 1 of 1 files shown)


lalitb (approved):
LGTM. Thanks for the fix.

tests/unittests/MemoryStorageTests.cpp
```diff
@@ -108,15 +93,40 @@ size_t addEvents(MemoryStorage& storage)
...
+
+    /// <summary>
+    /// Default test configuration
+    /// </summary>
+    std::unique_ptr<RuntimeConfig_Default>       testConfig;
+
+    /// <summary>
+    /// Storage observer
+    /// </summary>
+    TestObserver                testObserver;
```
absaroj, line 112: do we need TearDown for this one?

PR #1232, opened 2024-01-11
Cleaning leftover test files after test completion

absaroj:
Two test files were left out without cleanup after mbu test for ariasdk. Adding change to clean them up.

**Fix:**

_**test.ses:**_

reason for this being left out after mbu test is due to TEST method is being used insteasd of TEST_F which is with fixture responsible for calling the setup and tear down functions of the test class.
more info at: [link (Opens in new window or tab)](https://google.github.io/googletest/primer.html#same-data-multiple-tests)

**_sesfile.ses:_**
this is being left out since the cleanup funcitons are not being called after the setup.


mkoscumb (approved):
LGTM

PR #1233, opened 2024-01-24
Fix a few variable shadowing warnings in public headers.

mkoscumb:
I'm happy to change the names or suppress these warnings as well.


ThomsonTan (approved):
Thanks for the fix.

PR #1238, opened 2024-01-26
Fix deprecated CURL API usage.

utoni:
```
lib/./http/HttpClient_Curl.hpp: In member function ‘long int Microsoft::Applications::Events::CurlHttpOperation::Send()’:
lib/./http/HttpClient_Curl.hpp:211:39: error: ‘CURLINFO_LASTSOCKET’ is deprecated: since 7.45.0. Use CURLINFO_ACTIVESOCKET [-Werror=deprecated-declarations]
  211 |         res = curl_easy_getinfo(curl, CURLINFO_LASTSOCKET, &sockextr);
```

### DIFF
--- a/lib/http/HttpClient_Curl.hpp
+++ b/lib/http/HttpClient_Curl.hpp
@@ -208,7 +208,13 @@ public:
          * Note that this API takes a pointer to a 'long' while we use
          * curl_socket_t for sockets otherwise.
          */
+
+#if LIBCURL_VERSION_NUM >= 0x072D00 // Version 7.45.00
+        res = curl_easy_getinfo(curl, CURLINFO_ACTIVESOCKET, &sockextr);
+#else
         res = curl_easy_getinfo(curl, CURLINFO_LASTSOCKET, &sockextr);
+#endif
+
         if(CURLE_OK != res)
         {
             DispatchEvent(OnConnectFailed);     // couldn't connect - stage 2


lib/http/HttpClient_Curl.hpp
```diff
@@ -208,7 +208,7 @@ class CurlHttpOperation {
          * Note that this API takes a pointer to a 'long' while we use
          * curl_socket_t for sockets otherwise.
          */
-        res = curl_easy_getinfo(curl, CURLINFO_LASTSOCKET, &sockextr);
+        res = curl_easy_getinfo(curl, CURLINFO_ACTIVESOCKET, &sockextr);
```
lalitb, line 211: CURLINFO_ACTIVESOCKET was added in  7.45.0. Just to be on safer side, should we check for the libcurl version (probably using [LIBCURL_VERSION_NUM](https://github.com/curl/curl/blob/aee4ebe59161d0a5281743f96e7738ad97fe1cd4/include/curl/curlver.h#L62)), and use right one accordingly.

lalitb (approved):
Approve with suggestion :)

lib/http/HttpClient_Curl.hpp
utoni: Should I change that in the current PR or a follow-up? :)

ThomsonTan: Add a TODO comment here if it will be included in the following up PR?

utoni:
Fixed for the sake of simplicity.

ThomsonTan (approved):

PR #1240, opened 2024-02-02
Fixing release build unused variable error by adding conditional complation to the code run only for debug builds

absaroj:
Variable is only used for debug purposes and causes error on the local build due to being unused.


lalitb (approved):
Thanks for the fix.

PR #1241, opened 2024-02-02
adding comment about using the typealised enums in swift from objc 

absaroj:
Objc enums accessed in swift have their names removed from the enum values due to convention. Adding comment about it to let the users know.


mkoscumb (approved):
LGTM

PR #1248, opened 2024-02-26
Guard registerReceiver with RECEIVER_NOT_EXPORTED for Android 14 (SDK 34)

mgowtham:
## Problem

`HttpClient` registers a `BroadcastReceiver` for `ACTION_BATTERY_CHANGED` using the 2-arg `registerReceiver(receiver, filter)`. The maesdk targets **SDK 34** (`lib/android_build/tools.gradle`), and on **Android 14 (API 34)** registering a non-system receiver without specifying an export flag throws `SecurityException` — crashing `HttpClient` construction.

## Fix

Pass an explicit export flag, **version-guarded** so it stays compatible with the SDK's full supported range (`minSdk 23`):

```java
Intent status;
if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {  // API 33
  status = context.registerReceiver(m_power_receiver, filter, Context.RECEIVER_NOT_EXPORTED);
} else {
  status = context.registerReceiver(m_power_receiver, filter);
}
```

- **`RECEIVER_NOT_EXPORTED`** (rather than `RECEIVER_EXPORTED`): `ACTION_BATTERY_CHANGED` is a protected system broadcast delivered only by the OS. The receiver never needs other apps to broadcast to it, so not-exported is the secure, correct choice — and it satisfies the API-34 requirement.
- **Version guard**: the 3-arg `registerReceiver` overload is **API 26+** and the `RECEIVER_*` constants are **API 33+**. Using them unconditionally would throw `NoSuchMethodError` on API 23–25 (the SDK's `minSdk` is 23) and trip a `NewApi` lint error. The guard keeps the original 2-arg call on older devices.
- No new dependency (`android.content.Context` / `android.os.Build` are already imported); lint-clean.

## Validation

- Net change is a single receiver-registration block; verified no stray `RECEIVER_EXPORTED` references remain.
- Android CI (`build-android.yml`) builds the maesdk.

Branch brought up to date with `main`. Original change by @mgowtham (credited as co-author).


mkoscumb (approved):
LGTM

lalitb (approved):

lalitb:
@mgowtham - You may want to merge this PR as it is approved?

ThomsonTan:
@mgowtham any plan to merge this PR?

bmehta001:
Took this over to bring it up to date with `main` and adjust the fix for the SDK's full supported API range, @mgowtham — thanks for flagging the Android 14 issue.

Two changes from the original `RECEIVER_EXPORTED` edit:

1. **Version-guarded the call.** The 3-arg `registerReceiver` overload (API 26) and the `RECEIVER_*` constants (API 33) don't exist on the SDK's `minSdk 23` floor, so the unconditional form would throw `NoSuchMethodError` on API 23–25 (and trip a `NewApi` lint error). It now falls back to the original 2-arg call below API 33.
2. **`RECEIVER_NOT_EXPORTED` instead of `RECEIVER_EXPORTED`.** `ACTION_BATTERY_CHANGED` is a protected system broadcast, so the receiver never needs to be exported to other apps — not-exported is the more secure choice and still satisfies the API-34 requirement.

Your original commit is preserved in the history and you're credited as co-author. Updated the title/description to match.

bmehta001:
Closing this — after taking it over and **testing it on an emulator**, the premise turns out not to hold: the SDK's battery receiver doesn't actually crash on Android 14, so there's nothing to fix here.

**What I tested:** an API 36 emulator (which enforces the same Android-14 export-flag rule), with a test app targeting SDK 34. Registering a receiver for `ACTION_BATTERY_CHANGED` with the existing 2-arg `registerReceiver(receiver, filter)` **did not throw** `SecurityException`.

**Why:** `ACTION_BATTERY_CHANGED` is a protected **system broadcast**, and Android 14's "Runtime-registered receivers must specify export behavior" change has an explicit carve-out:

> **Exception for receivers that receive only system broadcasts** — If your app is registering a receiver only for system broadcasts … then it **shouldn't specify a flag** when registering the receiver.
> — https://developer.android.com/about/versions/14/behavior-changes-14

The SDK's `HttpClient` registers a receiver for **only** `ACTION_BATTERY_CHANGED`, so it's exempt. The current 2-arg call is the documented-correct form, and the docs say such receivers *shouldn't* specify a flag — so adding `RECEIVER_EXPORTED`/`RECEIVER_NOT_EXPORTED` is unnecessary.

Thanks @mgowtham for raising it — it was a very reasonable concern given the Android 14 messaging, and it was worth verifying. No code change needed; the existing receiver registration is correct. Closing as not-a-bug.

PR #1249, opened 2024-02-26
Adding new PiiKind values for Scrub Last Hextet and for dropping value.

aef123:
Adding 2 new PII Kind values for scrubbing last hextet and for dropping values.


lib/include/public/Enums.hpp
```diff
@@ -129,6 +129,10 @@ namespace MAT_NS_BEGIN
         PiiKind_Fqdn = 12,
         /// <summary>A legacy IPV4 Internet address.</summary>
         PiiKind_IPv4AddressLegacy = 13,
+        /// <summary>A legacy IPV6 Internet address.</summary>
+        IPv6ScrubLastHextets = 14,
+        /// <summary>Drop the value.</summary>
+        DropValue = 15,
```
lalitb, line 135: should these both be defined with `PiiKind` prefix to maintain consistency?

lib/bond/generated/CsProtocol.json
```diff
@@ -0,0 +1 @@
...:null,"fieldType":{"type":"vector","element":{"type":"vector","element":"string"}},"fieldName":"stringArray","fieldAttributes":[],"fieldOrdinal":10},{"fieldModifier":"Optional","fieldDefault":null,"fieldType":{"type":"vector","element":{"type":"vector","element":"int64"}},"fieldName":"longArray","fieldAttributes":[],"fieldOrdinal":11},{"fieldModifier":"Optional","fieldDefault":null,"fieldType":{"type":"vector","element":{"type":"vector","element":"double"}},"fieldName":"doubleArray","fieldAttributes":[],"fieldOrdinal":12},{"fieldModifier":"Optional","fieldDefault":null,"fieldType":{"type":"vector","element":{"type":"vector","element":{"type":"vector","element":"uint8"}}},"fieldName":"guidArray","fieldAttributes":[],"fieldOrdinal":13}],"declParams":[],"declNamespaces":[{"name":["CsProtocol"]}],"declName":"Value","declAttributes":[]},"type":"user"}},"fieldName":"properties","fieldAttributes":[],"fieldOrdinal":1}],"declParams":[],"declNamespaces":[{"name":["CsProtocol"]}],"declName":"Data","declAttributes":[]},"type":"user"}},"fieldName":"data","fieldAttributes":[],"fieldOrdinal":70}],"declParams":[],"declNamespaces":[{"name":["CsProtocol"]}],"declName":"Record","declAttributes":[]}]}
```
lalitb, line 1: Is this json file intended?

lib/bond/generated/CsProtocol_writers.hpp
```diff
@@ -233,15 +226,15 @@ void Serialize(TWriter& writer, ::CsProtocol::Device const& value, bool isBase)
     } else {
         writer.WriteFieldOmitted(BT_STRING, 9, nullptr);
     }
-#ifdef HAVE_CS4
+
```
lalitb, line 229: Is it intended to make CS4 as default serialization?

lib/bond/generated/CsProtocol_readers.hpp
```diff
@@ -315,14 +310,14 @@ bool Deserialize(TReader& reader, ::CsProtocol::Device& value, bool isBase)
                 }
                 break;
             }
-#ifdef HAVE_CS4
+
```
ThomsonTan, line 313: Why is the check of HAVE_CS4 removed?

lib/include/public/Enums.hpp
aef123: Yes. Fixed.

lib/bond/generated/CsProtocol_writers.hpp
aef123: No. I ran the bond compiler to regenerate the classes to add the new fields, and it overwrote this file. I didn't realize we were manually modifying a file that is supposed to only be generated by the bond compiler. this will need to be fixed.

lib/bond/generated/CsProtocol_readers.hpp
aef123: Because I re-ran the bond compiler to generate new classes with the updated fields, and I did not realize we had manually modified this generated file. I'll need to udpate this.

aef123:
I'm going to hand this off to the team that needs the SDK to make the changes.

lalitb:
Created issue #1258 for tracking.

PR #1259, opened 2024-04-01
Documenting LogManager Setup and Best Practices

lalitb:
Adding bit more guidance in our docs on setting up and managing loggers, especially since we've got those two paths: `LOGMANAGER_INSTANCE` for the straightforward cases and `LogManagerProvider` for when things get a bit more complex.

What's added?

- A breakdown of when and how to use `LOGMANAGER_INSTANCE`.
- A guide on leveraging `LogManagerProvider` for more advanced logging setups.
- Some advice on keeping things smooth and safe with loggers in multithreaded environments.
- Tips on choosing the right approach based on what you're building.


lalitb:
@maxgolov @mkoscumb -  whenever you have time to review this PR. Thanks.

docs/LogManager_Creation_and_Lifecycle_Management.md
```diff
@@ -0,0 +1,79 @@
...
+LogManagerProvider::Release(logManager); // Releases the LogManager instance
+```
+
+### LogConfiguration Lifecycle
+
+When creating a ILogManager instance, the provided configuration is used throughout the instance's lifecycle. The LogManagerImpl holds a reference to this configuration, meaning the configuration must remain valid for as long as the ILogManager instance is in use.
+
+# Choosing the Right Approach
+
+1. For simplistic, non-threaded applications where logging requirements are uniform and straightforward, employing the `LOGMANAGER_INSTANCE` pattern `MAY` be an optimal choice.
```
mkoscumb, line 76: I'd rather suggest partners manage the lifetime of their LogManagers explicitly, rather than rely on language level global object behaviors.

Perhaps the order should be swapped, and discourage usage of LOGMANAGER_INSTANCE.

mkoscumb (approved):

docs/LogManager_Creation_and_Lifecycle_Management.md
lalitb: Agree - have swapped the order, and explicitly mentioned it as NOT recommended.

PR #1264, opened 2024-04-15
Remove _CRT_SECURE_NO_WARNINGS macro usage

mkoscumb:
It seems like all the warnings that _CRT_SECURE_NO_WARNINGS disabled have been resolved. Removing usage of the macro, as its no longer necessary.

There was one warning fired by usage of gmtime vs gmtime_s. However, given that the result of gmtime was never used I simply removed it.


Solutions/win32-lib/win32-lib.vcxproj
```diff
@@ -319,7 +319,7 @@
       </PrecompiledHeader>
       <WarningLevel>Level4</WarningLevel>
       <Optimization>Disabled</Optimization>
-      <PreprocessorDefinitions>ZLIB_WINAPI;WIN32;_CRT_SECURE_NO_WARNINGS;WIN32;NOMINMAX;_DEBUG;_WINDOWS;_USRDLL;WINVER=_WIN32_WINNT_WIN7;NOMINMAX;%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
lalitb, line 322: Is removing `NOMINMAX` intended here, and few other places ?

lalitb (approved):

Solutions/win32-lib/win32-lib.vcxproj
mkoscumb: No, I have no idea why VS did that.

mkoscumb: Ah, because some are duplicated.

PR #1266, opened 2024-04-22
Contributing back patch to 1DS

lakshmir124:


lib/include/public/IECSClient.hpp
```diff
@@ -51,6 +52,10 @@ namespace Microsoft {
 
                     // [optional] enabled ECS telemetry
                     bool enableECSClientTelemetry = false;
+
+                    // [optional] Mandatory agents list. If not present in response, the payload
+                    // is determined as bad
+                    std::unordered_set<std::string> mandatoryAgents;
```
lalitb, line 58: Did you miss adding any file, as `mandatoryAgents` is not used anywhere?

lalitb:
Thanks @lakshmir124, can you please add more details to the PR description?

lib/include/public/IECSClient.hpp
lakshmir124: It will be used in the EXPCommonClient.cpp file in this PR - [Pull 230 - Contributing back patches to 1DS modules](https://github.com/microsoft/cpp_client_telemetry_modules/pull/230/files). I think this PR will have to be merged first before we merged the private one.

lib/http/HttpClient_WinInet.cpp
```diff
@@ -22,6 +23,8 @@
 
 namespace MAT_NS_BEGIN {
 
+const std::string kProxyRegKeyPath = "Software\\Microsoft\\Windows\\CurrentVersion\\Internet Settings";
```
mkoscumb, line 26: Don't create namespace scoped STL types as these cause heap allocaitons which are never released.
`const char* kProxyRegKeyPath`

lib/http/HttpClient_WinInet.cpp
```diff
@@ -136,6 +139,105 @@ class WinInetRequestWrapper
...
+        }
+        return data;
+    }
+
+    std::string GetStringRegKey(HKEY hKey, const std::string& subKey, const std::string& value) {
+        DWORD stringSize;
+        LONG getSizeResult = RegGetValueA(hKey, subKey.c_str(), value.c_str(), RRF_RT_REG_SZ, NULL, 
+            NULL, &stringSize);
+        if (getSizeResult != ERROR_SUCCESS) {
+            return "";
```
mkoscumb, line 158: use the default string constructor, rather than the c-string constructor:
`std::string{}`

lib/http/HttpClient_WinInet.cpp
```diff
@@ -136,6 +139,105 @@ class WinInetRequestWrapper
...
+        }
+
+        std::string data;
+        data.resize(stringSize);
+        LONG getStringResult = RegGetValueA(hKey, subKey.c_str(), value.c_str(), RRF_RT_REG_SZ, NULL, 
+            &data[0], &stringSize);
+        if (getStringResult != ERROR_SUCCESS) {
+            return "";
```
mkoscumb, line 166: Same.

lib/http/HttpClient_WinInet.cpp
```diff
@@ -136,6 +139,105 @@ class WinInetRequestWrapper
         return true;
     }
 
+    DWORD GetDWORDRegKey(HKEY hKey, const std::string& subKey, const std::string& value) {
+        DWORD data;
+        DWORD dataSize = sizeof(data);
+        LONG res = RegGetValueA(hKey, subKey.c_str(), value.c_str(), RRF_RT_REG_DWORD, NULL, &data, 
+            &dataSize);
+        if (res != ERROR_SUCCESS) {
+            return 0;
```
mkoscumb, line 148: How does a caller distinguish between a zero in the Registry value and an error?

lib/http/HttpClient_WinInet.cpp
mkoscumb, line 158: How does a caller distinguish between an empty string in the registry and an error?

lib/http/HttpClient_WinInet.cpp
```diff
@@ -136,6 +139,105 @@ class WinInetRequestWrapper
         return true;
     }
 
+    DWORD GetDWORDRegKey(HKEY hKey, const std::string& subKey, const std::string& value) {
```
mkoscumb, line 142: If these are always known at build time, don't do unnecessary heap allocations, use const char* rather than std::string

mkoscumb: std::string_view would be ideal, but this repo is still not on C++17 for some reason.

lib/http/HttpClient_WinInet.cpp
```diff
@@ -136,6 +139,105 @@ class WinInetRequestWrapper
...
+        }
+
+        // Remove the null terminator from Win32 API. The std::string will already cover this.
+        data.resize(stringSize - 1);
+
+        return data;
+    }
+
+    bool ProxyAuthRequired() {
+        return GetDWORDRegKey(HKEY_CURRENT_USER, kProxyRegKeyPath, "ProxyEnable");
```
mkoscumb, line 176: Don't rely on implicit conversions.

`GetDWORDRegKey(...) != 0`

lib/http/HttpClient_WinInet.cpp
```diff
@@ -136,6 +139,105 @@ class WinInetRequestWrapper
...
+                                                proxyServer.end());
+        }
+
+        // get proxy credentials from credential manager
+        PCREDENTIALW cred;
+        if (!::CredReadW(formattedProxyServer.c_str(), CRED_TYPE_GENERIC, 0, &cred)) {
+            return;
+        }
+        wchar_t* proxyUser = cred->UserName;
+        wchar_t* proxyPass = (wchar_t*)cred->CredentialBlob;
```
mkoscumb, line 221: No c-style casts in modern code, use static_cast<wchar_t*> or reinterpret_cast<wchar_t*>

mkoscumb (changes requested):

lib/http/HttpClient_WinInet.cpp
```diff
@@ -136,6 +139,105 @@ class WinInetRequestWrapper
...
+        }
+
+        if (proxyPass) {
+            ::InternetSetOptionW(m_hWinInetRequest,
+                                 INTERNET_OPTION_PROXY_PASSWORD,
+                                 static_cast<void*>(proxyPass),
+                                 static_cast<DWORD>(wcslen(proxyPass) + 1));
+        }
+
+        ::CredFree(cred);
```
mkoscumb, line 238: You should either wrap PCREDENTIALW in a helper type whose destructor calls ::CredFree, or refactor 216-238 into a helper function which is marked `noexcept`. That way if at any point (including if someone adds new code which can raise exceptions) an exception is raised there's no chance cred is leaked.

mkoscumb: The former is preferred as it's more resilient.
```c++
struct CredHolder final
{
    PCREDENTIALW cred;
    ~CredHolder() noexcept
    {
        ::CredFree(cred);
    }
};
```

PR #1275, opened 2024-05-13
Add Java to CodeQL scan list

ThomsonTan:
Switched to Windows for building C#. And the CI of Android also runs on Windows.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -99,9 +99,9 @@ jobs:
         git -c protocol.version=2 submodule update --init --force --depth=1
 
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -63,5 +72,70 @@ jobs:
     #   make bootstrap
     #   make release
 
+    - name: Perform CodeQL Analysis
+      uses: github/codeql-action/analyze@v2
+
+  analyze-java:
```
reyang, line 78: The PR title says "csharp", which one is true?

ThomsonTan: My typo. I updated the title but seems it is not saved. The current PR is for Java.

Got some issue in C# build, so reverted it for now.

lalitb (approved):
Thanks.

PR #1279, opened 2024-05-25
Fix build scripts and add visionOS support

tkukielk:
Multiple fixes:
- building on Apple Silicon Macs - remove the assumptions that the build machine is always x86_64
- fix badly broken build.sh for Mac, especially with getopts for options & params
- extend build-ios.sh to allow building for iOS-derived platforms like visionOS

The expected iOS build invocation is:
build-ios.sh [clean] [release|debug] ${ARCH} ${PLATFORM} where
ARCH = arm64|arm64e|x86_64
PLATFORM = iphoneos|iphonesimulator|xros|xrsimulator


lalitb (approved):
LGTM. Thanks for improving the build script too.

build-ios.sh
```diff
@@ -1,69 +1,99 @@
...
 
-# Set Platform: device or simulator
-if [ "$2" == "device" ] || [ "$3" == "device" ] || [ "$4" == "device" ]; then
-IOS_PLAT="iphoneos"
-else
+# the last param is expected to specify the platform name: iphoneos|iphonesimulator|xros|xrsimulator
+# so if it is non-empty and it is not "device", we take it as a valid platform name
+# otherwise we fall back to old iOS logic which only supported iphoneos|iphonesimulator
 IOS_PLAT="iphonesimulator"
+if [ -n "$1" ] && [ "$1" != "device" ]; then
```
swasti29, line 44: How can the first parameter denotes the architecture, build type and platform? You should have check for secound and third?

tkukielk: 'shift' command removes the consumed parameter, shifting $2 -> $1. so these params may be optional but are positional and the required order is: build-ios.sh [clean] [release|debug] ${ARCH} ${PLATFORM}

swasti29 (approved):
Just make sure to test for all architecture and platform: device, simulator

Issue #1269, opened 2024-04-30
MacOS CI failing


lalitb:
**Error:**

https://github.com/microsoft/cpp_client_telemetry/actions/runs/8744717219/job/23998207904?pr=1265

```
make[2]: *** No rule to make target `sqlite3-NOTFOUND', needed by `tests/functests/FuncTests'.  Stop.
make[1]: *** [tests/functests/CMakeFiles/FuncTests.dir/all] Error 2
make: *** [all] Error 2
./build-tests.sh: line 9: ./tests/functests/FuncTests: No such file or directory
```
The MacOS virtual environment has changed.
***Earlier:*** macOS 12.7.4, 21H1123, cmake 3.29.2,  Apple clang version 14.0.0
***Now*** macOS 14.4.1 23E224, cmake 3.29.2, Apple clang version 15.0.0

PR #1280, opened 2024-05-29
MacOS CI failure fix

lalitb:
Fixes: #1269 

It [seems ](https://github.com/actions/runner-images/issues/9272)that macOS-13 image has `runner:admin` permissions for `/usr/local/`. But macOS-14 has `root:wheel`. So modified `build.sh` so that the build-tools script for Apple is executed as `sudo`, just like for Linux.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -101,11 +101,12 @@ FILE=.buildtools
 OS_NAME=`uname -a`
 if [ ! -f $FILE ]; then
   case "$OS_NAME" in
-    *Darwin*) tools/setup-buildtools-apple.sh $MAC_ARCH ;;
-    *Linux*)  [[ -z "$NOROOT" ]] && sudo tools/setup-buildtools.sh || echo "No root: skipping build tools installation." ;;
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -101,11 +101,12 @@ FILE=.buildtools
 OS_NAME=`uname -a`
 if [ ! -f $FILE ]; then
   case "$OS_NAME" in
-    *Darwin*) tools/setup-buildtools-apple.sh $MAC_ARCH ;;
-    *Linux*)  [[ -z "$NOROOT" ]] && sudo tools/setup-buildtools.sh || echo "No root: skipping build tools installation." ;;
```
ThomsonTan, line 105: Should `sudo` be added or removed here?

lalitb: There is no change for Linux. Only sudo is added for Mac.

lalitb: No change for Linux. Only adding sudo for Mac.

lalitb:
The CI build is successful for Mac and Linux. Should be good to review the PR.

ThomsonTan (approved):

PR #1286, opened 2024-07-11
Introduce memset_s in sqlite3

ShrutiJaiswal1494:
Introduce memset_s in sqlite3


sqlite/sqlite3_debug.c
```diff
@@ -17,6 +17,41 @@
 ** language. The code for the "sqlite3" command-line shell is also in a
 ** separate file. This file contains only code for the core SQLite library.
 */
+
+/*
+** The memset_s implementation is added as a secure version of the traditional
+** memset function. It includes safety checks to prevent buffer overflows.
+** Please do not remove this method as it is required for the CodeQL sanity checks.
```
lalitb, line 24: nit - please remove this comment regarding CodeQL, as this is specific to the CodeQL configuration internally used, and doesn't go well in open source code.

sqlite/sqlite3_debug.c
```diff
@@ -17,6 +17,41 @@
...
+    if (n > smax) {
+        // Set memory up to the buffer size and return an error
+        memset(s, c, smax);
+        return EINVAL;
+    }
+    // Perform the memory set operation for the requested size
+    memset(s, c, n);
+    return 0;
+}
+
```
lalitb, line 54: please move this function in separate header file, and include that header file here. Also I believe we need to update `sqlite3_retail` with similar changes, so the header file can be used across both.

sqlite/sqlite3_debug.c
```diff
@@ -49629,7 +49664,8 @@ static SQLITE_NOINLINE PgHdr *pcacheFetchFinishWithInit(
   assert( pPage!=0 );
   pPgHdr = (PgHdr*)pPage->pExtra;
   assert( pPgHdr->pPage==0 );
-  memset(&pPgHdr->pDirty, 0, sizeof(PgHdr) - offsetof(PgHdr,pDirty));
+  memset_s(&pPgHdr->pDirty, sizeof(PgHdr) - offsetof(PgHdr, pDirty), 0,
+    sizeof(PgHdr) - offsetof(PgHdr, pDirty));
```
lalitb, line 49668: Should we invoke this conditionally under a macro? Although these bound validations enhance safety and prevent buffer overflows, they also introduce a slight performance overhead, which may not be acceptable for applications with strict performance requirements.  This is particularly relevant in scenarios where sqlite is already ensuring that calling `memset` does not result in an overflow. Something like:

```cpp
#if USE_ONEDS_SECURE_MEM_FUNCTIONS 
    memset_s(&pPgHdr->pDirty, sizeof(PgHdr) - offsetof(PgHdr, pDirty), 0,
             sizeof(PgHdr) - offsetof(PgHdr, pDirty));
#else
    memset(&pPgHdr->pDirty, 0, sizeof(PgHdr) - offsetof(PgHdr, pDirty));
#endif
```

lalitb:
@ShrutiJaiswal1494 - Can you address the open comments? It was agreed in today's community meeting that the changes can proceed.

lalitb (changes requested):
As mentioned in the comments.

sqlite/sqlite3_debug.c
ThomsonTan: Probably good to guard all the new code within the feature flag like `USE_ONEDS_SEURE_MEM_FUNCTIONS`, or there will be unused function of memset_s left.

lalitb (approved):
LGTM. Thanks for the PR.

ThomsonTan (approved):

lalitb:
Doing force merge, as the mac CI is failing for older virtual images

Issue #1287, opened 2024-07-12
CI failing for the use of macOS11 runner image


lalitb:
failure - https://github.com/microsoft/cpp_client_telemetry/actions/runs/9889178216/job/27314645563?pr=1286

macOS11 is not longer in the list of available runner images - https://github.com/actions/runner-images
The fix should be to upgrade to macOS12 runner image.

ThomsonTan:
@lalitb thanks for looking this one.

ThomsonTan:
@absaroj this is the issue I mentioned in today's community meeting. Thanks for looking into it.

bmehta001:
We are now using macos-14, so this can be closed.

PR #1289, opened 2024-07-12
Fix for build errors when building with gcc 13

### LINKED ISSUE #1288: cpp_client_telemetry fails to build with gcc13
Ubuntu Linux 24.04 with gcc 13

**Steps to reproduce.**
./build.sh

There are files which reference types and don't include necessary header files. It is likely that these header files were being inherited through other included headers and now they must be included explicitly in the project.

The first few encountered:
Enums.hpp: `uint64_t` requires `<cstdint>`
EventFilterCollection.cpp: `std::invalid_argument` requires `<stdexcept>`

The fix is to just go through building the repo repeatedly and adding missing include directives.

DSMitten:
A few files use types defined in headers that are missing, and this is revealed through build errors when building with gcc. Something about libstdc++ may have refactored header files such that headers that previously were inherited now must be explicitly included.

fixes #1288

### DIFF
--- a/tests/unittests/BondSplicerTests.cpp
+++ b/tests/unittests/BondSplicerTests.cpp
@@ -32,7 +32,7 @@ class ShadowBondSplicer : protected MAT::BondSplicer
     {
         FullDumpBinaryBlob output;
         static_cast<std::vector<uint8_t>&>(output) = MAT::BondSplicer::splice();
-        return std::move(output);
+        return output;
     }
 };
 
--- a/lib/api/CorrelationVector.cpp
+++ b/lib/api/CorrelationVector.cpp
@@ -7,6 +7,7 @@
 
 #include <vector>
 #include <random>
+#include <stdexcept>
 #include <limits>
 
 using std::string;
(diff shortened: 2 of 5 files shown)


tests/unittests/BondSplicerTests.cpp
```diff
@@ -32,7 +32,7 @@ class ShadowBondSplicer : protected MAT::BondSplicer
     {
         FullDumpBinaryBlob output;
         static_cast<std::vector<uint8_t>&>(output) = MAT::BondSplicer::splice();
-        return std::move(output);
```
DSMitten, line 35: The compiler flagged this as an error, saying that it is a redundent call to `std::move`

mkoscumb: Probably pessimizing move. Good fix.

mkoscumb (approved):
LGTM

lalitb (approved):

tests/unittests/BondSplicerTests.cpp
ThomsonTan: Seems return value optimization does apply here so explicit move is not required.

ThomsonTan:
Seems all iOS build are broken, are they relatied?

lalitb:
> Seems all iOS build are broken, are they relatied?

#1287

ThomsonTan (approved):

PR #1292, opened 2024-07-16
make _wrappedConfiguration into member variable

gietal-msft:
**PROBLEM:**
_wrappedConfiguration is a global variable which causes ODWLogConfiguration.getCopy() to return objects with a shared underlying ILogConfiguration instead of unique copies.
 
**SOLUTION:**
I believe _wrappedConfiguration was intended to be member variable of the ODWLogConfiguration, so patching it as such.


lalitb:
@absaroj Can you help review this?

bmehta001:
I opened #1539 as a follow-up to this fix. It makes the wrapped configuration instance-owned, includes ownership of the copied C++ configuration, and adds an XCTest for independent copies. Leaving this PR open for its author/maintainers to close or merge as appropriate.

bmehta001:
Update: the fix from #1539 passed its iOS XCTest matrix and is now included in #1529. #1539 is being closed as a duplicate branch; please track #1529 for the merged copy-isolation and ownership change.

PR #1293, opened 2024-07-18
Mark sqlite as lib folder in CodeQL scan

ShrutiJaiswal1494:
Add CodeQL.yml file to add sqlite as a library.

### DIFF
--- a/CodeQL.yml
+++ b/CodeQL.yml
@@ -1,0 +1,5 @@
+# See https://eng.ms/docs/cloud-ai-platform/devdiv/one-engineering-system-1es/1es-docs/codeql/troubleshooting/bugs/generated-library-code
+path_classifiers:
+  # Tagging the sqlite folder as 'library'.
+  library:
+    - "sqlite/**"


CodeQL.yml
```diff
@@ -0,0 +1,5 @@
+# See https://eng.ms/docs/cloud-ai-platform/devdiv/one-engineering-system-1es/1es-docs/codeql/troubleshooting/bugs/generated-library-code
+path_classifiers:
+  # Tagging the sqlite folder as 'library'.
+  library:
+    - "sqlite/**"
```
ThomsonTan, line 5: Will mark sqlite as 'library' exclude it from the scanning result?

ShrutiJaiswal1494: Yes, I believe so.

ThomsonTan: But why exclude sqlite from the result? as we forked it, we need make sure the fork is clean on CodeQL, is this right?

ShrutiJaiswal1494: The CodeQL support team provided documentation for addressing CodeQL issues in library code: [CodeQL Troubleshooting - Bugs in Generated Library Code](https://eng.ms/docs/cloud-ai-platform/devdiv/one-engineering-system-1es/1es-docs/codeql/troubleshooting/bugs/generated-library-code). I noticed that a couple of teams in MSO are using this configuration to mark the library files.

Alternatively, we could move the `memset` function to `memset_s`, as I proposed in my previous PR: [PR #1286 - Move to memset_s](https://github.com/microsoft/cpp_client_telemetry/pull/1286).

Please let me know which solution you think would be better.

lalitb:
I like the approach of configuring SQLite as a library for CodeQL, if that removes false positives. The usage of _memcpy_ and _memset_ within such a widely used database should generally be safe.  Adding any guarded functions to make CodeQL happy could introduce side effects that might be difficult to catch in normal tests. A more effective effort would be to upgrade the SQLite3 snapshot to the latest version.

p.s - 

An interesting usage spectrum for SQLite3 - https://www.sqlite.org/mostdeployed.html

```
SQLite is likely used more than all other database engines combined. Billions and billions of copies 
of SQLite exist in the wild.

it is seems likely that there are over one trillion (1e12) SQLite databases in active use.

```

ThomsonTan:
Based on our discussion, #1286 is preferred so close this one.

Issue #1296, opened 2024-08-22
ClangTidy issue in lib/system/Route.hpp


HIROSN:
**Describe your environment.** 
Issue found in Microsoft Edge and still exists in the current main.

**Steps to reproduce.**
Issue was reported by static analysis using ClangTidy.

**What is the expected behavior?**
No issue reported.

**What is the actual behavior?**
use-after-move reported.

**Additional context.**
https://microsoft.visualstudio.com/Edge/_workitems/edit/52434861
https://microsoft.visualstudio.com/Edge/_workitems/edit/38580805

HIROSN:
https://github.com/microsoft/cpp_client_telemetry/pull/1297

PR #1297, opened 2024-08-22
ClangTidy issue in lib/system/Route.hpp #1296

HIROSN:
In RouteSource::operator(), r-value in 'args' is used in a later loop iteration after it was forwarded in the first loop iteration.
Proposed fix is to remove usage of std::forward and allow l-value 'args' only.


lib/system/Route.hpp
```diff
@@ -102,12 +102,12 @@ namespace MAT_NS_BEGIN {
         void operator()(TRealArgs&& ... args) const
         {
             for (IRoutePassThrough<TArgs...>* target : m_passthroughs) {
-                if (!(*target)(std::forward<TRealArgs>(args) ...)) {
+                if (!(*target)(args ...)) {
```
ThomsonTan, line 105: Will `args` be coped for each call?

tests/unittests/RouteTests.cpp
```diff
@@ -60,25 +60,17 @@ class Canary : public std::string {
...
-    RouteSink<RouteTests, int>                                                             sink1{this, &RouteTests::handleSink1};
-    RoutePassThrough<RouteTests, int>                                                      passThrough1a{this, &RouteTests::handlePassThrough1a};
-    RoutePassThrough<RouteTests, int>                                                      passThrough1b{this, &RouteTests::handlePassThrough1b};
-    RouteSink<RouteTests, int, bool&, NonCopyableThing const&, Canary, std::vector<int>&&> sink5{this, &RouteTests::handleSink5x};
+    RouteSink<RouteTests>                                              sink0{this, &RouteTests::handleSink0};
+    RouteSink<RouteTests, int>                                         sink1{this, &RouteTests::handleSink1};
+    RoutePassThrough<RouteTests, int>                                  passThrough1a{this, &RouteTests::handlePassThrough1a};
+    RoutePassThrough<RouteTests, int>                                  passThrough1b{this, &RouteTests::handlePassThrough1b};
+    RouteSink<RouteTests, int, bool&, NonCopyableThing const&, Canary> sink4{this, &RouteTests::handleSink4};
```
ThomsonTan, line 67: >sink4 [](http://example.com/codeflow?start=71&length=5)

Is this change relating to the change in RouteSource::operator()?

PR #1298, opened 2024-09-05
Upgrade mockito test dependency to be compatible with new Java version

anod:
Make unit test compatible with Java 17


lalitb (approved):

lalitb:
force merging as macos build are failing for older virtual images.

PR #1301, opened 2024-09-10
Prevent `-Wcast-function-type-mismatch` diagnostics from clang

ZacharyHenkel:
-Wcast-function-type-mismatch is a new diagnostic issued by clang 19.  Add an additional cast to silence the diagnostic.  The underlying issue is that `GetProcAddress` unconditionally returns `FARPROC` instead of matching the real function signature.


lalitb (approved):

mkoscumb (approved):
LGTM

ThomsonTan (approved):

PR #1307, opened 2024-11-04
Fix arg callout in substring expansion (bash)

tomkaragounis:
Noticed that custom build flags were not being correctly assigned to the CUSTOM_CMAKE_CXX_FLAG variable in the build.sh script. Not sure how this got modified as I would still expect $1 to contain the full arg. Works with below fix.

Error examples:
./build.sh CUSTOM_BUILD_FLAGS="-DHAVE_MAT_EVT_TRACEID=1" debug
Current directory: /home/tomkar
BUILD_TYPE = Debug
Using current machine MAC_ARCH = x86_64
custom build flags = ""

./build.sh CUSTOM_BUILD_FLAGS -DHAVE_MAT_EVT_TRACEID=1 debug
Current directory: /home/tomkar
BUILD_TYPE = Debug
Using current machine MAC_ARCH = x86_64
custom build flags = "EID=1"

"EID=1" is 19 chars into -DHAVE_MAT_EVT_TRACEID=1, so offset is activated here, just for the next argument instead of the CUSTOM_BUILD_FLAGS


ThomsonTan:
As we expect `${1}` should match `$PARAM1`, could we try to understand why and how it is modified?

tomkaragounis:
> As we expect `${1}` should match `$PARAM1`, could we try to understand why and how it is modified?

I am actually not very well versed in bash, so I do not understand why the behavior is not working correctly. I have been testing it out locally and realized the problem when I try adding my own custom args

bmehta001:
Thanks for catching and reporting this, @tomkaragounis — the bug was real: the old block tested `$PARAM1/2/3` but extracted the value from `${1:19:999}` / `${2:...}` / `${3:...}`, so the substring offset landed on the wrong argument (exactly your `"EID=1"` example).

Since this PR was opened, that argument-parsing block was rewritten in **#1342 "More robust parsing and arg validation for build.sh"** (merged 2025-04-07). The new code parses args in a `case "$ARG" in … esac` loop and extracts with `CUSTOM_CMAKE_CXX_FLAG="\"${ARG:19:999}\""` — matching and extracting from the **same** `$ARG`, so the variable mismatch you fixed can no longer occur. `./build.sh CUSTOM_BUILD_FLAGS="-DHAVE_MAT_EVT_TRACEID=1" debug` now assigns the flags correctly on current `main`.

As a result the exact lines this PR patches (`$PARAM1`/`${1:19:999}` etc.) no longer exist on `main`, so it can't merge cleanly and the underlying issue is already resolved. Closing as superseded by #1342 — but thank you for the clear repro and the fix; it helped pin down the root cause.

PR #1308, opened 2024-11-19
update version to point to PG unified schema

nahidfgh:
Privacy Guard has an initiative for point to a unified schema. 
The new schema stores the data in the Privacy Concern Event. This idea has been implemented in https://github.com/microsoft/cpp_client_telemetry_modules/pull/253.

- Implemented Privacy Concern Event
- Integrated NotifyConcern with the new schema 
- Added tests to cover the new changes.

### DIFF
--- a/lib/include/public/Version.hpp
+++ b/lib/include/public/Version.hpp
@@ -6,8 +6,8 @@
 #define MAT_VERSION_HPP
 // WARNING: DO NOT MODIFY THIS FILE!
 // This file has been automatically generated, manual changes will be lost.
-#define BUILD_VERSION_STR "3.8.323.1"
-#define BUILD_VERSION 3,8,323,1
+#define BUILD_VERSION_STR "3.8.249.1"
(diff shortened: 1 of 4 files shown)


manuellopezleon (approved):

lib/include/public/Version.hpp
```diff
@@ -6,8 +6,8 @@
 #define MAT_VERSION_HPP
 // WARNING: DO NOT MODIFY THIS FILE!
 // This file has been automatically generated, manual changes will be lost.
-#define BUILD_VERSION_STR "3.8.249.1"
```
zgzjhzxhq, line 9: You can revert this change unless we need a new release. I think we will need a new release but it will be after the Java/Obj-C/Swift changes. So for now we just leave version.hpp as is.

zgzjhzxhq (approved):
Please revert version.hpp.

lalitb:
iOS builds are failing for older environment, which is not related to the PR. Will force-merge for now.

PR #1310, opened 2024-11-20
Privacy Concern Event wrapper

nahidfgh:


wrappers/obj-c/ODWPrivacyGuardInitConfig.h
```diff
@@ -24,6 +24,11 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property(readwrite, copy, nonatomic) NSString* semanticContextNotificationEventName;
 
+/*!
```
zgzjhzxhq, line 27: Please import header file

wrappers/obj-c/ODWPrivacyConcernEvent.h
```diff
@@ -0,0 +1,111 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+
+/*! 
+ @brief Represents a Privacy Concern Event.
+ This class is used to store and manage privacy concern event data including details
+ like the event name, server name, database name, and various other privacy-related
+ attributes.
+ */
+@interface PrivacyConcernEvent : NSObject
```
zgzjhzxhq, line 15: Please prefix `ODW` a.k.a. One DS Wrapper

wrappers/obj-c/ODWPrivacyGuard.mm
```diff
@@ -98,6 +98,10 @@ +(void)initializePrivacyGuard:(ILogger *)logger withODWPrivacyGuardInitConfig:(O
     {
         config.SemanticContextNotificationEventName = [[initConfigObject semanticContextNotificationEventName] UTF8String];
     }
+    if ([initConfigObject metadataProvider] != nill)
+    {
+        config.metadataProvider = [[]]
+    }
```
zgzjhzxhq, line 104: `config.metadataProvider` should be `config.MetadataProvider` defined [here](https://github.com/microsoft/cpp_client_telemetry_modules/blob/8539da62dd5e41e2847337df1d95f0f841d3a738/privacyguard/InitializationConfiguration.hpp). However, current implementation does not assign metadata provider to underlying data structure in PG C++. This is because in PG C++, Metadata provider is defined as 
`const std::shared_ptr<PrivacyConcernMetadataProvider> MetadataProvider;`, so you will need to convert your `ODWPrivacyConcernMetadataProvider` to this type. Unfortunately, there is no direct conversion between Obj-C and C++ class implementations. One way I found to achieve this is to use composition. In this [link](https://stackoverflow.com/a/9741994), it talks about how to do this.

I think this is what you need 
`
// h file
@interface ODWPrivacyConcernMetadataProvider: NSObject
{
    @public
    PrivacyConcernMetadataProvider *cppObject;
}

- (NSString *)getDatabaseNameForRecord:(id)record;
- // rest of interface functions

@end

// mm file
@implementation MyCPPWrapper

- (NSString *)getDatabaseNameForRecord:(id)record;
{
    if (cppObject)
        cppObject->GetDatabaseName(id);
}

@end
`

PR #1315, opened 2024-12-05
Support iOS Simulator on arm64

dgbeliveau:
Hi - I'm working on migrating my team's project to use the arm64 iOS simulator. I spent many hours today trying to track down this same issue in another one of our dependencies, so I thought I should contribute my findings back here.

It looks like for the x86_64 simulator, it will assume you want the simulator build even if you pass -miphoneos-version-min (because an x86 iphone makes no sense), but for the arm64 simulator, you need to specify you want the simulator with -mios-simulator-version-min.


lalitb (approved):
Confirmed with @dgbeliveau, the fix is working locally for him.

Issue #1316, opened 2024-12-07
Temporary path in Win32 needs to leverage GetTempPath2W when available


ankamat-msft:
Per recent documentation, GetTempPath2W() should be attempted before using GetTempPathW().

https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-gettemppathw#remarks

ThomsonTan:
@ankamat-msft I did see the preference on `GetTempPath2W()` in the doc link you shared. Could you please add the link?

PR #1317, opened 2024-12-10
Use GetTempPath2W whenever available

ankamat-msft:
Fixes microsoft/cpp_client_telemetry#1316

### DIFF
--- a/lib/utils/Utils.cpp
+++ b/lib/utils/Utils.cpp
@@ -125,10 +125,15 @@ namespace MAT_NS_BEGIN {
     std::string GetTempDirectory()
     {
 #ifdef _WIN32
+        auto lpGetTempPathW = reinterpret_cast<decltype(&::GetTempPathW)>(GetProcAddress(GetModuleHandle(TEXT("kernel32")), "GetTempPath2W"));
+        if (lpGetTempPathW == NULL)
+        {
+            lpGetTempPathW = ::GetTempPathW;
+        }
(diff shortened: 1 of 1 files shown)


lib/utils/Utils.cpp
```diff
@@ -125,10 +125,15 @@ namespace MAT_NS_BEGIN {
     std::string GetTempDirectory()
     {
 #ifdef _WIN32
+        auto lpGetTempPathW = reinterpret_cast<decltype(&::GetTempPathW)>(GetProcAddress(GetModuleHandle(TEXT("kernel32")), "GetTempPath2W"));
```
ThomsonTan, line 128: Could this be cached to avoid calling `GetProcAddress` every time to call `GetTempDirectory`?

ThomsonTan: Never mind, saw the below section.

> Apps should call [GetTempPath2](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-gettemppath2w) instead of GetTempPath.

ankamat-msft: Since this is related to security, it just feels better to keep this fully dynamic. Presumably this is not a super common method/call to cause performance issues, right?

lalitb: The current usage of `GetTempDirectory` is confined to the SDK initialization phase, and I believe the additional `GetProcAddress` call in this context is manageable.

lalitb (approved):

PR #1318, opened 2024-12-10
using COPY_FILE only with cmake 3.21+ where it is introduced

ipraveenMS:
This PR https://github.com/microsoft/cpp_client_telemetry/commit/c480743a8603dd34bb8b054877ced15bc48870c1 added COPY_FILE which is a new command in cmake 3.19 and not present in cmake 3.16.3-1ubuntu1.20.04.1. 

Fixing this by using COPY_FILE only with cmake 3.19+ where it is introduced

### DIFF
--- a/tests/functests/CMakeLists.txt
+++ b/tests/functests/CMakeLists.txt
@@ -35,7 +35,14 @@ if (EXISTS ${CMAKE_SOURCE_DIR}/lib/modules/exp/tests)
         ${CMAKE_SOURCE_DIR}/lib/modules/exp/tests/functests/ECSConfigCacheFuncTests.cpp
         )
     if (EXISTS ${CMAKE_SOURCE_DIR}/lib/modules/exp/tests/functests/test.json)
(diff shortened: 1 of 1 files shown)


tests/functests/CMakeLists.txt
```diff
@@ -35,7 +35,14 @@ if (EXISTS ${CMAKE_SOURCE_DIR}/lib/modules/exp/tests)
         ${CMAKE_SOURCE_DIR}/lib/modules/exp/tests/functests/ECSConfigCacheFuncTests.cpp
         )
     if (EXISTS ${CMAKE_SOURCE_DIR}/lib/modules/exp/tests/functests/test.json)
-      file(COPY_FILE ${CMAKE_SOURCE_DIR}/lib/modules/exp/tests/functests/test.json ${CMAKE_BINARY_DIR}/test.json)
+      if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.19")
```
ThomsonTan, line 38: It seems `COPY_FILE` was added in `3.21`?

https://cmake.org/cmake/help/latest/command/file.html

ipraveenMS: Thanks!. Fixed

ThomsonTan (approved):

PR #1322, opened 2025-01-22
Bump version for new release - "3.9.22.1"

lalitb:
Also fix the version generation script to increment the minor and patch release properly. Somehow it stopped worked with previous commits.

### DIFF
--- a/tools/version-node.js
+++ b/tools/version-node.js
@@ -1,17 +1,21 @@
 const fs = require('fs');
 
 function dayNumber() {
-  var now = new Date();
-  var start = new Date(now.getFullYear(), 0, 0);
-  var diff = now - start;
-  var oneDay = 1000 * 60 * 60 * 24;
-  var dayOfYear = Math.floor(diff / oneDay);
-  return dayOfYear;
+  const now = new Date();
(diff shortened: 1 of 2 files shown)


tools/version-node.js
```diff
@@ -9,6 +9,12 @@ function dayNumber() {
   return dayOfYear;
 }
 
+function calculateMinorVersion() {
+  var baseYear = 2020; // VER_MINOR starts at 4 in 2020
```
ThomsonTan, line 13: Where is this rule from? Probably just make minor version the lower 2 digits of year number which make it more visual friendly?

lalitb: It's documented here - https://github.com/microsoft/cpp_client_telemetry/blob/962db1ad555a0af634469f6ad7bb55b55c7c08a8/docs/versioning.md#release-and-version-strategy

VER_MINOR is 4. This number has to be incremented to 5 on January 1st, 2021.

lalitb: > Where is this rule from? Probably just make minor version the lower 2 digits of year number which make it more visual friendly?

Yes, can be done. Just prefer to maintain the incremental number from previous releases as of now.

PR #1325, opened 2025-02-03
Move json.hpp to a peer folder of other third-party dependencies - Stage 1

mkoscumb:
Beyond being good dependency hygiene, this will also allow adopters which have their own copy of json.hpp to more easily use that when integrating the 1DS C++ SDK (via include path modification)

This work will be broken down into three PRs. 
1. #This PR# Copy json.hpp from lib\include\mat => nlohmann\ update include paths accordingly.
2. Update include paths in the private modules repo.
3. Remove lib\include\mat\json.hpp and update the private modules repo commit to the result of Stage 2.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -301,7 +301,7 @@ include_directories(lib/pal)
 #  add_subdirectory(bondlite/tests)
 #endif()
 
-# Include repo root to allow includes of sqlite, zlib, and nlhomann
+# Include repo root to allow includes of sqlite, zlib, and nlohmann
 include_directories(${CMAKE_SOURCE_DIR})
 
 if(BUILD_HEADERS)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -302,7 +302,7 @@ include_directories(lib/pal)
 #endif()
 
 # Include repo root to allow includes of sqlite, zlib, and nlhomann
-include_directories(${CMAKE_CURRENT_SOURCE_DIR})
+include_directories(${CMAKE_SOURCE_DIR})
 
 if(BUILD_HEADERS)
   add_subdirectory(lib/include)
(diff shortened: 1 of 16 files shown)


lalitb (approved):

CMakeLists.txt
```diff
@@ -301,6 +301,9 @@ include_directories(lib/pal)
 #  add_subdirectory(bondlite/tests)
 #endif()
 
+# Include repo root to allow includes of sqlite, zlib, and nlhomann
```
ThomsonTan, line 304: >nlhomann [](http://example.com/codeflow?start=59&length=8)

nit: nlohmann

ThomsonTan:
The new json.hpp file seems cannot be loaded into github for review because it has too much lines.

Can we download this `json.hpp` file from github during build instead of copying it?

lalitb:
Unfortunately, macOS build is still not working, it's on me. Hopefully, this doesn't break it. We should be able to force-merge it for now.

lalitb:
> The new json.hpp file seems cannot be loaded into github for review because it has too much lines.
> Can we download this json.hpp file from github during build instead of copying it?

@ThomsonTan - It's actually copied from lib/mat directory, so content needn't be reviewed. Because of 3 step process, it can't be moved from lib/mat to root directory. I think downloading from Github can be tracked separately.

ThomsonTan:
> > The new json.hpp file seems cannot be loaded into github for review because it has too much lines.
> > Can we download this json.hpp file from github during build instead of copying it?
> 
> @ThomsonTan - It's actually copied from lib/mat directory to new location, so content needn't be reviewed. Because of 3 step process, it can't be moved from lib/mat to root directory. I think downloading from Github can be tracked separately.

Ok, comparing large files is not very intuitive in the revie UI, but I've verified the file move.

ThomsonTan (approved):
:shipit:

CMakeLists.txt
mkoscumb: Good catch, fixed.

PR #1330, opened 2025-02-11
Clean up lib\include\mat\json.hpp and update Private Modules Commit ID.

mkoscumb:
Beyond being good dependency hygiene, this will also allow adopters which have their own copy of json.hpp to more easily use that when integrating the 1DS C++ SDK (via include path modification)

This work will be broken down into three PRs.

1. This PR Copy json.hpp from lib\include\mat => nlohmann\ update include paths accordingly. - https://github.com/microsoft/cpp_client_telemetry/pull/1325
2. Update include paths in the private modules repo.
3. #This PR# Remove lib\include\mat\json.hpp and update the private modules repo commit to the result of Stage 2.


pablo-msft:
This PR is out of date and has conflicts because the modules have been updated. The deletion of lib/include/mat/json.hpp has been replicated in #1454. Closing.

PR #1332, opened 2025-03-04
Fix CodeQL violations of type 'cpp/path-injection in lib/pal

abuhsayem:
CodeQL has flagged one or more instances of violations of type "cpp/path-injection" in the file "third_party/cpp_client_telemetry/lib/pal/PAL.cpp".

Violation File Path: third_party/cpp_client_telemetry/lib/pal/PAL.cpp
Line Numbers: 123
Alert Message: Uncontrolled data used in path expression.
QueryID: cpp/path-injection

Example [fix](https://liquid.microsoft.com/Web/Object/Read/ScanningToolWarnings/Requirements/CodeQL.SM01937#Zguide)

### DIFF
--- a/lib/pal/PAL.cpp
+++ b/lib/pal/PAL.cpp
@@ -126,7 +126,7 @@ namespace PAL_NS_BEGIN {
 
             if (!pathExists || containsParentDirectory)
             {
-                std::cerr << "Invalid trace folder path." << std::endl;
+                throw std::runtime_error("Invalid trace folder path.");
                 return false;
             }
 
--- a/lib/pal/PAL.cpp
+++ b/lib/pal/PAL.cpp
@@ -126,7 +126,7 @@ namespace PAL_NS_BEGIN {
 
             if (!pathExists || containsParentDirectory)
             {
-                assert(false && "Invalid trace folder path.");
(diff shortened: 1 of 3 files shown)


lib/pal/PAL.cpp
```diff
@@ -113,8 +114,27 @@ namespace PAL_NS_BEGIN {
...
+            bool pathExists = (fileAttr != INVALID_FILE_ATTRIBUTES && (fileAttr & FILE_ATTRIBUTE_DIRECTORY));
+#else
+            bool pathExists = (access(traceFolderPath.c_str(), F_OK) != -1);
+#endif
+            // Check if the path contains ".."
+            bool containsParentDirectory = (traceFolderPath.find("..") != std::string::npos);
+
+            if (!pathExists || containsParentDirectory)
+            {
+                throw std::runtime_error("Invalid trace folder path.");
```
ThomsonTan, line 129: Is throwing exception necessary here? Or return false?

abuhsayem: It is not necessary since the function's return type is bool. Replaced with 'return false'.

ThomsonTan (approved):
:shipit:

PR #1340, opened 2025-04-02
[CI] set CMake minimum version requirement via env var for the break by CMake 4

ThomsonTan:


ThomsonTan:
The CMake 4.0 compatibility issue was fixed. The test failure is caused by some unstable test which is unrelated.

lalitb (approved):

Issue #1341, opened 2025-04-04
build.sh swallows -D switch if a positional argument follows it


DSMitten:
Repro (on Linux):

```bash
./build.sh -DBUILD_UNIT_TESTS=NO -DBUILD_FUNC_TESTS=NO noroot
```

The result is that the build script parses `noroot`, then executes a `shift` commands which causes the first argument (-DBUILD_UNIT_TESTS=NO) to not be evaluated by getopt

Positional arguments should probably only be valid if they precede the switches, and any arguments trailing the switches should probably result in an error.

lalitb:
#1342 fixes it.

PR #1342, opened 2025-04-04
More robust parsing and arg validation for build.sh

DSMitten:
The issue prompting this PR is [#1341](https://github.com/microsoft/cpp_client_telemetry/issues/1341), where a positional argument following a switch will cause the switch to be swallowed and not evaluated.

This PR changes how commands are parsed so that:
* All positional arguments are parsed before any switches, in a loop
* Any number of positional arguments will be evaluated
* If there is a conflict (such as 'release' and 'debug', or 'arm64' and 'x86_64'), the error will be detected and the script will terminate
* All parsing occurs before any work (mainly affects the clean step)
* If any arguments follow the final switch, the error will be detected and the script will terminate

Additionally, there was a conflict between the usage statement and logic, in that the usage statement referred to a positional argument `CUSTOM_CMAKE_CXX_FLAGS=x`, but the actual parser looked for an argument `CUSTOM_BUILD_FLAGS=x` and set a variable `CUSTOM_CMAKE_CXX_FLAG`. The usage statement is updated to reflect the argument.

Testing performed includes:
* All arguments and the variables they set
* Conflicting arguments result in errors
* Default values are used if no value is set for BUILD_TYPE or MAC_ARCH
* Non-switch arguments following switches result in errors
* `./build.sh noroot -DBUILD_UNIT_TESTS=NO -DBUILD_FUNC_TESTS=NO` builds successfully

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -2,16 +2,16 @@
 
 usage()
 {
-  echo "Usage: build.sh [clean] [arm64|x86_64|universal] [CUSTOM_CMAKE_CXX_FLAGS=x] [noroot] [release] [-h|-?] [-l (static|shared)] [-D CMAKE_OPTION] [-v]"
+  echo "Usage: build.sh [clean] [arm64|x86_64|universal] [CUSTOM_BUILD_FLAGS=x] [noroot] [release|debug] [-h|-?] [-l (static|shared)] [-D CMAKE_OPTION] [-v]"
   echo "                                                                                         "
   echo "options:                                                                                 "
   echo "                                                                                         "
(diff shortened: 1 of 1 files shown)


ThomsonTan (approved):
:shipit:

lalitb:
@DSMitten - can you ensure that this doesn't break the existing scripts which internally invoke build.sh . As here - https://github.com/microsoft/cpp_client_telemetry/blob/c380bc88d7e0b7de86c09bd3b3e9f894481e4323/build-docker.sh#L15

lalitb (approved):

Issue #1370, opened 2025-08-12
Crash on iOS [ODWLogManager flushAndTeardown]


denny-msft:
iOS ODWReachability::stopNotifier is missing a return causing a crash on tear down.

| **Module**                          | **Function / Method**                                                                 | **File**                        | **Line** |
|------------------------------------|----------------------------------------------------------------------------------------|----------------------------------|----------|
| `libsystem_pthread.dylib`          | `pthread_mutex_lock$VARIANT$armv81`                                                   | —                                | —        |
| `SystemConfiguration`              | `SCNetworkReachabilitySetCallback`                                                    | —                                | —        |
| `Telemetry`                        | `-[ODWReachability stopNotifier]`                                                     | `ODWReachability.m`              | 367      |
| `Telemetry`                        | `-[ODWReachability dealloc]`                                                          | `ODWReachability.m`              | 265      |
| `Telemetry`                        | `Microsoft::Applications::Events::PlatformAbstraction::NetworkInformation::~NetworkInformation()` | `NetworkInformationImpl.mm`     | 102      |
| `Telemetry`                        | `_ZNSt3__110shared_ptrIN9Microsoft12Applications6Events15ITaskDispatcherEEaSB8ne180100EOS5_` | `shared_ptr.h`                  | 157      |
| `Telemetry`                        | `Microsoft::Applications::Events::PlatformAbstraction::PlatformAbstractionLayer::shutdown()` | `PAL.cpp`                        | 626      |
| `Telemetry`                        | `Microsoft::Applications::Events::LogManagerImpl::FlushAndTeardown()`                 | `PAL.hpp`                        | 215      |
| `Telemetry`                        | `Microsoft::Applications::Events::LogManagerBase<Microsoft::Applications::Events::ModuleLogConfiguration>::FlushAndTeardown()` | `LogManagerBase.hpp`            | 290      |
| `Telemetry`                        | `+[ODWLogManager flushAndTeardown]`                                                   | `ODWLogManager.mm`               | 267      |

Introduced in #1323 with v3.9.163.1

PR #1371, opened 2025-08-13
Remove ODWReachability and iOS 11 macOS 10.14 code with EOL for those platforms

denny-msft:
Started investigating the crash from #1370 and realized that ODWReachability _should_ only be applicable for platforms below iOS 11 and macOS 10.14. However, a crash stack from an iOS 18 device clearly shows these classes being deallocated.

```
-[ODWReachability stopNotifier] ODWReachability.m:367
-[ODWReachability dealloc] ODWReachability.m:265
Microsoft::Applications::Events::PlatformAbstraction::NetworkInformation::~NetworkInformation() NetworkInformationImpl.mm:102
```

Seems like there was a bug where these reachability classes were being initialized without guarded availability checks (see NetworkInformation::SetupNetDetect:108) when they should only be used for lower platforms.

Regardless, given the EOL of these versions, this PR remove references to ODWReachability and fixes the crash.

# Testing
Verified with build-ios.sh.


lalitb:
@absaroj - Can you help with the review. thanks.

lib/pal/posix/NetworkInformationImpl.mm
```diff
@@ -84,117 +78,64 @@ virtual NetworkCost GetNetworkCost()
...
         }
     }
 
     void NetworkInformation::SetupNetDetect()
     {
         auto weak_this = std::weak_ptr<NetworkInformation>(shared_from_this());
 
-        m_reach = [ODWReachability reachabilityForInternetConnection];
-        void (^block)(NSNotification*) = ^(NSNotification*)
+        m_monitor = nw_path_monitor_create();
```
lalitb: We should handle scenario where `nw_path_monitor_create` fails, and return `nil` - may be rare scenario but under system resource exhaustion?

lalitb: This is existing code, so probably not blocker here.

thomasameisel:
> One question: This change raises the minimum iOS version from 10.0 to 12.0. Are there any concerns about existing customer devices still running iOS 10/11, or is this acceptable given that Apple has long since ended support for these versions?

I can only speak for iOS Outlook, but our min version is 18.0.

denny-msft:
> One question: This change raises the minimum iOS version from 10.0 to 12.0. Are there any concerns about existing customer devices still running iOS 10/11, or is this acceptable given that Apple has long since ended support for these versions?

App Store apps are required to be built with Xcode 16 or later as of [April 2025](https://developer.apple.com/news/upcoming-requirements/?id=02212025a), and only supports [iOS 15 or later](https://developer.apple.com/support/xcode/), so we should be okay.

bmehta001:
Closing in favor of #1431, which fixes the same #1370 crash (the SDK was initializing `ODWReachability` in `NetworkInformationImpl::SetupNetDetect` without availability checks) and additionally preserves the vendored `ODWReachability` public surface by re-backing it with `NWPathMonitor` for external consumers, while gating the legacy `SCNetworkReachability` notifier path behind `ODW_LEGACY_REACHABILITY_REQUIRED` for sub-iOS-12 / sub-macOS-10.14 deployment targets. The `build-ios.sh` default deployment-target bump to 12.0 from this PR has been ported into #1431. Thanks @denny-msft for the original investigation.

PR #1376, opened 2025-09-03
fix extra semi

frank-msodsp:
fix extra semi


ThomsonTan:
What is the effect of the incorrect semi?

frank-msodsp:
> What is the effect of the incorrect semi?

it was brought to my attention that unix builds are failing locally

PR #1379, opened 2025-09-11
add san to log manager

frank-msodsp:
Add JNI support for san
fix san jni with improper use of c_str
Added  san support to logmanager


lalitb:
@ThomsonTan - Probably we need force merge as CI failure is unrelated, while we are looking into CI failure?

PR #1388, opened 2025-10-29
update version sync modules

frank-msodsp:
update version
sync modules commit


lalitb:
force merge - as the errors are not related.

PR #1415, opened 2026-03-18
Fix CI workflows, compiler flags, and flaky tests

bmehta001:
CI/build/test fixes only — no runtime behavior changes. Runtime fixes (data races, memory leak, shutdown safety) are in #1429.

### Modernize CI workflows (11 files)

- Update all GitHub Actions to latest versions (checkout@v4, setup-java@v5, setup-android@v3, etc.)
- Add concurrency groups with `cancel-in-progress` scoped to `pull_request` events only — push builds on `main` always run to completion so a rapid second push can't hide a broken commit
- Add 30-minute `timeout-minutes` to iOS CI to prevent deadlock hangs (the `CancelAllRequests` spin loop can hang indefinitely if an NSURLSession completion handler never fires on the simulator)
- Fix iOS simulator destination: resolve simulator UUID via `xcrun simctl list` to avoid ambiguous `-destination` matching
- Fix Android SDK path detection in CodeQL workflow

### Fix compiler flags (2 files)

- Split GCC and Clang warning flags — `-Wno-unknown-warning-option` is Clang-only and was breaking GCC builds
- Add `-fno-finite-math-only` to all non-MSVC release flags — restores IEEE 754 compliance (`INFINITY` macro) needed by sqlite3 and tests when `-ffast-math` is enabled
- Remove global `-Wno-reorder`, scope to `Sanitizer.cpp` only via `set_source_files_properties` — the only file that triggers the warning is in the `lib/modules` submodule where `Sanitizer.hpp` declares members in a different order than the constructor init list

### Improve iOS test script and documentation (3 files)

- Rewrite `build-tests-ios.sh` to dynamically resolve simulator UUID for unambiguous xcodebuild destination
- Document python3 requirement in README
- Update `docs/cpp-start-ios.md` with current build instructions

### Fix flaky functional tests (6 files)

- **Unique phase2 DB paths** (`APITest.cpp`): replace `PAL::sleep(500)` with unique-per-invocation DB paths using an atomic counter. The SDK closes SQLite via `sqlite3_close_v2()` which defers file-descriptor cleanup — reusing a fixed path and `std::remove()`-ing files with deferred fds still open caused iOS `vnode unlinked while in use` errors
- **SQLite journal cleanup** (`APITest.cpp`, `AISendTests.cpp`): remove `-wal`, `-shm`, `-journal` files in `CleanStorage()` to prevent cross-test contamination
- **Config contamination** (`BasicFuncTests.cpp`): reset `CFG_INT_CACHE_FILE_SIZE` and `CFG_INT_STORAGE_FULL_PCT` in test `Initialize()` — `LogManager::GetLogConfiguration()` returns a global singleton, so tests that modify config fields contaminate subsequent tests
- **yield() → sleep(10)** (`BasicFuncTests.cpp`): reduces CPU contention on single-core iOS simulator runners and gives the network stack time to deliver
- **Multi-batch upload flakiness** (`BasicFuncTests.cpp`): add `SetTransmitProfile(TransmitProfile_RealTime)` before `UploadNow()` to set all timer delays to 0 — prevents events from being held back in lower-priority batches
- **Use 127.0.0.1** (`BasicFuncTests.cpp`): replace `localhost` with `127.0.0.1` for the local test HTTP server — avoids DNS resolution failures on iOS simulator
- **Increased timeouts**: widen `waitForEvents` timeouts across tests to accommodate slow CI runners
- **Atomic counter** (`MultipleLogManagersTests.cpp`): fix data race on shared event counter

### DIFF
--- a/.github/workflows/build-android.yml
+++ b/.github/workflows/build-android.yml
@@ -17,13 +17,18 @@ on:
     - main
     - dev
 
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 jobs:
   build:
     runs-on: windows-latest
     name: Build for Android
     steps:
     - name: Checkout
-      uses: actions/checkout@v3
+      uses: actions/checkout@v4
       with:
         submodules: false
     - name: Update submodules
@@ -32,7 +37,7 @@ jobs:
         git config --global submodule.lib/modules.update none
         git -c protocol.version=2 submodule update --init --force --depth=1
     - name: Setup Java
-      uses: actions/setup-java@v3
+      uses: actions/setup-java@v5
       with:
         distribution: 'adopt'
         java-version: '17'
@@ -40,14 +45,13 @@ jobs:
       # Workaround for: 'Unable to decrypt local Maven settings credentials'
       run: rm $Env:USERPROFILE\.m2\settings.xml
     - name: Setup Android SDK
-      uses: android-actions/setup-android@v2
+      uses: android-actions/setup-android@v3
     - name: Install NDK
       run: |
         java -version
         gci env:* | sort-object name
-        new-item "C:\Users\runneradmin\.android\repositories.cfg" -ItemType "file"
-        echo yes | .\sdkmanager.bat "ndk-bundle" "cmake;3.10.2.4988404" "ndk;27.0.12077973" --sdk_root=$Env:ANDROID_SDK_ROOT
-      working-directory: ${{ env.ANDROID_SDK_ROOT }}\cmdline-tools\7.0\bin
+        new-item "$Env:USERPROFILE\.android\repositories.cfg" -ItemType "file"
+        echo yes | sdkmanager "ndk-bundle" "cmake;3.10.2.4988404" "ndk;27.0.12077973" --sdk_root=$Env:ANDROID_SDK_ROOT
     - name: Chocolatey
       run: |
         choco install --no-progress -y ninja
--- a/.github/workflows/build-ios-mac.yml
+++ b/.github/workflows/build-ios-mac.yml
@@ -19,36 +19,42 @@ on:
   schedule:
   - cron: 0 2 * * 1-5
 
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 jobs:
   build:
     strategy:
       matrix:
-        os: [macos-13, macos-15]
+        os: [macos-14, macos-15]
         config: [release, debug]
         simulator: ["'iPhone 15'", "'iPad Pro (11-inch) (4th generation)'", "'iPhone 16'", "'iPad Air 11-inch (M2)'"]
         exclude:
-          - os: macos-13
+          - os: macos-14
             simulator: "'iPhone 16'"
-          - os: macos-13
+          - os: macos-14
             simulator: "'iPad Air 11-inch (M2)'"
           - os: macos-15
             simulator: "'iPhone 15'"
           - os: macos-15
             simulator: "'iPad Pro (11-inch) (4th generation)'"
     runs-on: ${{ matrix.os }}
+    timeout-minutes: 30
     env:
       CMAKE_POLICY_VERSION_MINIMUM: "3.5"
     steps:
     - name: Grant write permissions to /usr/local
       run: |
         sudo chown -R $USER:staff /usr/local
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v4
       with:
         submodules: 'true'
       continue-on-error: true
     - name: build
       run: |
-        if [[ "${{ matrix.os }}" == "macos-13" ]]; then
+        if [[ "${{ matrix.os }}" == "macos-14" ]]; then
           export IOS_DEPLOYMENT_TARGET=13.0;
         elif [[ "${{ matrix.os }}" == "macos-15" ]]; then
           export IOS_DEPLOYMENT_TARGET=15.0;
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,23 +115,26 @@ message("-- CMAKE_CXX_COMPILER_ID:  ${CMAKE_CXX_COMPILER_ID}")
 include(tools/ParseOsRelease.cmake)
 
 if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
-set(WARN_FLAGS "/W4 /WX")
+  set(WARN_FLAGS "/W4 /WX")
+elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+  # -Wno-unknown-warning-option is Clang-only, omitted here
+  set(WARN_FLAGS "-Wall -Werror -Wextra -Wno-unused-parameter -Wno-unused-but-set-variable")
 else()
-# No -pedantic -Wno-extra-semi -Wno-gnu-zero-variadic-macro-arguments
-set(WARN_FLAGS "-Wall -Werror -Wextra -Wno-unused-parameter -Wno-unknown-warning-option -Wno-unused-but-set-variable")
+  # Clang / AppleClang
+  set(WARN_FLAGS "-Wall -Werror -Wextra -Wno-unused-parameter -Wno-unknown-warning-option -Wno-unused-but-set-variable")
 endif()
 
 if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
   # Using GCC with -s and -Wl linker flags
-  set(REL_FLAGS "-s -Wl,--gc-sections -Os ${WARN_FLAGS} -ffunction-sections -fdata-sections -fmerge-all-constants -ffast-math")
+  set(REL_FLAGS "-s -Wl,--gc-sections -Os ${WARN_FLAGS} -ffunction-sections -fdata-sections -fmerge-all-constants -ffast-math -fno-finite-math-only")
 elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
   set(REL_FLAGS "${WARN_FLAGS}")
 elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "AppleClang")
   # AppleClang does not support -ffunction-sections and -fdata-sections with the -fembed-bitcode and -fembed-bitcode-marker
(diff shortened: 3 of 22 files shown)


lalitb (changes requested):
for the comments.

.github/workflows/build-android.yml
```diff
@@ -17,13 +17,18 @@ on:
     - main
     - dev
 
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
ThomsonTan, line 23: Please scope `cancel-in-progress` to PR events. As written, a rapid second push to `main` cancels the first build — so we could merge a broken commit and only find out on the next trigger.

lib/pal/WorkerThread.cpp
```diff
@@ -56,22 +56,26 @@ namespace PAL_NS_BEGIN {
...
-                LOG_WARN("m_queue is not empty!");
-            }
-            if (!m_timerQueue.empty())
-            {
-                LOG_WARN("m_timerQueue is not empty!");
+            // Clean up any tasks remaining in the queues after shutdown.
+            // Only safe after join() — the thread has fully exited.
+            // After detach(), the thread still needs the shutdown item
+            // and may still be accessing the queues.
+            if (joined) {
```
ThomsonTan, line 74: Good catch on the detach-vs-free ordering. But we've also lost the only signal that shutdown dropped pending work - the old `LOG_WARN` on non-empty queues. Could we log the sizes before detach so operators can still see it?

CMakeLists.txt
```diff
@@ -115,23 +115,30 @@ message("-- CMAKE_CXX_COMPILER_ID:  ${CMAKE_CXX_COMPILER_ID}")
 include(tools/ParseOsRelease.cmake)
 
 if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
-set(WARN_FLAGS "/W4 /WX")
+  set(WARN_FLAGS "/W4 /WX")
+elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+  # -Wno-unknown-warning-option is Clang-only, omitted here
+  set(WARN_FLAGS "-Wall -Werror -Wextra -Wno-unused-parameter -Wno-unused-but-set-variable")
+  # -Wno-reorder is C++-only; added to CXX_FLAGS below (suppresses member-init-order warnings in submodule code)
+  set(CXX_EXTRA_WARN_FLAGS "-Wno-reorder")
```
ThomsonTan, line 123: Enable `-Wno-reorder` to supress member-init-order may hide bugs. Could this option to moved to the specific module which triggers it?

lib/tpm/TransmissionPolicyManager.cpp
```diff
@@ -185,10 +185,9 @@ namespace MAT_NS_BEGIN {
             return;
         }
         m_runningLatency = latency;
-        m_scheduledUploadTime = std::numeric_limits<uint64_t>::max();
-
         {
             LOCKGUARD(m_scheduledUploadMutex);
+            m_scheduledUploadTime = std::numeric_limits<uint64_t>::max();
```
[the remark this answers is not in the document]
ThomsonTan: Agreed, atomic works here.

Nit: a 1-line comment above L187–188 noting the unlocked stores are intentional would help. The sentinel write is safe because `m_isUploadScheduled = false` under the lock is what actually publishes "no scheduled upload" to readers.

Not blocking.

.github/workflows/build-android.yml
bmehta001: Fixed

lib/tpm/TransmissionPolicyManager.cpp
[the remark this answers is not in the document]
bmehta001: Added

CMakeLists.txt
bmehta001: Moved to Sanitizer

lib/pal/WorkerThread.cpp
bmehta001: Logging

lib/http/HttpClient_Apple.mm
```diff
@@ -161,14 +139,38 @@ void Cancel()
...
+            NSURLSessionConfiguration* sessionConfig = [NSURLSessionConfiguration ephemeralSessionConfiguration];
+            m_session = (__bridge_retained void*)[NSURLSession sessionWithConfiguration:sessionConfig];
+        }
+    }
+}
+
+void* HttpClient_Apple::GetOrCreateSession()
+{
+    ensureSession();
+    return m_session;
```
ThomsonTan, line 171: Fix the similar issue on L219 which does torn read on m_requests while all updates are locked?

bmehta001: Thanks, missed that! Fixed.

build-tests-ios.sh
```diff
@@ -7,14 +7,23 @@ set -e
...
-
 cd tests/unittests
 
 xcrun simctl list devices available
 echo 'End of xcrun simctl list devices available'
 
-xcodebuild test -scheme iOSUnitTests -destination "platform=iOS Simulator,name=$SIMULATOR"
+# Resolve simulator UUID to avoid ambiguous destination matching.
+# Grab the last (newest OS) available device matching the requested name.
+SIM_ID=$(xcrun simctl list devices available | grep "$SIMULATOR" | grep -oE '[A-F0-9-]{36}' | tail -1)
```
lalitb: Can we resolve the simulator from `xcrun simctl list devices available --json` and require `device.name == "$SIMULATOR"` before taking the UDID?  

`grep "$SIMULATOR"` is more of substring match, and can resolve to multiple devices.

.github/workflows/build-ios-mac.yml
```diff
@@ -19,37 +19,43 @@ on:
...
+    - uses: actions/checkout@v4
       with:
         submodules: 'true'
       continue-on-error: true
     - name: build
       run: |
-        if [[ "${{ matrix.os }}" == "macos-13" ]]; then
-          export IOS_DEPLOYMENT_TARGET=13.0;
+        if [[ "${{ matrix.os }}" == "macos-14" ]]; then
+          export IOS_DEPLOYMENT_TARGET=14.0;
```
lalitb: The runner update from `macos-13` to `macos-14` makes sense, but this also raises the CI deployment target from `IOS_DEPLOYMENT_TARGET=13.0` to `14.0`. Is that intentional?

  The current `macos-14` log shows the job is using Xcode 15.4, which still supports lower iOS deployment targets, and `README.md` still documents iOS 12+ as supported. Since runner OS and deployment target are independent, I think we should keep this leg on `macos-14` but continue exporting `IOS_DEPLOYMENT_TARGET=13.0` unless we are intentionally dropping that lower-iOS coverage

lib/http/HttpClient_Apple.hpp
```diff
@@ -25,10 +25,13 @@ namespace MAT_NS_BEGIN {
 
         void Erase(IHttpRequest* req);
         void Add(IHttpRequest* req);
+        void* GetOrCreateSession();
 
     private:
+        void ensureSession();
         std::mutex m_requestsMtx;
         std::map<std::string, IHttpRequest*> m_requests;
+        void* m_session = nullptr;  // NSURLSession*, opaque in header
```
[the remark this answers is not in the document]
lalitb: I think we can fix the original cancellation bug with a smaller change: keep the existing shared `NSURLSession` (`defaultSessionConfiguration` + `dispatch_once`) and just remove
 the `[session getTasksWithCompletionHandler:]` block from `HttpRequestApple::Cancel()`, so we only cancel `m_dataTask`.

  That async block is the source of the cross-instance cancel: it uses the static session after SDK teardown and can end up cancelling tasks from a newly initialized `HttpClient_Apple`. Scoping cancellation to `m_dataTask` should fix that without changing session ownership.

  The per-instance session and `ephemeralSessionConfiguration` changes feel broader than needed and introduce the `m_session` lifetime concern discussed above. If there is a separate stale-connection/test flake being addressed by `ephemeralSessionConfiguration`, I’d prefer to handle that separately rather than changing production session behavior in this PR.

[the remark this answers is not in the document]
bmehta001: Thanks, I will make this change.

.github/workflows/build-ios-mac.yml
bmehta001: Not intentional - reverting

build-tests-ios.sh
bmehta001: Using json and python3 for parsing for more specific device resolution

lib/http/HttpClient_Apple.hpp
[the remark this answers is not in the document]
bmehta001: Will consider a middle-ground solution in a separate PR that prevents stale connections from causing issue, but for now, ignoring that case

lalitb:
@bmehta001 -  The CI fixes are important and also more safe, but this PR also changes some runtime behavior in the HTTP client, WorkerThread, TPM, Logger, and HttpResponseDecoder.

  Can we keep this PR scoped to the CI/build/test fixes, and move the runtime race/shutdown fixes to a separate PR? If any of these runtime changes are needed to unblock CI, can you call that out with the specific failure they address and how it was validated?

lib/pal/WorkerThread.cpp
```diff
@@ -56,22 +56,32 @@ namespace PAL_NS_BEGIN {
...
-                if (m_hThread.joinable() && (m_hThread.get_id() != this_id))
+                if (m_hThread.joinable() && (m_hThread.get_id() != this_id)) {
                     m_hThread.join();
-                else
+                    joined = true;
+                } else {
                     m_hThread.detach();
+                }
             }
             catch (...) {};
```
ThomsonTan: The catch-all here swallows `std::system_error` from `join()`/`detach()` and leaves `joined=false`, so the new cleanup block silently no-ops and tasks leak with zero diagnostic output.

`std::thread::join()` legitimately throws (`ESRCH`/`EINVAL`/`EDEADLK`/`EPERM`) on the iOS-simulator and forced-teardown paths this PR is targeting - exactly the environments where we want a log line, not silence.

Suggest catching `std::system_error` (and `std::exception` as fallback) and emitting a `LOG_ERROR` with `code()`+`what()`. Also worth moving the queue-depth `LOG_WARN`s out of `if (joined)` so the detach path keeps the shutdown-with-pending-work signal.

bmehta001: Thank you, fixed

bmehta001:
@lalitb - Yes, will remove all functional changes to put in another PR, even if a few tests fail in the CI on this branch, just to keep things simple.

lalitb (approved):
Thanks @bmehta001 - Nice work, Appreciate making the CI green!

PR #1427, opened 2026-04-20
Remove dead headers

### LINKED ISSUE #1418: New compiler error in Xcode 26.4
When building the SDK in Xcode 26.4, I am now getting a new error message:

ODWReachability.m:32:9 Use of private header from outside its module: 'netinet6/in6.h'

I believe this is is due to new compiler settings in Xcode as it wasn't an issue in Xcode 26.3. 

According to CoPilot, the include of netinet6/in6.h isn't necessary as nothing is used from that header file (not sure if this is true, but I did comment the line out and it built).

Please advise,  I can create a PR with that line removed if that would be helpful.

bmehta001:
- Remove dead netinet6/in6.h and other imports from ODWReachability and its Objective-C unit test so Apple builds no longer fail on the private-header error from main

Fixes #1418

### DIFF
--- a/tests/unittests/obj-c/ODWReachabilityTests.mm
+++ b/tests/unittests/obj-c/ODWReachabilityTests.mm
@@ -11,6 +11,8 @@
 #import <XCTest/XCTest.h>
 #import "ODWReachability.h"
 
+#import <sys/socket.h>
+#import <netinet/in.h>
 #import <arpa/inet.h>
 
 @interface ODWReachabilityTests : XCTestCase
@@ -104,4 +106,3 @@
 }
 
 @end
-
--- a/tests/unittests/obj-c/ODWReachabilityTests.mm
+++ b/tests/unittests/obj-c/ODWReachabilityTests.mm
@@ -11,12 +11,7 @@
 #import <XCTest/XCTest.h>
 #import "ODWReachability.h"
 
-#import <sys/socket.h>
-#import <netinet/in.h>
 #import <arpa/inet.h>
-#import <ifaddrs.h>
-#import <netdb.h>
-#import <Foundation/Foundation.h>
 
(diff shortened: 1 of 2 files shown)


tests/unittests/obj-c/ODWReachabilityTests.mm
```diff
@@ -11,13 +11,7 @@
 #import <XCTest/XCTest.h>
 #import "ODWReachability.h"
 
-#import <sys/socket.h>
```
ThomsonTan, line 14: The `netinet6/in6.h` removal fixes #1418, but this file still uses `struct sockaddr_in`, `AF_INET`, `IN_LINKLOCALNETNUM`, and `htonl`, whose canonical headers are `/`.

third_party/Reachability/ODWReachability.m
```diff
@@ -27,13 +27,7 @@ CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
 
 #import "ODWReachability.h"
 
-#import <sys/socket.h>
```
ThomsonTan, line 30: Same as https://github.com/microsoft/cpp_client_telemetry/pull/1427/changes#r3119482564

tests/unittests/obj-c/ODWReachabilityTests.mm
bmehta001: I was relying on headers being brought transitively (arpa/inet.h brings in netinet/in.h and ODWReachability.h brings in SystemConfiguration, which brings in sys/socket.h), so had removed the headers, but I am now reverting that changes to keep the canonical header files where these are defined explicit. I will now only use this PR to remove dead headers.

ThomsonTan (approved):

PR #1441, opened 2026-05-18
Fix Windows desktop OS build version reporting

bmehta001:
## Summary
- Stop using stale BuildLabEx to populate the Windows desktop OsBuild / xt.os.ver field.
- Use the Windows-team-recommended source of truth: RtlGetVersion().dwBuildNumber for the running OS build, combined with the registry UBR (Update Build Revision) for serviced builds.
- Add Windows PAL unit tests for the formatted OS full version with UBR present, UBR absent, and a zero UBR.

Fixes #1407

## ⚠️ Behavior change for downstream telemetry consumers

**Please include in the release notes for the next tagged build.**

Before this PR, `DeviceInfo.OsBuild` / `ext.os.ver` on Windows desktop could be a `BuildLabEx`-style string (e.g. `26100.1.amd64fre.ge_release.240331-1435`). After this PR it is always a numeric `major.minor.build[.ubr]` value (e.g. `10.0.26100.4061`).

- ✅ Queries that already expected the four-dotted-numeric form will start receiving correct, servicing-aware values (this fixes #1407).
- ❌ Dashboards / Kusto queries that parsed the `BuildLabEx` form (split on `.`, substring extraction for branch / flavor / date) will need to be updated.

If you maintain downstream consumers of `DeviceInfo.OsBuild` / `ext.os.ver`, please audit those queries before the next release containing this change rolls out.

## Why this source

Confirmed with the Windows team:

> Ideally use the API, and it is always safe to combine with UBR. The API and the registry should always be in sync. The reason the registry can be used is for offline reading.

So we prefer `RtlGetVersion()` (kernel-authoritative for the running OS) and add `UBR` from `HKLM\\SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion\\UBR`. The registry `CurrentBuildNumber` / `CurrentBuild` precedence chain is not needed here — those are useful for offline-reading another Windows installation's hive, not the live running OS.

## Validation
- `msbuild Solutions\MSTelemetrySDK.sln /t:Tests\UnitTests /p:Configuration=Release /p:Platform=x64 /p:PlatformToolset=v145`
- `UnitTests.exe --gtest_filter=PalTests.*` → 10 tests passed, including the three new `WindowsOsFullVersion*` cases
- `UnitTests.exe` (full run) → 463 tests passed


tests/unittests/PalTests.cpp
```diff
@@ -95,6 +109,18 @@ TEST_F(PalTests, FormatUtcTimestampMsAsISO8601)
     EXPECT_THAT(PAL::formatUtcTimestampMsAsISO8601(2147483647999ll), Eq("2038-01-19T03:14:07.999Z"));
 }
 
+#if defined(_WIN32) || defined(_WIN64)
+TEST_F(PalTests, WindowsOsFullVersionUsesServicingBuildValues)
```
ThomsonTan: Thanks for the fix. I think we still need deterministic tests for build source precedence/fallback (CurrentBuildNumber, CurrentBuild, rtl build, UBR present/missing). Current test only validates formatting, so key/fallback regressions could slip through. Can we add coverage before merge?

bmehta001: Thanks, added some more tests

bmehta001:
@GovM-MSFT from the Windows org is familiar with versioning and his team was the one that reported the issue, so I requested him to review the PR. He could not access the PR here for some reason, but was able to review a mirror I created on my fork with his personal GitHub account and has approved it.

lib/pal/desktop/WindowsDesktopSystemInformationImpl.cpp
```diff
@@ -139,17 +140,95 @@ namespace PAL_NS_BEGIN {
...
+        if (hasUpdateBuildRevision)
+        {
+            version += "." + std::to_string(updateBuildRevision);
+        }
+
+        return version;
+    }
+
+    std::string getWindowsOsFullVersionFromSources(
+        unsigned long majorVersion,
```
lalitb: One question, and I’m not a Windows expert, so please correct me if I’m missing something: when `RtlGetVersion()` succeeds, should we use `rtlOsvi.dwBuildNumber` as the base build and only read `UBR` from the registry?

This PR currently prefers registry `CurrentBuildNumber` / `CurrentBuild` over the RTL build. Those should normally match, but if they ever differ, using RTL as the source of truth for the running OS build seems safer to me.

lalitb:
One compatibility note: this changes the shape of the Windows value that flows into `DeviceInfo.OsBuild` / `ext.os.ver`. Previously this could be a `BuildLabEx`-style string, and after this change it becomes a numeric `major.minor.build[.ubr]` value. We should call it out clearly in the PR description and also in release notes so downstream telemetry consumers are aware of the field format change.


Apart from that, changes look good to me - but would let @ThomsonTan approve this, as he knows this part of code better.

bmehta001:
Updated this in commit 78e6db48. Confirmed with the Windows team:

> Ideally use the API, and it is always safe to combine with UBR. The API and the registry should always be in sync. The reason the registry can be used is for offline reading.

So this now uses `RtlGetVersion().dwBuildNumber` as the build, with `UBR` from registry. The registry `CurrentBuildNumber` / `CurrentBuild` precedence chain (and its helper) is removed entirely.

lib/pal/desktop/WindowsDesktopSystemInformationImpl.cpp
bmehta001: From Gov: 

> Ideally use the api, and it is always safe to combine with UBR. The api and the registry should always be in sync. The reason the registry can be used is for offline reading

I am changing accordingly, so this PR now uses RtlGetVersion().dwBuildNumber as the build, with UBR from registry. The registry CurrentBuildNumber / CurrentBuild precedence chain (and its helper) is removed entirely.

bmehta001:
Modified the PR body

Spelled out both sides explicitly:
- queries that already expected `major.minor.build[.ubr]` start getting correct servicing-aware values (fixes #1407)
- queries that parsed `BuildLabEx` (split on `.`, branch/flavor/date substrings) will need updates

baijumeswani (approved):

PR #1442, opened 2026-05-19
Normalize vendored zlib to upstream 1.2.11

bmehta001:
## Summary
- Remove the old SIMD/motley zlib patch files and restore the core zlib implementation files to upstream v1.2.11.
- Update `cgmanifest.json` to reference the upstream zlib v1.2.11 commit.
- Mark test/debug-only third-party registrations as development dependencies and update the OSS component docs to no longer claim Intel SIMD zlib optimizations.

This is intended as a follow-up into #1439 / `user/pablo/cg` so the CGManifest describes a standard zlib snapshot instead of the old patched tree.

## Validation
- `MSBuild.exe Solutions\MSTelemetrySDK.sln /target:zlib:Rebuild /p:BuildProjectReferences=true /p:Configuration=Release /p:Platform=x64 /p:PlatformToolset=v145 /verbosity:minimal`
- `MSBuild.exe Solutions\MSTelemetrySDK.sln /target:sqlite:Rebuild,zlib:Rebuild,Tests\gmock:Rebuild,Tests\gtest:Rebuild,Tests\UnitTests:Rebuild /p:BuildProjectReferences=true /p:Configuration=Release /p:Platform=x64 /p:PlatformToolset=v145 /verbosity:minimal`
- `UnitTests.exe --gtest_color=no` (460 passed)
- `MSBuild.exe Solutions\MSTelemetrySDK.sln /target:Tests\FuncTests:Rebuild /p:BuildProjectReferences=true /p:Configuration=Release /p:Platform=x64 /p:PlatformToolset=v145 /verbosity:minimal`
- `FuncTests.exe --gtest_color=no` (45 passed)
- WSL direct CMake configure/build with `-DLINK_STATIC_DEPENDS=OFF -DBUILD_SHARED_LIBS=OFF`
- WSL `UnitTests` (496 passed)
- WSL `FuncTests --gtest_filter=-APITest.C_API_Test` (38 passed)


pablo-msft (dismissed):

pablo-msft (approved):

third_party/cgmanifest.json
```diff
@@ -0,0 +1,35 @@
...
+    },
+    {
+      "Component": {
+        "Type": "git",
+        "git": {
+          "RepositoryUrl": "https://github.com/clibs/sha1/",
+          "CommitHash": "d5b29e11ec4871b78b370e8e00bd0cd56d798dda"
+        }
+      },
+      "DevelopmentDependency": true
```
lalitb, line 32: I believe `sha1` is  a production dependent. Good to validate if `DevelopmentDependency` should be `false` here.

lalitb (approved):
Nice work. Thanks for the cleanup. As a follow-up, it may be worth considering whether the Windows build can move to zlib from vcpkg instead of continuing to carry a checked-in snapshot, so it is easier to stay current in the future.

third_party/cgmanifest.json
bmehta001: From my understanding, the third_party/sha1 snapshot is only used by the tracing/debug API shared-items project (`lib/tracing/api/tracing-api.vcxitems`), and from the OSS docs, I understood that this SHA-1 copy is not included in shippable SDK bits and is for internal developer debug builds only. 

As a result, I used `DevelopmentDependency: true`, but not sure if Component Governance needs to have `DevelopmentDependency: false` for any checked-in source regardless of whether it ships. I can change it back to `false` in a separate PR, if that would be better.

bmehta001:
Yes, will use zlib via vcpkg in my vcpkg PR

PR #1443, opened 2026-05-20
Update vendored zlib to 1.3.2

bmehta001:
## Summary

- Refresh vendored zlib sources, headers, docs, and minizip files to upstream v1.3.2 (`da607da739fa6047df13e66a2af6b8bec7c2a498`).
- Update CGManifest to point at the upstream v1.3.2 commit
- Keep `zlib/` closer to the upstream layout while preserving the repo-specific integrations that are still required:  
  - Move the repo-owned Visual Studio zlib project wiring out of `zlib/contrib/vstudio/vc14` and into `third_party/Solutions/zlib/vc14`, then update the solution, SDK project, sample, and test references to the new location. Upstream zlib 1.3.2 removed the pre-made Visual Studio project directories, but this repo still builds zlib through the legacy `zlibvc.vcxproj` path from `Solutions/MSTelemetrySDK.sln`, SDK projects, samples, and tests.  
  - Remove the unused legacy `zlib/contrib/vstudio/vc9`, `vc10`, and `vc11` directories from the vendored zlib tree.  
  - Keep the SDK's `act_z_` symbol-prefix shim through `zlib/names.h` and the `#include "names.h"` hooks in `zconf.h`, `zconf.h.in`, and `zconf.h.cmakein`. This still needs to be visible from zlib's public/config headers while the repo uses a prefixed vendored zlib.  
  - Keep the repo-specific `gzgetc` prefix handling in `gzread.c` / `zlib.h`; this remains tied to the custom `act_z_` prefix map because using upstream's `Z_PREFIX_SET` path directly would break zlib's `deflateInit` / `inflateInit` macro behavior for this repo's prefix scheme.
  - Fix the zlib CMake shim so out-of-source builds do not rename `zconf.h` and non-Windows builds do not get Windows-only macros.


pablo-msft:
> Preserve repo-specific zlib integration instead of doing a pure upstream folder replacement:
> - Keep the existing zlib/contrib/vstudio/vc9, vc10, vc11, and vc14 project directories. Upstream zlib 1.3.2 removed these, but this repo still builds zlib through zlib/contrib/vstudio/vc14/zlibvc.vcxproj from Solutions/MSTelemetrySDK.sln, SDK projects, samples, and tests.
> - Keep the SDK's act_z_ symbol-prefix shim through zlib/names.h and the #include "names.h" hooks in zconf.h, zconf.h.in, and zconf.h.cmakein.
> - Keep the repo-specific gzgetc prefix handling in gzread.c / zlib.h.

Is it possible to move some (or all) of these to somewhere outside of zlib? For example, the vcxproj could move to third_party/Solutions/zlib (lining up with my hope of moving zlib itself to third_party/zlib) -- that would make it clear what the changes are. Very similar to the PR you just did regarding resetting zlib to a clean copy.

bmehta001:
Done in `7f2331fc`. I moved the repo-owned Visual Studio zlib project wiring out of `zlib/contrib/vstudio/vc14` to `third_party/Solutions/zlib/vc14`, updated the solution/project references, and removed the unused `vc9`, `vc10`, and `vc11` project directories from the vendored zlib tree.

I also checked whether the prefix shim could move out cleanly. I kept `names.h` / the small `gzgetc` handling in `zlib` for now because the `act_z_` map has to be visible through zlib's config/public headers, and treating it as upstream `Z_PREFIX_SET` would break zlib's `deflateInit` / `inflateInit` macro behavior for this repo's prefix scheme.

pablo-msft (approved):

third_party/Solutions/zlib/vc14/zlibvc.vcxproj
```diff
@@ -1170,46 +1170,46 @@ bld_ml64.bat</Command>
...
-    <ClInclude Include="..\..\..\infblock.h" />
-    <ClInclude Include="..\..\..\infcodes.h" />
-    <ClInclude Include="..\..\..\inffast.h" />
-    <ClInclude Include="..\..\..\inftrees.h" />
-    <ClInclude Include="..\..\..\infutil.h" />
-    <ClInclude Include="..\..\..\zconf.h" />
-    <ClInclude Include="..\..\..\zlib.h" />
-    <ClInclude Include="..\..\..\zutil.h" />
+    <ClInclude Include="..\..\..\..\zlib\deflate.h" />
+    <ClInclude Include="..\..\..\..\zlib\infblock.h" />
```
lalitb, line 1202: Non-blocking: these headers `infblock.h` / `infcodes.h` / `infutil.h` are stale entries. These already existed before this move, so I do not think this is a regression. They should only matter if something actually tries to include/use those headers. Longer term, as we move away from carrying a repo-owned zlib snapshot, we should clean up this kind of stale project wiring too. No action needed as of now, just flagging it for now.

lalitb (approved):
LGTM. Nice work.

One observation - I noticed our internal `vcpkg` port may drift from this `zlib` update because it has its own pinned source and [patch](  https://github.com/microsoft/cpp_client_telemetry/blob/7f2331fca441968609d83ebecd19ecfe009b317a/tools/ports/mstelemetry/portfile.cmake#L34-L40
). I think the vcpkg port is already stale in lots of areas, so worth keeping in mind for later cleanup if we continue to carry that port.

third_party/Solutions/zlib/vc14/zlibvc.vcxproj
bmehta001: Thanks - addressed all the stale headers/filters, etc. that I could find in https://github.com/microsoft/cpp_client_telemetry/pull/1443, which Baiju approved for me.

PR #1445, opened 2026-05-22
Auto-enable Pages on first deploy

bmehta001:
## Summary
- Pass `enablement: true` to `actions/configure-pages@v5` so the Deploy documentation to GitHub Pages workflow provisions the Pages site on first run.

## Why
The post-merge deploy job has been failing on every push to `main` with::

    Error: Get Pages site failed. Please verify that the repository has Pages
    enabled and configured to build using GitHub Actions, or consider exploring
    the `enablement` parameter for this action.
    Error: Not Found - https://docs.github.com/rest/pages/pages#get-a-apiname-pages-site

`gh api repos/microsoft/cpp_client_telemetry/pages` returns 404, which means GitHub Pages has never been provisioned for this repository. `actions/configure-pages` accepts an `enablement` parameter that asks the action to create the Pages site on first run; the workflow already grants the `pages: write` and `id-token: write` permissions it needs to do so.

## Fallback
If the org-level `GITHUB_TOKEN` restrictions still prevent automatic provisioning, a repository admin only needs to flip **Settings -> Pages -> Build and deployment -> Source** to `GitHub Actions` once; subsequent deploys will then succeed without the failure.

## Validation
- YAML parsed cleanly with `python -c "import yaml; yaml.safe_load(open(...))"`
- Diff is a two-line addition under the existing `actions/configure-pages@v5` step

### DIFF
--- a/.github/workflows/deploy-docs-pages.yml
+++ b/.github/workflows/deploy-docs-pages.yml
@@ -72,6 +72,8 @@ jobs:
     steps:
       - name: Configure GitHub Pages
         uses: actions/configure-pages@v5
+        with:
+          enablement: true
 
       - name: Deploy to GitHub Pages
         id: deployment


.github/workflows/deploy-docs-pages.yml
```diff
@@ -72,6 +72,8 @@ jobs:
     steps:
       - name: Configure GitHub Pages
         uses: actions/configure-pages@v5
+        with:
+          enablement: true
```
lalitb, line 76: Are you sure this will work? As per the Github [docs](https://github.com/actions/configure-pages/blob/983d7736d9b0ae728b81ab479565c72886d7745b/action.yml#L19) - this option requires a token other than `GITHUB_TOKEN` to be provided.

bmehta001: You are correct — I misread the docs. `actions/configure-pages` requires `administration:write` (PAT or GitHub App) to provision the Pages site; with only `GITHUB_TOKEN`, `enablement: true` will still 403. So this two-line diff is effectively a no-op.

Two real paths:
1. **Close this PR** and have an admin flip **Settings → Pages → Source: GitHub Actions** once. One-time, no secret to maintain, cleanest.
2. **Add a `PAGES_DEPLOY_TOKEN` repo secret** with admin:write and pass it via `with.token` here.

I'd lean (1) unless we expect this workflow to be cloned to new repos. Which would you prefer? Happy to push (2) or close in favor of (1).

bmehta001: Sorry, CoPilot keeps leaving comments for me automatically, and I need to figure out how to stop it - please ignore the previous comment.

bmehta001: Ok, thanks, you were right. I was trying different things without understanding what the real solution was, but since you made me an admin yesterday, I checked the status of Pages deployment in Settings and realized it was stuck on legacy instead of GitHub Actions. I changed it to use GitHub Actions now, so a website is now deploying. Closing this PR.

PR #1448, opened 2026-05-25
Reduce C++ CodeQL build scope

bmehta001:
Reduce the C++ CodeQL extraction size so the standard Windows runner is less likely to hit query-analysis OOM.

### Changes

- replace the C++ CodeQL `build-all.bat` invocation with a single representative `x64 Release` MSBuild call
- keep the native desktop and UWP library targets plus their SQLite/zlib dependencies covered
- stop building duplicate Debug/Release, Win32/x64, MD/MT, test, sample, and C#-oriented targets during C++ CodeQL extraction

### Notes

This intentionally does **not** cap CodeQL analysis threads. The first-line fix is reducing the database size; thread capping can still be added later if the smaller extraction is not enough.


.github/workflows/codeql-analysis.yml
```diff
@@ -56,14 +56,15 @@ jobs:
...
         PlatformToolset: v143
         VSTOOLS_VERSION: vs2022
         WindowsSDKVersion: 10.0.22621.0
       shell: cmd
       if: matrix.language == 'cpp'
-      run: build-all.bat
+      run: |
+        call tools\gen-version.cmd
+        git -c submodule."lib/modules".update=none submodule update --init --recursive --depth=1
+        call tools\RunMsBuild.bat x64 Release "sqlite:Rebuild,zlib:Rebuild,sqlite-uwp:Rebuild,win32-lib:Rebuild,win10-lib:Rebuild"
```
lalitb: nit - it may be cleaner to put the reduced CodeQL C++ build steps in a small script, for example a CodeQL-specific .bat file, instead of keeping the target list directly in the workflow.

lalitb (approved):

.github/workflows/codeql-analysis.yml
bmehta001: Thanks - have split it into new file

baijumeswani (approved):

PR #1452, opened 2026-05-27
Added EVT_DROPPED reason codes via param2

mogiligarimidi23:
**Summary**

Make SDK-level event drops observable to DebugEventListeners by attaching a reason code to existing EVT_DROPPED dispatches via DebugEvent.param2, dispatching EVT_DROPPED from the previously-silent teardown-timeout path, anddispatching EVT_DROPPED from the catastrophic-resize branch of the SQLite offline store. Two new values are appended to EventDroppedReason for cases the existing enum doesn't cover.

The change is purely additive — no existing dispatch semantics are altered, and param2 is 0 for any caller that hasn't been updated.

**Motivation**

Several EVT_DROPPED sites today emit only param1 = recordCount, leaving listeners unable to distinguish why records were dropped (storage trim vs. profile-disabled vs. teardown-timeout vs. storage-save-failed).Application-layer telemetry-health dashboards can therefore report "events dropped" but not the cause. This change carries the existing internal attribution out to listeners.

**Changes**

lib/include/public/Enums.hpp — append two values to EventDroppedReason:

 DROPPED_REASON_TEARDOWN_TIMEOUT,
 DROPPED_REASON_LATENCY_DISABLED_BY_PROFILE,

All existing values keep their numeric assignments; the new values are added at the end.

lib/include/public/IOfflineStorage.hpp — add non-pure virtual GetRemainingRecordCountForShutdown() const with a default that returns GetRecordCount().
ABI note: third-party implementations stay source-compatible but must be recompiled against this SDK version (the vtable grows). Documented in the method's <remarks>.

lib/offline/OfflineStorage_SQLite.cpp — the catastrophic-resize branch (m_DbSizeEstimate > 2 × m_DbSizeLimit, which deletes all events) now dispatches EVT_DROPPED with param1 = count and param2 = DROPPED_REASON_OFFLINE_STORAGE_OVERFLOW. Previously this branch returned silently. The trim branch (≤ 2× limit) is unchanged.

Teardown-timeout dispatch — EVT_DROPPED now fires with param2 = DROPPED_REASON_TEARDOWN_TIMEOUT when records remain after the shutdown deadline.

Profile-disabled dispatch — EVT_DROPPED now fires with param2 = DROPPED_REASON_LATENCY_DISABLED_BY_PROFILE when records are dropped because their latency tier is disabled by the active transmit profile.

**Tests**

No new tests; existing offline-storage and telemetry-system tests cover the modified dispatch paths. Verified on Android


lib/offline/StorageObserver.cpp
```diff
@@ -180,6 +181,7 @@ namespace MAT_NS_BEGIN {
             DebugEvent evt;
             evt.type = EVT_DROPPED;
             evt.param1 = overallCount;
+            evt.param2 = static_cast<size_t>(DROPPED_REASON_OFFLINE_STORAGE_OVERFLOW);
```
lalitb: I think the reason here should be for the records deleted for exceeding the max retry count, not storage overflow.

lib/include/public/Enums.hpp
```diff
@@ -565,7 +565,9 @@ namespace MAT_NS_BEGIN
         DROPPED_REASON_SERVER_DECLINED_5XX,
         DROPPED_REASON_SERVER_DECLINED_OTHER,
         DROPPED_REASON_RETRY_EXCEEDED,
-        DROPPED_REASON_COUNT
+        DROPPED_REASON_COUNT,
+        DROPPED_REASON_TEARDOWN_TIMEOUT,
+        DROPPED_REASON_LATENCY_DISABLED_BY_PROFILE
```
lalitb: Can we move the new values before `DROPPED_REASON_COUNT` so that the  variable `gc_NumDroppedReasons` defined below (line 573) remains valid, and also in case any listeners validates using expetression ` param2 < DROPPED_REASON_COUNT`.

lib/system/TelemetrySystem.cpp
```diff
@@ -76,7 +76,17 @@ namespace MAT_NS_BEGIN {
                     auto uploadTime = GetUptimeMs() - stopTimes[0];
                     if (uploadTime >= (1000L * timeoutInSec))
                     {
-                        // Hard-stop if it takes longer than planned
+                        // Hard-stop if it takes longer than planned.
+                        // Emit an EVT_DROPPED carrying the remaining record count
+                        // so listeners can attribute teardown-timeout losses.
+                        const size_t remaining = storage.GetRecordCount();
```
lalitb: Can we also include the in-flight upload count here? `GetRecordCount()` can be 0 during teardown even when records are still being uploaded, so this may miss some dropped records.

lalitb:
Can we double-check other delete/drop paths too? For example, the SQLite “DB is too big” path does `DELETE FROM` events and returns before emitting `EVT_DROPPED`, so some drops may still have no reason code.

lib/offline/StorageObserver.cpp
mogiligarimidi23: Updated this with DROPPED_REASON_RETRY_EXCEEDED

lib/system/TelemetrySystem.cpp
mogiligarimidi23: Done

lib/include/public/Version.hpp
```diff
@@ -6,8 +6,8 @@
 #define MAT_VERSION_HPP
 // WARNING: DO NOT MODIFY THIS FILE!
```
bmehta001, line 7: You can revert the changes to this file, since it is automatically generated

lib/system/TelemetrySystem.cpp
```diff
@@ -76,7 +76,19 @@ namespace MAT_NS_BEGIN {
                     auto uploadTime = GetUptimeMs() - stopTimes[0];
                     if (uploadTime >= (1000L * timeoutInSec))
                     {
-                        // Hard-stop if it takes longer than planned
+                        // Hard-stop if it takes longer than planned.
+                        // Emit an EVT_DROPPED carrying the remaining record count
+                        // so listeners can attribute teardown-timeout losses.
+                        // Include in-flight upload records too: GetRecordCount() can be 0
+                        // during teardown while records are still being uploaded.
+                        const size_t remaining = storage.GetRecordCount() + tpm.inflightRecordCount();
```
lalitb: Thanks for adding this. I realized the count works differently by storage type: memory storage does not include in-flight records in `GetRecordCount()`, but SQLite/Room already do. So adding `inflightRecordCount()` here can double-count for SQLite/Room.

Maybe we can move this behind the storage layer instead, e.g.:

```cpp
size_t GetRemainingRecordCountForShutdown();

// MemoryStorage
return GetRecordCount() + GetReservedCount();

// SQLite / Room
return GetRecordCount();
```

Then TelemetrySystem can just call that one method without knowing storage-specific details.

lib/include/public/Version.hpp
lalitb: +1. The version bump is done with new release, not with every PR.

lib/offline/OfflineStorage_SQLite.cpp
```diff
@@ -954,24 +954,27 @@ namespace MAT_NS_BEGIN {
                 LOG_TRACE("DB is too big, deleting...");
                 Execute("DELETE FROM " TABLE_NAME_EVENTS);
                 Execute("VACUUM");
-                return true;
+                eventsDropped = count;
             }
```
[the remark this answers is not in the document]
mogiligarimidi23: dispatching EVT_DROPPED is intentional. That path was previously silent, which meant catastrophic resizes (DB > 2× limit) produced zero drop telemetry even though every event in the DB was lost. Surfacing those drops is one of the goals of this PR, not an accidental side effect. I'll update the PR description to call this outexplicitly so it isn't framed as a purely additive change.

lib/include/public/IOfflineStorage.hpp
```diff
@@ -329,6 +329,18 @@ namespace MAT_NS_BEGIN {
...
+        /// Returns the total number of records that should be considered as
+        /// "still pending" at shutdown for teardown-timeout reporting.
+        /// </summary>
+        /// <remarks>
+        /// Default implementation returns GetRecordCount(), which is correct for
+        /// persistent storages (e.g. SQLite, Room) where in-flight records are
+        /// still counted in the on-disk total. Volatile storages (e.g. memory)
+        /// must override to also include their in-flight / reserved records.
+        /// </remarks>
+        virtual size_t GetRemainingRecordCountForShutdown() { return GetRecordCount(); }
```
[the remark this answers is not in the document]
mogiligarimidi23: done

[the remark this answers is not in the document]
mogiligarimidi23: done

lib/offline/MemoryStorage.hpp
```diff
@@ -67,7 +67,9 @@ namespace MAT_NS_BEGIN {
 
         virtual size_t GetRecordCount(EventLatency latency = EventLatency_Unspecified) const override;
 
-        virtual size_t GetReservedCount();
+        virtual size_t GetRemainingRecordCountForShutdown() const override;
+
+        virtual size_t GetReservedCount() const;
```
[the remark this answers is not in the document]
mogiligarimidi23: done

lib/offline/MemoryStorage.cpp
```diff
@@ -514,11 +514,20 @@ namespace MAT_NS_BEGIN {
...
     }
 
+    /// <summary>
+    /// Memory storage does not include in-flight (reserved) records in
+    /// GetRecordCount(), so add them here for accurate shutdown reporting.
+    /// </summary>
+    size_t MemoryStorage::GetRemainingRecordCountForShutdown() const
+    {
+        return GetRecordCount() + GetReservedCount();
+    }
```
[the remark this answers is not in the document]
mogiligarimidi23: done

lalitb (approved):
LGTM. One note: this change is not ABI-compatible for consumers/applications that bring their own `IOfflineStorage` implementation, because the interface `vtable` changes. We should explicitly call this out during release: those consumers may not need code changes, but they need to recompile against the new SDK.

PR #1453, opened 2026-05-27
Remove ATL

pablo-msft:
Eliminates forced dependency on ATL for a single function call that can trivially be inlined


pablo-msft:
Very likely we can delete refs to ATL and ATLMFC and so on; however, I am unprepared to take on the scope of that work at this time

bmehta001 (approved):

PR #1457, opened 2026-05-29
sqlite: upgrade vendored snapshot 3.34.1 -> 3.53.1 (CVE fixes)

### LINKED ISSUE #1313: SqlLite Project fails to build due to contradictory command line arguments
Building on Windows 11 with VS 2022 17.12.1 for UWP on x64/ARM64

**Steps to reproduce.**
- Invoke `.\build-win.ps1 release x64,arm64` after cloning the repo

Successful compilation

Compiler reports:
 `error D8016: '/O1' and '/RTC1' command-line options are incompatible [C:\Code\Git\SurfaceOneDS\cpp_client_telemetry\sqlite\sqlite.vcxproj]`

**Additional context.**
On the one hand, it seems like I'm doing something wrong because there haven't been any changes to these build settings in years.  On the other hand, the options **are** contradictory. `O1` means "minimize size" and `RTC1` means "add extra run time checks", and you can't really do both of those at the same time.  So it could be that more recent versions of Visual Studio are just stricter about requiring the users to be clear about what they want, but I don't think that's true either.  I'm not sure how this ever compiled... Which brings me back to I'm missing something...

If it is an issue that needs to be addressed, the fix is very simple and I'm happy to submit a PR but want to make sure it isn't user-error on my part first.

bmehta001:
# SQLite upgrade: 3.34.1 -> 3.53.1

Addresses [`cpp_client_telemetry_modules#248`](https://github.com/microsoft/cpp_client_telemetry_modules/issues/248) and its duplicate [`cpp_client_telemetry_modules#266`](https://github.com/microsoft/cpp_client_telemetry_modules/issues/266).

Also fixes #1313 by aligning the SQLite release-MT MSVC runtime-check setting with the existing Release configs.

## Why

The in-tree SQLite snapshot has been on **3.34.1** (March 2021) since commit `03c78ece`. That leaves us exposed to a series of public CVEs against SQLite:

| CVE | Description | Fixed in |
|---|---|---|
| CVE-2022-35737 | String overflow in printf APIs | 3.39.2 |
| CVE-2023-7104 | Heap buffer overflow in `sqlite3session.c` | 3.43.1 |
| CVE-2024-0232 | Use-after-free in JSON parser | 3.43.2 |
| CVE-2025-6965 | Aggregate function memory corruption | 3.50.2 |
| CVE-2025-70873 | zipfile extension info disclosure | post-3.51.1 |
| CVE-2025-7709 | FTS5 integer overflow -> controlled OOB write | 3.50.3 |

Upgrading directly to the **current stable amalgamation (3.53.1)** clears all of the above and avoids landing on an interim version that would require another CVE-bump round trip later.

## What's in the diff

| File | Change |
|---|---|
| `sqlite/sqlite3.h` | Refreshed from upstream 3.53.1 amalgamation |
| `sqlite/sqlite3ext.h` | Refreshed from upstream 3.53.1 amalgamation |
| sqlite/sqlite.vcxproj | Removes release-MT /RTC1 runtime checks that conflict with /O1 on modern MSVC toolsets |
| `sqlite/shell.c` | Refreshed from upstream 3.53.1 amalgamation |
| `sqlite/sqlite3_debug.c` | Refreshed; MEMSET_S patches re-applied |
| `sqlite/sqlite3_retail.c` | Refreshed; MEMSET_S patches re-applied |
| `sqlite/version.txt` | New source URL + rewritten integration steps |
| `cgmanifest.json` | SQLite `CommitHash` bumped to `version-3.53.1` tag (`ccd445d...`) |

Source verification:

- URL: <https://www.sqlite.org/2026/sqlite-amalgamation-3530100.zip>
- SHA3-256: `3c07136e4f6b5dd0c395be86455014039597bc65b6851f7111e88f71b6e06114`
- GitHub mirror tag: [`version-3.53.1`](https://github.com/sqlite/sqlite/releases/tag/version-3.53.1) (`ccd445d76a9362c63add000354fac84ba9022176`)

## MEMSET_S patch reapplication (from PR #1286)

The optional `USE_ONEDS_SECURE_MEM_FUNCTIONS` define wraps a fixed set of bulk-clearing `memset` calls inside SQLite with the bounds-checking `MEMSET_S` macro. Four of the original five patch sites still exist verbatim in 3.53.1:

1. `pcacheFetchFinishWithInit` - clearing `&pPgHdr->pDirty`
2. `sqlite3VdbeCreate` - clearing `&p->aOp`
3. `whereClauseInsert` - clearing `&pTerm->eOperator`
4. `sqlite3WhereBegin` - clearing `&pWInfo->nOBSat`

The **fifth** site (`sqlite3ExprListAppend` - clearing `&pItem->zEName`) was refactored upstream into a single `*pItem = zeroItem;` struct assignment, so it no longer needs the wrapper.

The CMake option, `memset_s.h` header, and the wrapper `sqlite/sqlite3.c` are all unchanged.

## Trade-off: debug `#line` directives

The previous `sqlite3_debug.c` was generated with the SQLite build tool's `--linemacros` flag and carried **802 `#line` directives** mapping back to the unamalgamated `tsrc/*.c` files; `sqlite3_retail.c` had them stripped.

The upstream amalgamation download has **no `#line` directives** by default (matching the previous retail flavor). Regenerating the `--linemacros` variant requires Tcl plus the full source tree, which we don't have a local pipeline for.

For this upgrade both files are byte-identical copies of the upstream amalgamation. The retail/debug **file split** (and the `sqlite/sqlite3.c` wrapper that selects between them via `NDEBUG`) is preserved so the `.vcxproj` / build files don't need any changes; only the source-file-mapping convenience in **debug** SQLite-internal stack traces is lost, and that capability was never tested for the retail flavor.

A follow-up cleanup (matching the eventual direction in the internal `mip/3.9.324.1/custom` branch) could collapse the two files back into a single `sqlite3.c`; intentionally kept out of scope here to minimize the project-file blast radius.

## Out of scope

- The Android build (per @anod's earlier comment in modules#248) uses `androidx.Room` by default and does not link the vendored SQLite, so it is unaffected.
- Linux / macOS consumers that use the platform SQLite (e.g., via vcpkg or system package) are likewise unaffected by this snapshot bump.

## Reproducibility

```powershell
# Anyone can re-derive sqlite3_{debug,retail}.c from this commit:
Invoke-WebRequest https://www.sqlite.org/2026/sqlite-amalgamation-3530100.zip -OutFile a.zip
Expand-Archive a.zip

# Apply the four MEMSET_S substitutions to a.\sqlite-amalgamation-3530100\sqlite3.c
# (matching the wrap pattern from PR #1286)
# Copy result to both sqlite/sqlite3_debug.c and sqlite/sqlite3_retail.c.
```

## Pre-merge checklist for reviewers

- [ ] CI: all Windows/Linux/macOS build matrix entries pass
- [ ] CI: `OfflineStorage_SQLite` unit tests pass (this is the main SDK consumer of the vendored SQLite)
- [ ] CI: `Analyze cpp` (clang static analysis) is clean
- [ ] Spot-check: confirm `MEMSET_S(...)` appears exactly 4x in each of `sqlite3_debug.c` and `sqlite3_retail.c`
- [ ] Spot-check: confirm `SQLITE_VERSION "3.53.1"` in `sqlite3.h`

Closes microsoft/cpp_client_telemetry_modules#248
Closes microsoft/cpp_client_telemetry_modules#266


lalitb (approved):
Thanks for adding this.

PR #1460, opened 2026-06-03
Fix cancelAllRequests drain race and spin

bmehta001:
## Summary
- extracts the `HttpClientManager` cancel/drain fix from #1429 into a small PR
- locks `m_httpCallbacks` reads in `cancelAllRequests()` and `requestCount()`
- replaces the tight `yield()` drain loop with a 50ms sleep to avoid high CPU while waiting for callbacks to drain
- applies the same synchronized drain check to `HttpClient_Apple::CancelAllRequests()` and removes its redundant `yield()` after the existing sleep
- removes the session-wide `getTasksWithCompletionHandler` cancel-all from each Apple request cancel; `CancelAllRequests()` already cancels each SDK-tracked request individually

## Context
This keeps shutdown drain behavior intentionally unbounded: returning before in-flight callbacks drain can leave callbacks touching HTTP client manager/request state after teardown. The companion HTTP adapter fix should make cancellation complete promptly; these changes prevent SDK-side container races and avoid burning CPU while waiting.

## Validation
- `msbuild Solutions\MSTelemetrySDK.sln /m:1 /nr:false /t:Tests\UnitTests /p:Configuration=Release /p:Platform=x64 /p:PlatformToolset=v145`
- `Solutions\out\Release\x64\UnitTests\UnitTests.exe --gtest_filter=HttpClientManagerTests.*`

### DIFF
--- a/lib/http/HttpClientManager.cpp
+++ b/lib/http/HttpClientManager.cpp
@@ -152,44 +152,9 @@ namespace MAT_NS_BEGIN {
     {
         cancelAllRequestsAsync();
 
-        // Defense-in-depth hardening for the telemetryd_v2 96 percent CPU
-        // bug. The original loop had two issues:
-        //
-        //   1. Data race (UB): m_httpCallbacks.empty() was read WITHOUT
-        //      holding m_httpCallbacksMtx, while onHttpResponse (a few
-        //      lines above) calls m_httpCallbacks.remove(callback) under
-        //      that same mutex. Concurrent access to a std::list's internal
-        //      state from two threads without synchronization is undefined
-        //      behavior; it only "worked" because empty() happens to be
-        //      cheap and rarely visibly miscompiled.
-        //
-        //   2. Liveness / CPU: std::this_thread::yield() on Darwin maps to
-        //      sched_yield / swtch_pri, which the scheduler treats as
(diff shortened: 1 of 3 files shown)


lib/http/HttpClientManager.cpp
```diff
@@ -149,8 +151,56 @@ namespace MAT_NS_BEGIN {
     void HttpClientManager::cancelAllRequests()
     {
         cancelAllRequestsAsync();
-        while (!m_httpCallbacks.empty())
-            std::this_thread::yield();
+
+        // Defense-in-depth hardening for the telemetryd_v2 96 percent CPU
+        // bug. The original loop had two issues:
```
Kr1sso: // Defense-in-depth hardening for the telemetryd_v2 96 percent CPU
        // bug.

You should remove that

Kr1sso (approved):
Great, this will unblock us downstream!

lib/http/HttpClientManager.cpp
bmehta001: Fixed in commit b222b867. Replaced the incident-specific block with a short neutral comment about synchronized callback draining and sleeping between polls.

bmehta001: Thanks, removed

lalitb (approved):

PR #1464, opened 2026-06-05
Make Curl CA path ownership safe

bmehta001:
This is split out of #1416 to keep the review surface small and focused.

Changes:
- Store Curl CA path strings with stable ownership.
- Avoid keeping pointers to transient or externally-owned string data in Curl option storage.

Validation performed locally:
- `git diff --check`


lib/http/HttpClient_Curl.hpp
```diff
@@ -434,6 +435,7 @@ class CurlHttpOperation {
     // Request values
     std::string m_method;
     std::string m_url;
+    std::string m_sslCaInfo;
     const std::map<std::string, std::string>& requestHeaders;
     const std::vector<uint8_t>& requestBody;
```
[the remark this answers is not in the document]
bmehta001: It's not necessary to store the header/body in practice given how we use this. The header is copied by libcurl and the body is sent, so there is no reason to store them

lib/http/HttpClient_Curl.hpp
```diff
@@ -140,14 +140,16 @@ class CurlHttpOperation {
...
         }
         // HTTP/2 please, fallback to HTTP/1.1 if not supported
         curl_easy_setopt(curl, CURLOPT_HTTP_VERSION, CURL_HTTP_VERSION_2_0);
 
-        // Specify our custom headers
-        for(auto &kv : this->requestHeaders)
+        // Headers are copied into m_headersChunk during construction and the
+        // curl_slist is kept alive until destruction, so the original map does
+        // not need operation-lifetime storage.
+        for(auto &kv : requestHeaders)
```
baijumeswani:
```suggestion
        for(const auto &kv : requestHeaders)
```

lib/http/HttpClient_Curl.hpp
baijumeswani: Not sure I understand why the `.c_str()` is needed here since `requestHeaders` seems to be of type `const std::map<std::string, std::string>&`

```suggestion
const std::string header = kv.first + ":" + kv.second;
```

bmehta001: Yes, you're right, it's not necessary - I will add some broader cleanups to this file as part of this PR.

baijumeswani (approved):
looks good. copilot had some suggestions. not sure if they need to be addressed.

lib/http/HttpClient_Curl.hpp
```diff
@@ -109,9 +111,9 @@ class CurlHttpOperation {
             m_callback(callback),
             m_method(method),
             m_url(url),
+            m_sslCaInfo(sslCaInfo),
 
             // Local vars
-            requestHeaders(requestHeaders),
             requestBody(requestBody)
```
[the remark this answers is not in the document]
bmehta001: In normal internal usage, or even in tests, this is not a risk but still safeguarding against it in 9e01cc7e. I removed the defaulted temporary arguments for `requestHeaders` and `requestBody`, so a caller can no longer bind a temporary to the `requestBody` reference member. Callers must now pass values that outlive the operation: production already does (`HttpClient_Curl.cpp` passes the request's own headers/body), and the unit tests now use fixture-lifetime members instead of temporaries.

lib/http/HttpClient_Curl.hpp
```diff
@@ -92,9 +94,10 @@ class CurlHttpOperation {
             std::string method,
             std::string url,
             IHttpResponseCallback* callback,
-            // Default empty headers and empty request body
-            const std::map<std::string, std::string>& requestHeaders = std::map<std::string, std::string>(),
-            const std::vector<uint8_t>& requestBody                  = std::vector<uint8_t>(),
+            // Caller-owned headers and request body; both must outlive this
+            // operation (requestBody is read by Send(), not copied).
+            const std::map<std::string, std::string>& requestHeaders,
+            const std::vector<uint8_t>& requestBody,
```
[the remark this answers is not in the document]
bmehta001: Done in 9acc16b3 - updated the constructor comment to clarify that `requestHeaders` is copied into the curl_slist during construction and need not outlive the operation; only `requestBody` is stored by reference and must outlive it.

lib/http/HttpClient_Curl.hpp
bmehta001: This is already applied - the loop uses a range-based for over requestHeaders and builds the header string without .c_str().

baijumeswani (approved):

PR #1466, opened 2026-06-05
Rearchitect vcpkg port for official submission

### LINKED ISSUE #1037: Remove unused variables in OfflineStorage_Room.cpp and HttpClient_Android.cpp
Remove unused variables in these CPP files that cause build breaks on various compilers.
See OTELCPP for Details.

bmehta001:
This replaces #1414 with a clean history based directly on current `main`. The earlier PR had accumulated too much unrelated branch history, so this PR carries only the vcpkg/CMake port layer.

## Goals
- Build the SDK as a vcpkg overlay port using CMake, not legacy scripts/MSBuild.
- Use vcpkg-provided dependencies when building under the vcpkg toolchain.
- Install CMake package config/targets for downstream consumers.
- Add consumer validation scripts for Windows, Linux, macOS, iOS, and Android.

## Scope
- CMake vcpkg dependency mode (`MATSDK_USE_VCPKG_DEPS`) and install/export support.
- Modern `tools/ports/mstelemetry` overlay port files.
- `tests/vcpkg` consumer project and platform scripts.
- Example/docs/wrapper CMake updates needed for vcpkg consumption.
- Remove the unused `tools/vcpkg` submodule; tests and CI bootstrap vcpkg from an external `VCPKG_ROOT`.

## Core SDK only (no private modules)
- The port builds the core `MSTelemetry::mat` target only.
- The optional Microsoft-proprietary modules (Privacy Guard, Sanitizer, ECS/EXP, Compliant-by-Default, AFD, CDS) in the private `lib/modules` submodule are **not** available via the vcpkg port: official vcpkg source archives pin an immutable commit and do not fetch git submodules, and that submodule is an internal-only repository.
- Module-only public headers (`CompliantByDefaultFilterApi.hpp`, `IAFDClient.hpp`, `ICdsFactory.hpp`, `IECSClient.hpp`) are excluded from the installed headers.
- The Obj-C/Swift wrapper layers are not built by the port (`BUILD_OBJC_WRAPPER`/`BUILD_SWIFT_WRAPPER=OFF`); the core SDK — including the Apple `HttpClient_Apple` HTTP stack — is still built on macOS/iOS.
- `docs/building-with-vcpkg.md` documents this scope and how to build modules from a full `--recurse-submodules` source checkout.

## HTTP client per platform
- Windows: WinInet. macOS/iOS: NSURLSession (Apple HTTP). Linux/Android: libcurl (`curl[openssl]`), declared only for `linux | android` in `vcpkg.json`.

## Android API level
- vcpkg's built-in Android triplets default to API 28. This repo's Android `minSdk` is 23, so the Android consumer test defaults to API 23 using custom overlay triplets in `tests/vcpkg/triplets` (`VCPKG_CMAKE_SYSTEM_VERSION 23`) and requires Ninja so the host IDE generator does not select a different NDK.
- The whole dependency graph must be built with the same Android API triplet; do not mix API 28-built dependencies into an API 23 consumer binary.

## Not in scope
- Public API/ABI changes.
- Runtime shutdown/cancellation fixes.
- The Curl CA path ownership cleanup (remains in #1464).

## Validation performed
- Hand-authored vcpkg/CMake files pass `git diff --check`.
- Windows vcpkg consumer test passed (`x64-windows-static`, C++11 consumer project, `vcpkg_test` including `LogManager::Initialize()` / `FlushAndTeardown()`).
- Windows vcpkg install header check: module-only headers are not installed; core headers remain.
- Linux vcpkg consumer test passed from WSL ext4 (`x64-linux`).
- Android vcpkg cross-build passed for both API 23 (custom `arm64-android-api23` overlay triplet) and API 28 (built-in `arm64-android`) using native libcurl HTTP.
- Verified the upstream archive SHA512 used by `tools/ports/mstelemetry/portfile.cmake`.
- Real send check: built ONNX Runtime for Android with `--use_telemetry` against this branch's SDK source and ran a telemetry harness on an emulator; telemetry events were emitted successfully.
- Shell syntax checks passed for the Linux, macOS, iOS, and Android vcpkg scripts.

## Follow-up after this merges
- Submit the official port to `microsoft/vcpkg` using this repo's merged commit/tag as the source (and run `vcpkg format-manifest` + `vcpkg x-add-version` in the vcpkg repo).
- Decide whether to keep or remove the in-repo overlay port once the official vcpkg port is available.
- Cut a release/tag only if the official vcpkg port should target a stable release rather than a source commit.
- Add optional vcpkg features for module support later if/when `cpp_client_telemetry_modules` can be fetched from a suitable source.
- **TODO (example CMake):** Replace the stale `#set(CMAKE_SYSTEM_PROCESSOR i386)` hint in the example `CMakeLists.txt` files. Setting `CMAKE_SYSTEM_PROCESSOR` after `project()` does not cross-compile — it only changes the multiarch lib-dir lookup and the `armv7l` branch, so uncommenting it still yields a host-arch binary. A real i386 build needs `-m32`/multilib or a toolchain file set before `project()`. Document the correct approach or drop the misleading comment.

## Backward compatibility / migration

The previous overlay port detected a local `--recurse-submodules` checkout and built the SDK from it (via `build.sh`/`install.sh` or MSBuild), which **included** the private `lib/modules` submodules. That path is removed; the port now always builds the **core** SDK from a pinned, immutable upstream archive.

- `vcpkg install --head --overlay-ports=tools/ports mstelemetry` (built from local checkout, with private modules) -> `vcpkg install mstelemetry` builds the core SDK only, from a pinned commit.
- `response_file_linux.txt` / `response_file_mac.txt` are removed; use a standard triplet.
- Module users should build from a full `--recurse-submodules` source checkout instead.

Impact assessment: nothing visible depends on the old module-overlay path -- no repo CI uses it, no public GitHub repo references it, and modern consumers (e.g. ONNX Runtime) consume the SDK via CMake FetchContent rather than the overlay port. The only potentially affected group is internal pipelines that built the SDK with private modules via the overlay. See `docs/building-with-vcpkg.md` for the full migration note.


lib/CMakeLists.txt
```diff
@@ -267,71 +302,201 @@ endif()
...
+################################################################################################
+if(MATSDK_USE_VCPKG_DEPS)
+  # vcpkg mode: all deps resolved via find_package() in root CMakeLists.txt
+  target_link_libraries(mat
+    PRIVATE
+      unofficial::sqlite3::sqlite3
+      ZLIB::ZLIB
+      nlohmann_json::nlohmann_json
+      ${LIBS}
+  )
```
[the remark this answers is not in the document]
bmehta001: Addressed in 960379ab: vcpkg dependency targets are now linked PUBLIC so static-library consumers get the transitive link set through MSTelemetry::mat.

lib/include/CMakeLists.txt
```diff
@@ -7,10 +7,26 @@ list(GET MATSDK_BUILD_VERSION_SPLIT 3 BUILD_NUMBER)
 #TODO: allow regeneration of Version.hpp using template
 #configure_file(public/Version.hpp.template "${CMAKE_CURRENT_SOURCE_DIR}/public/Version.hpp")
 
+set(MATSDK_PUBLIC_HEADER_INSTALL_EXCLUDES
+  PATTERN "public/*.template" EXCLUDE
+)
```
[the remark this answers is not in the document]
bmehta001: Addressed in e43b55df: the install rule now uses `PATTERN "*.template" EXCLUDE`, so `Version.hpp.template` is not installed.

tests/functests/CMakeLists.txt
```diff
@@ -124,13 +132,18 @@ else()
...
-  if(NOT BUILD_IOS)
-    target_link_libraries(FuncTests curl)
+  # Link curl only when the SDK actually uses the curl HTTP client (Linux, and
+  # macOS without Apple HTTP). The tests don't use curl directly, and on the
+  # Android legacy path mat uses HttpClient_Android (no system curl in the NDK).
+  # MATSDK_NEEDS_CURL already excludes iOS/Apple-HTTP; use the CURL::libcurl
+  # imported target so this is also correct under vcpkg (matches ZLIB::ZLIB above).
+  if(MATSDK_NEEDS_CURL)
+    target_link_libraries(FuncTests CURL::libcurl)
   endif()
```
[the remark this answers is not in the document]
bmehta001: Addressed in 841c6f72: the link rule now guards on `if(TARGET CURL::libcurl)` and falls back to `\` on CMake < 3.12, mirroring the guarded pattern already used in the root CMakeLists.txt. find_package(CURL REQUIRED) runs before add_subdirectory(tests), so CURL_LIBRARIES is in scope.

tests/unittests/CMakeLists.txt
```diff
@@ -189,13 +197,18 @@ else()
...
-  if(NOT BUILD_IOS)
-    target_link_libraries(UnitTests curl)
+  # Link curl only when the SDK actually uses the curl HTTP client (Linux, and
+  # macOS without Apple HTTP). The tests don't use curl directly, and on the
+  # Android legacy path mat uses HttpClient_Android (no system curl in the NDK).
+  # MATSDK_NEEDS_CURL already excludes iOS/Apple-HTTP; use the CURL::libcurl
+  # imported target so this is also correct under vcpkg (matches ZLIB::ZLIB above).
+  if(MATSDK_NEEDS_CURL)
+    target_link_libraries(UnitTests CURL::libcurl)
   endif()
```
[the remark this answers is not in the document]
bmehta001: Addressed in 841c6f72: the link rule now guards on `if(TARGET CURL::libcurl)` and falls back to `\` on CMake < 3.12, mirroring the guarded pattern already used in the root CMakeLists.txt. find_package(CURL REQUIRED) runs before add_subdirectory(tests), so CURL_LIBRARIES is in scope.

tools/ports/mstelemetry/vcpkg.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "mstelemetry",
+  "version": "3.10.100.1",
+  "port-version": 0,
```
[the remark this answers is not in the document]
bmehta001: Thanks, but I believe this is a false positive — `"version"` is a supported vcpkg manifest key and is the correct choice here.

- The [vcpkg.json reference](https://learn.microsoft.com/en-us/vcpkg/reference/vcpkg-json) lists `version` / `version-semver` / `version-date` / `version-string` together as the valid "Upstream version information" field. `"version"` is used for relaxed, dot-separated, primarily-numeric versions, which is exactly what `"3.10.100.1"` is.
- Official ports use it the same way — e.g. `microsoft/vcpkg` `ports/sqlite3/vcpkg.json` declares `"version": "3.53.2"` (same field, same numeric format).
- vcpkg populates `${VERSION}` from whichever of the four version fields is present (including `"version"`), so `-DBUILD_VERSION=${VERSION}` resolves to `3.10.100.1`, not empty.
- For numeric versions, `"version"` is actually the recommended field; `version-string` is discouraged when a relaxed/semver scheme applies, so switching would be a step backward for official submission.

Leaving the manifest as-is.

docs/building-with-vcpkg.md
```diff
@@ -1,94 +1,258 @@
...
 
-## Installing open source mstelemetry package
+The port provides the core SDK — the `MSTelemetry::mat` target and its public
+C++ headers. The optional Microsoft-proprietary modules (Privacy Guard,
+Sanitizer, ECS/EXP, Compliant-by-Default, AFD, CDS) are not part of the public
+package; to use those, build from a source checkout with submodules:
 
-The following command can be used to install the public open source release:
+```console
+git clone --recurse-submodules https://github.com/microsoft/cpp_client_telemetry
```
walbourn, line 11: Why is submodules still in use here?

examples/c/SampleC-Guest/CMakeLists.txt
```diff
@@ -1,46 +1,23 @@
-cmake_minimum_required(VERSION 3.1.0)
+cmake_minimum_required(VERSION 3.5)
```
walbourn: I agree that moving to a newer CMake is useful and important. How was "3.5"? chosen?

I would recommend using 3.21 which is the oldest version that shipped with VS 2022, but VS 2022 (17.12) is the oldest one that is still in lifecycle so you could go with 3.29.

tools/ports/mstelemetry/vcpkg.json
walbourn: If the port number is 0, the guidance is to not specify a `port-version` field.

docs/building-with-vcpkg.md
bmehta001: This is for the private submodules (Privacy Guard, Sanitizer, etc.), not dependencies like sqlite, json, curl, and zlib. 

Because the repo they live in is private, I have removed direct support for it in this new vcpkg port (users can still compile it manually), so that it can be submitted to the official vcpkg.io repo.

tools/ports/mstelemetry/vcpkg.json
bmehta001: Thanks - removed

examples/c/SampleC-Guest/CMakeLists.txt
bmehta001: 3.5 was the minimum that would allow support for newer versions of CMake (4.0+) while trying to still keep the version as low as possible to not break support for older CMake versions (since Ubuntu 20.04 ships 3.16, by default, for example), but based on what you said, 3.29 makes sense.

I can use `3.5...3.29` for now so that users using an older version of CMake don't have broken builds. Then, after the next release, I will make the update to 3.29, so that users of older CMake versions still have a vcpkg port they can use.

baijumeswani (approved):

Issue #345, opened 2020-05-06
Upgrade to more modern Visual Studio 2019 and Win 10 SDK


maxgolov:
Several fundamental issues:

- paths in build scripts currently assume vs2017

- solution files are optimized for vs2017

- we are not using the latest Win 10 SDK, e.g.

```
    <WindowsTargetPlatformVersion>10.0.17763.0</WindowsTargetPlatformVersion>
    <WindowsTargetPlatformMinVersion>10.0.17763.0</WindowsTargetPlatformMinVersion>
```
is what we use, versus
```
    <WindowsTargetPlatformVersion>10.0</WindowsTargetPlatformVersion>
    <WindowsTargetPlatformMinVersion>10.0.18362.0</WindowsTargetPlatformMinVersion>
```
What's latest.

- toolset is old'ish :
```
<PlatformToolset>v141</PlatformToolset>
```
vs
```
    <PlatformToolset>v142</PlatformToolset>
```

There are several ways to solve this:
- use conditions and specify all these 'variables' under conditions
- forcedly upgrade to vs2019+vc142+latest Win 10 SDK by default

PR #1473, opened 2026-06-10
fix: guard HAVE_MAT_LIVEEVENTINSPECTOR/PRIVACYGUARD against redefinition

bmehta001:
The modules-repo (`cpp_client_telemetry_modules`) CI workflow `build-posix-latest-exp` has been failing on Linux and macOS for ~2 weeks with:

```
lib/include/mat/config-default.h:36: error: 'HAVE_MAT_LIVEEVENTINSPECTOR' macro redefined [-Werror,-Wmacro-redefined]
lib/include/mat/config-default.h:37: error: 'HAVE_MAT_PRIVACYGUARD' macro redefined [-Werror,-Wmacro-redefined]
```

### Root cause

When `BUILD_LIVEEVENTINSPECTOR` / `BUILD_PRIVACYGUARD` (default `YES`) is on and the respective module dir exists, `tests/functests/CMakeLists.txt` and `tests/unittests/CMakeLists.txt` add `-DHAVE_MAT_LIVEEVENTINSPECTOR` / `-DHAVE_MAT_PRIVACYGUARD` to the compile command. The three `config-default*.h` headers then `#define` the same macros unconditionally. With `-Werror` (added by #1415), the `-Wmacro-redefined` warning becomes fatal on GCC/Clang.

Other macros in the same block (`HAVE_MAT_JSONHPP`, `HAVE_MAT_ZLIB`, `HAVE_MAT_STORAGE`, etc.) are not added by CMake `-D` flags, so this only affects these two.

### Fix

Wrap the two `#define`s in `#ifndef` guards in all three `config-default*.h` headers (`config-default.h`, `config-default-cs4.h`, `config-default-exp.h`). Pure-additive:
- No command-line `-D`: header defines them as before. Unchanged.
- With command-line `-D`: header skips the redefine, no warning.

### Verification

- Sister `cpp_client_telemetry_modules` open PRs (#277, #345, #349) are all blocked by this regression on Linux/macOS — even master itself is red.
- Windows is unaffected (MSVC handles the redefine differently).
- No behavior change: macros still end up defined either way.


bmehta001:
Force-pushed to extend the fix to all three `config-default*.h` headers (`config-default-cs4.h`, `config-default-exp.h` had the same unconditional defines). Same trivial 2-line guard per file. Verified no conflict with #1466.

baijumeswani (approved):

PR #1487, opened 2026-06-18
Surface per-batch record count (param2) in HTTP response debug events

mogiligarimidi23:
## Summary
 
 Populates `evt.param2` with `ctx->recordIdsAndTenantIds.size()` across
 all 5 HTTP debug-event dispatch paths in `HttpResponseDecoder.cpp`:
 
 - `EVT_HTTP_OK` (successful upload)
 - `EVT_HTTP_ERROR` (rejected by server)
 - `EVT_HTTP_FAILURE` × 3 (no response, temporary server failure,
   temporary network failure)
 
 ## Motivation
 
 Debug event listeners currently receive the HTTP status code in `param1`
 but have no way to know how many records were in the batch. This makes
 it impossible to accurately count events delivered vs. dropped per HTTP
 round-trip. The `EVT_REJECTED` path already sets `param2`; this change
 brings the other paths to parity.
 
 ## Impact
 
 - No behavioral change to existing consumers that ignore `param2`
 - Enables accurate per-batch attribution for telemetry drop tracking

### DIFF
--- a/lib/http/HttpResponseDecoder.cpp
+++ b/lib/http/HttpResponseDecoder.cpp
@@ -91,6 +91,7 @@ namespace MAT_NS_BEGIN {
                 DebugEvent evt;
                 evt.type = DebugEventType::EVT_HTTP_OK;
                 evt.param1 = response.GetStatusCode();
+                evt.param2 = ctx->recordIdsAndTenantIds.size();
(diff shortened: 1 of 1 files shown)


lalitb (approved):
Thanks,

PR #1493, opened 2026-06-22
Low-severity hardening: CSPRNG UUIDs (POSIX) and null-safe zlib error logs

bmehta001:
Two low-severity hardening fixes from a repo-wide review.

## 1. POSIX/Android UUID generation used `std::rand()` (predictable)

`PAL::generateUuidString()`'s non-Windows/non-Apple fallback built the entire 128-bit UUID from `std::rand()`, seeded once with `srand(time(0) ^ nanos)`. `std::rand()` is a weak, predictable PRNG and a time-based seed is guessable, so identifiers derived from it (`m_sessionId`, `SESSION_ID_LEGACY`, `sessionSDKUid`, Signals `InstanceId`, per-event ids) were predictable.

These are correlation identifiers (not auth material — the upload credential is the tenant token), hence Low. Fix: source the bytes from **`std::random_device`** (backed by `/dev/urandom` on Linux/Android), matching the existing `CorrelationVector.cpp` / `PseudoRandomGenerator` usage. The Windows (`CoCreateGuid`) and Apple (`CFUUIDCreate`) paths are unchanged.

**Test:** `PalTests.UuidGeneration` now also asserts a batch of **1000 generated UUIDs are all distinct** (on top of the existing length / format / entropy checks).

## 2. zlib error logs pass a possibly-NULL `msg` to `%s`

`ZlibUtils.cpp:51` and `HttpDeflateCompression.cpp:47,83` logged `zs.msg` / `stream.msg` via `%s`. zlib leaves `msg == Z_NULL` for several error codes (`Z_MEM_ERROR`, `Z_BUF_ERROR`, after a failed `deflateInit2`), and `printf("%s", NULL)` is undefined behavior — benign `(null)` on glibc but not guaranteed across the MSVC/Android/Apple CRTs this SDK targets. Guard with `msg ? msg : "(null)"`.

## Validation

Built and ran `UnitTests` on Linux (host): `PalTests` (incl. the new UUID-uniqueness check) and `ZlibUtilsTests` pass.


lib/pal/PAL.cpp
```diff
@@ -369,19 +369,17 @@ namespace PAL_NS_BEGIN {
...
-        static std::once_flag flag;
-        std::call_once(flag, [](){
-            auto now = std::chrono::high_resolution_clock::now();
-            auto nanos = std::chrono::duration_cast<std::chrono::nanoseconds>(now.time_since_epoch()).count();
-            std::srand(static_cast<unsigned int>(std::time(0) ^ nanos));
-        });
+        // Use a CSPRNG-backed source (std::random_device reads from /dev/urandom
+        // on Linux/Android) rather than std::rand()/srand(time(0)), so the session
+        // and event identifiers built from this are not predictable.
+        std::random_device rd;
```
lalitb: Since `generateUuidString()` is on the event logging path, do we have any perf data for constructing `std::random_device` on every UUID, especially on Linux/Android? The security improvement makes sense; I just want to make sure we are not adding a noticeable hot-path cost.

bmehta001: The old approach took ~115 ns, and this approach takes ~2.0 µs. I think `generateUuidString()` should be called once or twice per event.

`std::random_device` is `thread_local` , so it should only be constructed once per thread. If this is too much of a cost (17x slower), then I can close this PR.

tests/unittests/PalTests.cpp
```diff
@@ -68,6 +69,16 @@ TEST_F(PalTests, UuidGeneration)
         diff += (uuid0[i] != uuid1[i]);
     }
     EXPECT_THAT(diff, Gt(20u));
+
+    // A batch of generated UUIDs must all be distinct (guards against a stuck
+    // or low-entropy generator).
+    std::set<std::string> uuids;
+    for (size_t i = 0; i < 1000; i++) {
```
baijumeswani: Define these consts as variable names (maybe with constexpr) so it is clear what their purpose is.


```
constexpr size_t uuid_batch_size = 1000;
constexpr size_t uuid_length = 36;
```

Or more meaningful names.

baijumeswani (approved):

tests/unittests/PalTests.cpp
bmehta001: Fixed

baijumeswani (approved):

PR #1496, opened 2026-06-22
Prevent event loss when a disk write fails during Flush()

bmehta001:
## Summary

Fixes a latent **event-loss** bug in `OfflineStorageHandler::Flush()`, surfaced by Copilot's review of #1491.

`Flush()` pulled records out of the in-memory queue with `GetRecords()` — which **removes** them — *before* writing them to the disk store via `StoreRecords()`. If the disk write failed for some or all records (e.g. a SQLite write error), those records had already been dropped from memory and were never re-queued, so the events were **silently lost**; only a smaller `totalSaved` count hinted at it.

## Fix — reserve, then confirm-delete

| Step | Before | After |
|------|--------|-------|
| Pull from memory | `GetRecords()` (removes immediately) | `GetAndReserveRecords()` with a nominal lease (originals retained in the reserved set) |
| Persist | `StoreRecords()` (count only) | per-record `StoreRecord()`, tracking which ids succeeded/failed |
| Persisted records | implicitly gone | `DeleteRecords(persistedIds)` — dropped only once safely on disk |
| Failed records | **lost** | `ReleaseRecords(failedIds)` — returned to the in-memory queue, retried next flush |

The reserve lease is nominal: records are reserved and resolved synchronously under `m_flushLock`, and `reservedUntil` has no reaper in `MemoryStorage`.

## Test

`OfflineStorageHandlerFlushTests.FailedDiskWriteDuringFlushReturnsRecordsToMemory`: records the SQLite store rejects (`timestamp <= 0`, accepted by the RAM queue but rejected on disk) must still be in memory after `Flush()`.

- **Verified it fails against the old `GetRecords()`-based Flush** (`GetRecordCount() == 0` — records lost).
- Passes with this change (`GetRecordCount() == 5`).

## Validation

WSL `UnitTests`: storage + LogManager suites **82/82 pass**, including the new test.

## Context

Split out of the #1491 review as requested, keeping #1491 scoped to its original empty-filter-delete guard and synchronous-store failure propagation.

### DIFF
--- a/lib/offline/OfflineStorageHandler.cpp
+++ b/lib/offline/OfflineStorageHandler.cpp
@@ -177,28 +177,51 @@ namespace MAT_NS_BEGIN {
         size_t dbSizeBeforeFlush = m_offlineStorageMemory->GetSize();
         if ((m_offlineStorageMemory) && (dbSizeBeforeFlush > 0) && (m_offlineStorageDisk))
         {
-            // This will block on and then take a lock for the duration of this move, and
-            // StoreRecord() will then block until the move completes.
-            auto records = m_offlineStorageMemory->GetRecords(false, EventLatency_Unspecified);
-            std::vector<StorageRecordId> ids;
+            // Reserve the records in memory (with a nominal lease) instead of
+            // removing them outright. The lease duration is not time-critical
+            // here because the records are persisted and resolved synchronously
+            // under m_flushLock; reserving simply keeps the originals retrievable
+            // so any that fail to persist can be returned to the queue.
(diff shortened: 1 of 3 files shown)


bmehta001:
Folded into #1491 — the Flush() event-loss fix and the StoreRecord-failure-reporting fix are two halves of the same offline-storage data-safety change, so they're now combined in #1491 (which already contained the StoreRecord half). The Flush logic here (drain-and-re-insert-failed) and its regression test were moved verbatim into #1491.

PR #1497, opened 2026-06-23
Offline storage: batch flush inserts in one transaction

bmehta001:
## Summary

`OfflineStorage_SQLite::StoreRecords()` looped `StoreRecord()`, and each `StoreRecord()` opens its own `DbTransaction` (`BEGIN EXCLUSIVE` … `COMMIT`). So a flush of **N** records did **N** transactions = **N** WAL fsyncs.

This batches the whole vector into a **single** transaction, sharing the per-record logic via extracted private helpers (`isValidRecord`, `insertRecordUnsafe`, `checkStorageSizeLimits`). Implements the existing in-code TODO (`// consider running the batch in transaction`).

**Measured** against the SDK's own vendored sqlite (WAL, `synchronous=NORMAL`, `BEGIN EXCLUSIVE`):

| N records | per-record txn | single txn | speedup |
|-----------|----------------|-----------|---------|
| 200 | ~3.3 ms | ~0.3 ms | **~11×** |
| 1000 | ~64 ms | ~1.5 ms | **~40×** |

### Behavior notes
The batching keeps validation/reporting semantics equivalent to the old per-record path, with these intentional consequences:
- Validation runs up front (reporting rejects) **before** the transaction, so no observer callback runs while `BEGIN EXCLUSIVE` is held.
- `insertRecordUnsafe()` now checks the `execute()` result: a failed insert is **not** counted in `stored` and does not inflate `m_DbSizeEstimate`, and the write failure is reported via `OnStorageFailed("Database write failed")` **after** the transaction closes. (Previously the per-record insert result was ignored — this is a correctness improvement that overlaps the separate data-safety PR #1491; the two will need a coordinated merge.)
- The size-full notification / `ResizeDb()` check runs once after the batch (and still runs even if inserts failed, so a full DB can recover).

## Tests
- Adds `OfflineStorageTests_SQLite.StoreRecordsBatchStoresAllRecords` (asserts each record is individually retrievable).
- Full WSL `UnitTests`: **524/524 pass**.

> Note: an earlier revision also moved (instead of copied) the record on the `leaseTimeMs==0` read path; that was dropped as low-value and slightly fragile, leaving this PR focused on the transaction batching.

### DIFF
--- a/lib/offline/MemoryStorage.cpp
+++ b/lib/offline/MemoryStorage.cpp
@@ -154,20 +154,32 @@ namespace MAT_NS_BEGIN {
                 StorageRecord & record = m_records[latency].back();
 
                 size_t recordSize = record.blob.size() + sizeof(record);
-                StorageRecord forConsumer(record);
-                if (leaseTimeMs)
+
+                if (leaseTimeMs == 0)
                 {
-                    forConsumer.reservedUntil = PAL::getUtcSystemTimeMs() + leaseTimeMs;
+                    // No reservation requested (e.g. the flush-read path): the
+                    // record is consumed and removed unconditionally, so move it
+                    // straight to the consumer instead of deep-copying it first.
+                    bool wantMore = consumer(std::move(record));
+                    m_records[latency].pop_back();
+                    m_size -= std::min(m_size, recordSize);
+                    maxCount--;
+                    m_lastReadCount++;
+                    if (!wantMore) {
+                        return true;
+                    }
+                    continue;
                 }
+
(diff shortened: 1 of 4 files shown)


bmehta001:
Folded into #1491. The batched StoreRecords() optimization is combined with #1491's Flush() data-loss fix: StoreRecords() is now all-or-nothing (rolls back on any failure) and Flush() uses it, so the batch speedup and the no-loss/no-duplicate guarantee cooperate instead of conflicting. Driven to a clean Copilot review here over ~9 rounds; the final code lives in #1491.

PR #1502, opened 2026-07-08
Add Edge-specific DEPS file pinning lib/modules to c637015f

lkarra2:
Adds a top-level `DEPS` file so that gclient can resolve the `lib/modules` dependency (microsoft.cpp_client_telemetry_modules) when this SDK is consumed as a nested dependency of a larger gclient-managed source tree.

**What it does**
- Pins `lib/modules` to revision `c637015fbbe904ed556d27e3b9072f6f2a5ee401` via a `deps` entry.
- Uses `use_relative_paths = True` so the entry resolves relative to this repo's checkout location.

**Why**
Downstream consumers that vendor this SDK via gclient need an explicit `DEPS` entry to fetch `lib/modules`; without it the module sources are missing and dependent code fails to build.

**Notes**
- The pinned revision should be kept in sync with the `lib/modules` submodule pointer.

### DIFF
--- a/DEPS
+++ b/DEPS
@@ -1,0 +1,21 @@
+#
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+
+use_relative_paths = True
+
+git_dependencies = 'DEPS'
+
+vars = {
+  'edge_git': 'https://microsoft.visualstudio.com/edge/_git',
+  'edge_git_suffix': '',
+  'oneds_modules_revision': 'c637015fbbe904ed556d27e3b9072f6f2a5ee401',
+}
+
+deps = {
+  'lib/modules': {
+    'url': Var('edge_git') + '/microsoft.cpp_client_telemetry_modules' + Var('edge_git_suffix') + '@' + Var('oneds_modules_revision'),
+    'condition': 'checkout_ms_src_internal'
+  },
+}


DEPS
```diff
@@ -0,0 +1,21 @@
...
+vars = {
+  'edge_git': 'https://microsoft.visualstudio.com/edge/_git',
+  'edge_git_suffix': '',
+  'oneds_modules_revision': 'c637015fbbe904ed556d27e3b9072f6f2a5ee401',
+}
+
+deps = {
+  'lib/modules': {
+    'url': Var('edge_git') + '/microsoft.cpp_client_telemetry_modules' + Var('edge_git_suffix') + '@' + Var('oneds_modules_revision'),
+    'condition': 'checkout_ms_src_internal'
```
bmehta001, line 19: `checkout_ms_src_internal` is referenced here but is not defined in the `vars` block above. I traced this through depot_tools `gclient_eval.py::EvaluateCondition`: an undefined name in a condition is **not** a `NameError` (contrary to the earlier automated comment) — the code *"implicitly converts unrecognized names to strings"*, yielding the non-empty (truthy) string `'checkout_ms_src_internal'`, and `gclient.py::_should_process()` returns that value directly, so the dep is **processed**.

Net effect: any gclient consumer that does not define this var will try to clone `lib/modules` from the Edge-internal `microsoft.visualstudio.com/edge/_git/...` URL and fail — the opposite of "no effect on other teams."

Fix — give it a `False` default so the safe path is "skip", and let Edge''s parent `.gclient` override it via `custom_vars`:
```python
vars = {
  'edge_git': '...',
  'edge_git_suffix': '',
  'oneds_modules_revision': 'c637015f...',
  'checkout_ms_src_internal': False,   # Edge sets this True via custom_vars
}
```

DEPS
```diff
@@ -0,0 +1,21 @@
+#
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+
+use_relative_paths = True
+
+git_dependencies = 'DEPS'
```
bmehta001, line 8: `git_dependencies = 'DEPS'` declares DEPS authoritative over git submodules for this repo. For a gclient consumer that recurses into this DEPS, the `lib/modules` **submodule won''t be synced**; combined with the condition-gated dep (once the `False` default above is added), a non-Edge gclient consumer would then fetch `lib/modules` from *neither* path. Harmless today (no other gclient consumers of this repo) and completely inert for non-gclient users, but worth a comment so the interaction is intentional.

DEPS
```diff
@@ -0,0 +1,21 @@
...
+
+vars = {
+  'edge_git': 'https://microsoft.visualstudio.com/edge/_git',
+  'edge_git_suffix': '',
+  'oneds_modules_revision': 'c637015fbbe904ed556d27e3b9072f6f2a5ee401',
```
bmehta001, line 13: This pins `lib/modules` to a hardcoded SHA on an **Edge Azure DevOps mirror** (`edge_git`), while `.gitmodules` points `lib/modules` at `../cpp_client_telemetry_modules.git` (GitHub `microsoft/cpp_client_telemetry_modules`). That is two sources of truth on two different remotes, with nothing enforcing they stay in sync (the PR body notes "keep in sync" but it''s manual). Consider a CI check asserting `oneds_modules_revision` equals the submodule SHA, or a comment naming the submodule as canonical.

DEPS
```diff
@@ -0,0 +1,21 @@
...
+}
+
```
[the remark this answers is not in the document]
bmehta001: Not quite correct - please take a look at my comments instead

lkarra2:
Abandoning the pull request for now, exploring option to add this file to Edge downstream branch.

PR #1520, opened 2026-08-09
Default Win32 transport to WinHTTP and harden request lifetimes

### LINKED ISSUE #1333: GetAppLocalTempDirectory leaks a RoInitialize call
In Microsoft::Applications::Events::GetAppLocalTempDirectory (lib/utils/utils.cpp), we have this block of code:

`            auto hr = RoInitialize(RO_INIT_MULTITHREADED)
            /* Ignoring result from call to `RoInitialize` as either initialzation is successful, or else already
             * initialized and it should be ok to proceed in both the scenarios */
            UNREFERENCED_PARAMETER(hr);`

If you call RoInitialize without a matching call to RoUninitialize, you leak the com initialization on this thread. See the guidance at https://learn.microsoft.com/en-us/windows/win32/api/roapi/nf-roapi-roinitialize:

> Each successful call to RoInitialize, including those that return S_FALSE, must be balanced by a corresponding call to RoUninitialize.

This is caught by Application Verifier as AVRF_STOP_THREADPOOL_DIRTY_COM_STATE.

### LINKED ISSUE #1437: 100% CPU spin on CancelAllRequests
**Describe your environment.** macOS

**Steps to reproduce.**
Customer hit it randomly, I have their spindump.

Not 100% CPU :)

100% CPU usage due to spin yield-loop in CancelAllRequests

**Additional context.**

Two problems:
1. No lock on m_httpCallbacks — it reads m_httpCallbacks.empty() without holding m_httpCallbacksMtx, but every other access to m_httpCallbacks (in handleSendRequest and onHttpResponse) is guarded by that mutex. This is a data race (undefined behavior). On arm64 especially, the thread may never see the list become empty due to store/load ordering.

2. The yield-loop can spin forever — CancelAllRequests() tells the HTTP client to cancel, but the callbacks are removed from m_httpCallbacks in onHttpResponse(), which is scheduled asynchronously via scheduleOnHttpResponse() → scheduleTask(). If the task dispatcher is blocked or shut down (e.g., PauseTransmission has stopped dispatching), the response handler never runs, m_httpCallbacks never empties, and this loop spins indefinitely — which is exactly what the spindump shows.
 

The call chain in the spindump confirms this: PauseTransmission() calls into cancelAllRequests(), but pausing transmission likely also stops or stalls the task dispatcher that would normally run onHttpResponse to drain the callback list. The yield-loop has no timeout or fallback, so it burns CPU forever.

A fix would need to either:
Use a condition variable signaled from onHttpResponse instead of polling
Add a timeout to the loop and forcibly clear callbacks after it expires
Ensure onHttpResponse can still execute even when transmission is paused (or drain synchronously)
 
Spindump:
```
CPU Time:         9.756s (42.6G cycles, 143.3G instructions, 0.30c/i, 4.37GHz avg)
Num threads:      7
Note:             2 idle work queue threads omitted

  Thread 0x3b1aee    1001 samples (1-1001)    priority 0-20 (base 20)    cpu time 9.755s (42.6G cycles, 143.3G instructions, 0.30c/i, 4.37GHz avg)
  1001  thread_start + 8 (libsystem_pthread.dylib + 7196) [0x185ac6c1c]
    1001  _pthread_start + 136 (libsystem_pthread.dylib + 27736) [0x185acbc58]
      1001  void* std::__1::__thread_proxy[abi:ne190102]
<redacted>

        1001  sense::Telemetry::OneDsTelemetryClient::PauseTransmission() (in telemetryd_v2.dwarf) (/Users/cloudtest/vss/_work/1/s/src/edr/libtelemetry/1ds/OneDsTelemetryClient.cpp:348) [0x104c99f64]
          1001  Microsoft::Applications::Events::LogManagerImpl::PauseTransmission() (in telemetryd_v2.dwarf) (/Users/cloudtest/vss/_work/1/s/src/edr/MAT/MAT/lib/api/LogManagerImpl.cpp:461) [0x104be8344]
            1001  std::__1::__function::__func<Microsoft::Applications::Events::TelemetrySystem::TelemetrySystem(Microsoft::Applications::Events::ILogManager&, Microsoft::Applications::Events::IRuntimeConfig&, Microsoft::Applications::Events::IOfflineStorage&, Microsoft::Applications::Events::IHttpClient&, Microsoft::Applications::Events::ITaskDispatcher&, Microsoft::Applications::Events::IBandwidthController*, Microsoft::Applications::Events::LogSessionDataProvider&)::$_2, std::__1::allocator<Microsoft::Applications::Events::TelemetrySystem::TelemetrySystem(Microsoft::Applications::Events::ILogManager&, Microsoft::Applications::Events::IRuntimeConfig&, Microsoft::Applications::Events::IOfflineStorage&, Microsoft::Applications::Events::IHttpClient&, Microsoft::Applications::Events::ITaskDispatcher&, Microsoft::Applications::Events::IBandwidthController*, Microsoft::Applications::Events::LogSessionDataProvider&)::$_2>, bool ()>::operator()() (in telemetryd_v2.dwarf) (/Applications/Xcode_16.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX15.5.sdk/usr/include/c++/v1/__functional/function.h:313) [0x104bda300]
              996   Microsoft::Applications::Events::HttpClientManager::cancelAllRequests() (in telemetryd_v2.dwarf) (/Users/cloudtest/vss/_work/1/s/src/edr/MAT/MAT/lib/http/HttpClientManager.cpp:153) [0x104c01174]
                801   swtch_pri + 8 (libsystem_kernel.dylib + 3208) [0x185a87c88] (running)
                193   swtch_pri + 8 (libsystem_kernel.dylib + 3208) [0x185a87c88]
                 *94    ??? (kernel.release.t6050 + 24180) [0xfffffe000b3dde74] (running)
                 *62    ??? (kernel.release.t6050 + 24180) [0xfffffe000b3dde74]
                   *59    ??? (kernel.release.t6050 + 2149104) [0xfffffe000b5e4af0] (running)
                   *3     ??? (kernel.release.t6050 + 2150656) [0xfffffe000b5e5100]
                     *2     ??? (kernel.release.t6050 + 2122228) [0xfffffe000b5de1f4]
                       *2     ??? (kernel.release.t6050 + 843264) [0xfffffe000b4a5e00]
                         *2     ??? (kernel.release.t6050 + 842980) [0xfffffe000b4a5ce4] (running)
                     *1     ??? (kernel.release.t6050 + 2122276) [0xfffffe000b5de224] (running)
                 *21    ??? (kernel.release.t6050 + 843532) [0xfffffe000b4a5f0c] (runnable)
                 *15    ??? (kernel.release.t6050 + 27000) [0xfffffe000b3de978] (running)
                 *1     ??? (kernel.release.t6050 + 27000) [0xfffffe000b3de978]
                   *1     ??? (kernel.release.t6050 + 843632) [0xfffffe000b4a5f70]
                     *1     ??? (kernel.release.t6050 + 2168000) [0xfffffe000b5e94c0]
                       *1     ??? (kernel.release.t6050 + 2148624) [0xfffffe000b5e4910]
                         *1     ??? (kernel.release.t6050 + 26064) [0xfffffe000b3de5d0]
                           *1     ??? (kernel.release.t6050 + 316424) [0xfffffe000b425408] (running)
                2     cthread_yield + 36 (libsystem_pthread.dylib + 15400) [0x185ac8c28] (running)
              3     cthread_yield + 44 (libsystem_pthread.dylib + 15408) [0x185ac8c30] (running)
              2     Microsoft::Applications::Events::HttpClientManager::cancelAllRequests() (in telemetryd_v2.dwarf) (/Users/cloudtest/vss/_work/1/s/src/edr/MAT/MAT/lib/http/HttpClientManager.cpp:152) [0x104c01168] (running)


  Thread 0x3b1aec    1001 samples (1-1001)    priority 20 (base 20)    last ran 1011.253s ago
  1001  thread_start + 8 (libsystem_pthread.dylib + 7196) [0x185ac6c1c]
    1001  _pthread_start + 136 (libsystem_pthread.dylib + 27736) [0x185acbc58]
      1001  void* std::__1::__thread_proxy[abi:ne190102]<std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct>>, void (*)(void*), void*>>(void*) (in telemetryd_v2.dwarf) (/Applications/Xcode_16.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX15.5.sdk/usr/include/c++/v1/__thread/thread.h:209) [0x104c20c58]
        1001  Microsoft::Applications::Events::PlatformAbstraction::WorkerThread::threadFunc(void*) (in telemetryd_v2.dwarf) (/Users/cloudtest/vss/_work/1/s/src/edr/MAT/MAT/lib/pal/WorkerThread.cpp:0) [0x104c20410]
          1001  Microsoft::Applications::Events::PlatformAbstraction::Event::wait(unsigned int) const (in telemetryd_v2.dwarf) (/Users/cloudtest/vss/_work/1/s/src/edr/MAT/MAT/lib/./pal/WorkerThread.hpp:55) [0x104c20924]
            1001  bool std::__1::condition_variable::wait_for[abi:ne190102]<long long, std::__1::ratio<1l, 1000l>, Microsoft::Applications::Events::PlatformAbstraction::Event::wait(unsigned int) const::'lambda0'()>(std::__1::unique_lock<std::__1::mutex>&, std::__1::chrono::duration<long long, std::__1::ratio<1l, 1000l>> const&, Microsoft::Applications::Events::PlatformAbstraction::Event::wait(unsigned int) const::'lambda0'()) (in telemetryd_v2.dwarf) (/Applications/Xcode_16.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX15.5.sdk/usr/include/c++/v1/__condition_variable/condition_variable.h:205) [0x104c20a54]
              1001  std::__1::condition_variable::__do_timed_wait(std::__1::unique_lock<std::__1::mutex>&, std::__1::chrono::time_point<std::__1::chrono::system_clock, std::__1::chrono::duration<long long, std::__1::ratio<1l, 1000000000l>>>) + 104 (libc++.1.dylib + 26864) [0x1859e18f0]
                1001  <patched truncated backtrace>
                  1001  __psynch_cvwait + 8 (libsystem_kernel.dylib + 17676) [0x185a8b50c]
                   *1001  psynch_cvcontinue + 0 (pthread + 17948) [0xfffffe000b316abc]

  Thread 0x3b1aed    1001 samples (1-1001)    priority 20 (base 20)    last ran 22618.041s ago
  1001  thread_start + 8 (libsystem_pthread.dylib + 7196) [0x185ac6c1c]
    1001  _pthread_start + 136 (libsystem_pthread.dylib + 27736) [0x185acbc58]
      1001  void* std::__1::__thread_proxy[abi:ne190102]
< redacted >
        1001  sense::Telemetry::OneDsTelemetryClient::LogEvent(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, bool) (in telemetryd_v2.dwarf) (/Users/cloudtest/vss/_work/1/s/src/edr/libtelemetry/1ds/OneDsTelemetryClient.cpp:344) [0x104c99de8]
          1001  Microsoft::Applications::Events::Logger::LogEvent(Microsoft::Applications::Events::EventProperties const&) (in telemetryd_v2.dwarf) (/Users/cloudtest/vss/_work/1/s/src/edr/MAT/MAT/lib/api/Logger.cpp:334) [0x104bedfac]
            1001  Microsoft::Applications::Events::Logger::submit(CsProtocol::Record&, Microsoft::Applications::Events::EventProperties const&) (in telemetryd_v2.dwarf) (/Users/cloudtest/vss/_work/1/s/src/edr/MAT/MAT/lib/api/Logger.cpp:589) [0x104bf15cc]
              1001  Microsoft::Applications::Events::LogManagerImpl::sendEvent(Microsoft::Applications::Events::IncomingEventContext* const&) (in telemetryd_v2.dwarf) (/Users/cloudtest/vss/_work/1/s/src/edr/MAT/MAT/lib/api/LogManagerImpl.cpp:726) [0x104be8f10]
                1001  std::__1::recursive_mutex::lock() + 16 (libc++.1.dylib + 29536) [0x1859e2360]
                  1001  _pthread_mutex_firstfit_lock_slow + 216 (libsystem_pthread.dylib + 6368) [0x185ac68e0]
                    1001  <patched truncated backtrace>
                      1001  __psynch_mutexwait + 8 (libsystem_kernel.dylib + 14812) [0x185a8a9dc]
                       *1001  psynch_mtxcontinue + 0 (pthread + 9696) [0xfffffe000b314a80] (blocked by turnstile waiting for telemetryd_v2 [29655] [unique pid 129217] thread 0x3b1aee)
```

bmehta001:
## Summary
- Make WinHTTP the default Win32 desktop telemetry transport while preserving WinInet as an explicit compatibility option.
- Select and link exactly one Win32 HTTP backend across CMake, Visual Studio projects, samples, tests, and the vcpkg packaging path.
- Add bounded request draining so cancellation and shutdown complete without deadlocks, use-after-free, or duplicate terminal callbacks across WinHTTP, WinInet, Apple, Curl, and C API transports.
- Harden TLS and response handling: Curl always verifies certificates and hostnames, Windows Microsoft-root checks operate on negotiated HTTPS certificates, and transport setup/response failures retain accurate outcomes.
- Strengthen CI diagnostics and deterministic cancellation coverage, including stale macOS kqueue event handling in the test reactor.

## Notable fixes
- Prevent WinHTTP teardown deadlocks when sends fail synchronously or cancellation races with callbacks.
- Keep WinHTTP callback and request state alive until native handles finish closing.
- Guarantee exactly one terminal completion for Apple and Curl setup, cancellation, callback, and worker-failure paths.
- Correct Curl HTTP-version selection, callback signatures, socket handling, option validation, and request-body lifetime.
- Make MSVC Annex K overlap validation overflow-safe and avoid process-terminating invalid-parameter paths.
- Improve Windows test timeout/minidump diagnostics and remove timing or external-network dependencies from cancellation tests.

## Compatibility notes
- The supported Windows floor remains Windows 10 and Windows Server 2016, matching the repository support matrix. Windows 7, Windows 8, and Windows 8.1 are not supported; legacy target macros retained in project files are for binary compatibility and do not extend that support matrix.
- The managed wrapper now requires .NET Framework 4.8. Applications targeting .NET Framework 4.0 through 4.7.2 cannot reference the new assembly and must remain on an earlier SDK release or update their target framework.
- WinInet rejects HTTPS requests with `http.msRootCheck=true` before any headers are sent because WinInet cannot safely evaluate the negotiated certificate before committing sensitive headers. Use the default WinHTTP transport when this policy is required.
- The in-repository vcpkg port now fails clearly when its pinned fallback source predates the Windows transport selector. The release pin and SHA remain a separate post-release update.
- Win32 desktop builds now use WinHTTP by default. Set `MATSDK_USE_WININET=ON` in CMake or `MATSDK_USE_WININET=true` in Visual Studio builds to retain WinInet.
- `http.sslVerify=false` remains accepted for configuration compatibility but is ignored. Curl always enables peer and hostname verification; development environments using private or self-signed certificates must configure `http.sslCaInfo` with a trusted CA bundle.
- When `http.msRootCheck=true`, HTTPS requests fail closed if the negotiated server chain cannot be retrieved or the Microsoft-root policy cannot be evaluated. Plain HTTP requests are unaffected.
- Apple `CancelAllRequests()` now retires created-but-unsent requests instead of waiting indefinitely; a later send receives exactly one `Aborted` callback.
- The vcpkg pin and SHA will be updated in a separate PR.

## Validation
- Visual Studio 2026, x64 Release, WinHTTP: all 639 unit tests passed.
- Visual Studio 2026, x64 Release, WinInet: the pre-send MS-root rejection regression passed.
- Linux/WSL Debug: 14 targeted worker, dispatcher, storage, manager-lifetime, and Curl tests passed.
- Linux/WSL with `-fno-exceptions`: the Curl transport translation unit compiled successfully; CI now enforces this configuration.
- GitHub Actions covers WinHTTP and WinInet Windows matrices, Linux/macOS/iOS builds, vcpkg consumers, Android, public headers, and CodeQL.


lib/offline/OfflineStorageHandler.cpp
```diff
@@ -163,43 +340,105 @@ namespace MAT_NS_BEGIN {
...
+                m_offlineStorageMemory->GetRecords(false, EventLatency_Unspecified);
+            std::vector<StorageRecord> persistentRecords;
+            persistentRecords.reserve(memoryRecords.size());
+            for (auto& record : memoryRecords)
+            {
+                if (record.persistence != EventPersistence_DoNotStoreOnDisk)
+                {
+                    persistentRecords.push_back(std::move(record));
+                }
+            }
```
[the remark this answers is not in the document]
bmehta001: Fixed in 2a9cd8f0. Flush now restores EventPersistence_DoNotStoreOnDisk records to the RAM queue before persisting the disk-eligible batch, with regression coverage proving the memory-only record remains retrievable.

lib/http/HttpClient_Apple.mm
```diff
@@ -288,43 +519,46 @@ void Cancel()
...
+            {
+                ids.push_back(item.first);
+            }
+        }
+        for (const auto& id : ids)
+        {
+            CancelRequestAsync(id);
         }
         PAL::sleep(100);
     }
```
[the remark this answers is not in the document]
bmehta001: Fixed in c82e48cc. Apple delegate callbacks are now marked reentrant; CancelAllRequests cancels peer tasks but does not wait on the serialized delegate queue, while non-callback callers still drain normally.

lib/offline/OfflineStorageHandler.cpp
```diff
@@ -101,25 +250,53 @@ namespace MAT_NS_BEGIN {
...
+            {
+                m_offlineStorageDisk->Shutdown();
+            }
+        }
+        if (notifySaved)
         {
-            m_offlineStorageDisk->Shutdown();
+            OnStorageRecordsSaved(savedRecords);
         }
+        FinishTeardown();
```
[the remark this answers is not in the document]
bmehta001: Fixed in c82e48cc. Shutdown now always calls FinishTeardown on exceptional exits and rethrows the original exception. FailedShutdownCompletesTeardown covers a throwing storage shutdown followed by another Shutdown call.

tests/functests/APITest.cpp
```diff
@@ -1252,8 +1285,66 @@ TEST(APITest, LogManager_BadStoragePath_Test)
...
+            return RequestOutcome{};
+        }
+        windowsClient->SetMsRootCheck(enforceMsRoot);
+
+        std::unique_ptr<IHttpRequest> request(client->CreateRequest());
+        request->SetMethod("POST");
+        request->SetUrl("https://mobile.events.data.microsoft.com/OneCollector/1.0/");
+        std::vector<uint8_t> body {'{', '}'};
+        request->SetBody(body);
+        client->SendRequestAsync(request.release(), &callback);
```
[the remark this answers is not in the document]
bmehta001: Fixed in c82e48cc. The test retains the request unique_ptr through both waits and passes request.get() to the borrowing transport API.

lib/http/HttpClient_CAPI.cpp
```diff
@@ -177,48 +259,126 @@ namespace MAT_NS_BEGIN {
...
-        auto operation = std::make_shared<HttpClient_Operation>(simpleRequest, callback, m_cancelFn);
+        auto operation = std::make_shared<HttpClient_Operation>(
+            state->ownerId, requestId, callback, cancelFn);
         AddPendingOperation(requestId, operation);
 
-        m_sendFn(&capiRequest, &OnHttpResponse);
+        try
+        {
+            sendFn(&capiRequest, &OnHttpResponse);
+        }
```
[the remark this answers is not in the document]
bmehta001: Fixed in e707f7c7 (pushed via cfd9480f). Completions arriving during the borrowed-pointer handoff are copied and deferred without blocking the external completion thread, then delivered after the send hook returns. ConcurrentCompletionWaitsForSendHookToReturn covers the request-lifetime boundary.

lib/http/HttpClient_WinHttp.cpp
```diff
@@ -0,0 +1,1647 @@
...
+        std::wstring wMethod = to_utf16_string(m_request->m_method);
+        m_isHttps = (urlc.nScheme == INTERNET_SCHEME_HTTPS);
+        // Latch the policy for this request: the callbacks that enforce it run
+        // long after send() returns, and the setting can be changed at any time.
+        m_msRootCheckRequired =
+            m_clientState->msRootCheck.load(std::memory_order_acquire);
+        m_hRequest = ::WinHttpOpenRequest(
+            m_hConnect, wMethod.c_str(), path, NULL, WINHTTP_NO_REFERER,
+            WINHTTP_DEFAULT_ACCEPT_TYPES,
+            WINHTTP_FLAG_REFRESH | (m_isHttps ? WINHTTP_FLAG_SECURE : 0));
```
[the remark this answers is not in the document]
bmehta001: Fixed in e707f7c7. WinHttpCrackUrl now returns dynamically sized component views; the object name retains the query and stops before any client-only fragment. QueryStringIsPreservedWithoutFragment covers both behaviors.

tests/unittests/HttpClientCAPITests.cpp
```diff
@@ -173,15 +209,230 @@ TEST(HttpClientCAPITests, Cancel)
...
+        });
+
+    std::thread sender([&] {
+        httpClient.SendRequestAsync(request.get(), &responseCallback);
+    });
+    {
+        std::unique_lock<std::mutex> lock(gateMutex);
+        ASSERT_TRUE(gateCV.wait_for(
+            lock, std::chrono::seconds(5), [&] { return sendEntered; }));
+    }
```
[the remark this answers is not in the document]
bmehta001: Fixed in e707f7c7. The timeout branch now releases the gate, joins the sender, cancels the pending operation, and only then reports the fatal failure.

lib/http/HttpClient_Apple.mm
```diff
@@ -162,68 +200,216 @@ - (void)URLSession:(NSURLSession*)session
...
+                if (!registered)
+                {
+                    bool cancelled = false;
+                    {
+                        std::lock_guard<std::mutex> lock(m_mutex);
+                        cancelled = m_cancelRequested;
+                    }
+                    Complete(cancelled ? HttpResult_Aborted : HttpResult_LocalFailure);
+                    return;
+                }
```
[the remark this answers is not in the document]
bmehta001: Fixed in e707f7c7. Failed delegate registration now cancels the suspended task before locally completing and removing the request.

lib/http/HttpClient_Apple.mm
```diff
@@ -162,68 +200,216 @@ - (void)URLSession:(NSURLSession*)session
...
-                m_dataTask = [session uploadTaskWithRequest:m_urlRequest fromData:postData];
+                [task cancel];
+                return;
             }
-
-            // Register before resume so the streaming delegate has the buffer and
-            // completion handler in place before any response data arrives.
-            [sessionDelegate registerTask:m_dataTask handler:m_completionMethod];
-            [m_dataTask resume];
+            Complete(cancelled ? HttpResult_Aborted : HttpResult_LocalFailure);
```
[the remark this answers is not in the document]
bmehta001: Fixed in e707f7c7. The exception path now cancels any created but unregistered task; registered tasks still defer terminal delivery to the delegate.

lib/http/HttpClientManager.cpp
```diff
@@ -105,7 +169,30 @@ namespace MAT_NS_BEGIN {
             static_cast<unsigned>(ctx->recordIdsAndTenantIds.size()), ctx->latency, latencyToStr(ctx->latency), static_cast<unsigned>(ctx->packageIds.size()),
             ctx->httpRequest->GetId().c_str(), static_cast<unsigned>(ctx->httpRequest->GetSizeEstimate()));
 
-        m_httpClient.SendRequestAsync(ctx->httpRequest, callback);
+        try
+        {
+            m_httpClient.SendRequestAsync(ctx->httpRequest, callback);
+        }
+        catch (const std::exception& ex)
```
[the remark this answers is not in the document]
bmehta001: Fixed in c4c8a5fc. Exception-only request rejection and cancellation handlers now compile behind HAVE_EXCEPTIONS, with direct-call paths for no-exception builds. Both Win32 mini projects compile successfully with VS 2026 after retargeting.

lib/http/HttpClient_CAPI.cpp
```diff
@@ -127,14 +214,31 @@ namespace MAT_NS_BEGIN {
...
         }
     }
 
+    HttpClient_CAPI::~HttpClient_CAPI() noexcept
+    {
+        try
+        {
+            CancelAllRequests();
+        }
+        catch (const std::exception& ex)
```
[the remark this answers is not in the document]
bmehta001: Fixed in c4c8a5fc. C API teardown, send handoff, cancellation, and completion handlers now have explicit HAVE_EXCEPTIONS branches and ordinary direct-call behavior when exceptions are disabled.

lib/offline/OfflineStorageHandler.cpp
```diff
@@ -101,25 +250,61 @@ namespace MAT_NS_BEGIN {
...
+        if (!BeginTeardown())
         {
-            m_offlineStorageMemory->ReleaseAllRecords();
-            Flush();
-            m_offlineStorageMemory->Shutdown();
+            return;
         }
-        if (nullptr != m_offlineStorageDisk)
+
+        try
```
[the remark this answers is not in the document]
bmehta001: Fixed in c4c8a5fc. Offline shutdown now uses a scope guard to finish teardown on every exit, while exception-only flush recovery and queue rollback are guarded with direct no-exception paths.

lib/offline/SQLiteWrapper.hpp
```diff
@@ -216,15 +216,30 @@ namespace MAT_NS_BEGIN {
...
 
-        bool initialize(std::string const& filename, bool deletePrevious, size_t maxHeapLimit = 0)
+        ~SqliteDB()
+        {
+            // Finalize prepared statements and close the database even if
+            // shutdown() was not called explicitly (e.g. the owning storage was
+            // destroyed without Shutdown()). shutdown() is idempotent -- it
+            // returns immediately once m_db is null -- so an earlier explicit
+            // shutdown() makes this a no-op.
+            shutdown();
```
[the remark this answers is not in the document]
bmehta001: Fixed in 01654744. `OfflineStorage_SQLite` now has a defaulted destructor, allowing its `unique_ptr<SqliteDB>` member to run the idempotent RAII shutdown instead of asserting first in Debug. Added `DestructionWithoutShutdownClosesDatabase`, which verifies cleanup and SQLite instance-count release on Windows and WSL Debug.

examples/cmake/MSTelemetrySample.cmake
```diff
@@ -13,9 +13,20 @@ if(NOT EXISTS "${MATSDK_LIB_DIR}/libmat.a"
...
+  set(MATSDK_LIBRARY MSTelemetry::mat)
+  set(MATSDK_SAMPLE_DEPENDENCY_LIBS "")
+else()
+  find_library(MATSDK_LIBRARY NAMES mat HINTS "${MATSDK_LIB_DIR}" NO_DEFAULT_PATH)
+  if(NOT MATSDK_LIBRARY)
+    message(FATAL_ERROR "Could not find libmat under ${MATSDK_LIB_DIR}. Set MATSDK_INSTALL_DIR or MATSDK_LIB_DIR.")
+  endif()
+  find_package(CURL REQUIRED)
+  find_package(ZLIB REQUIRED)
+  set(MATSDK_SAMPLE_DEPENDENCY_LIBS CURL::libcurl ZLIB::ZLIB)
```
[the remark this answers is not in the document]
bmehta001: Fixed in 01654744. The legacy fallback now always resolves Zlib but resolves and links Curl only on platforms whose default transport uses Curl; Win32, Apple-native, and Android-native fallbacks no longer require it. I validated the Win32 fallback with Curl discovery disabled and built `SampleCppMini` with VS 2026.

lib/filter/EventFilterCollection.cpp
```diff
@@ -52,9 +86,10 @@ namespace MAT_NS_BEGIN
         {
             return true;
         }
-        std::lock_guard<std::mutex> lock(m_filterLock);
-        return std::all_of(m_filters.cbegin(), m_filters.cend(), 
-            [&properties](const std::unique_ptr<IEventFilter>& filter)
+
+        auto filters = std::atomic_load(&m_filters);
```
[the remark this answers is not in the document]
bmehta001: Fixed in 01654744. `m_filters` is now the single source of truth: `CanEventPropertiesBeSent`, `Size`, and `Empty` each atomically load the published immutable snapshot, and the redundant `m_size` publication was removed. This also removes the double atomic read from the event path.

examples/cs/SampleCsNet48/SampleCsNet48.csproj
```diff
@@ -11,7 +11,7 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>CLI</RootNamespace>
     <AssemblyName>SampleCsNet40</AssemblyName>
-    <TargetFrameworkVersion>v4.0</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.8.1</TargetFrameworkVersion>
```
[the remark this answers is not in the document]
bmehta001: Fixed in 63d36189 and 01654744. Renamed the directory, project file, assembly metadata, solution entry, build target, and documentation to SampleCsNet48, matching the .NET Framework 4.8.1 target. The separate lowercase net40 native wrapper remains because it genuinely targets .NET Framework 4.0.

lib/offline/OfflineStorageHandler.cpp
```diff
@@ -163,65 +247,124 @@ namespace MAT_NS_BEGIN {
...
+                        recordsToRecover = m_offlineStorageMemory->GetRecords(
+                            false, EventLatency_Unspecified, MAX_RECORDS_PER_STORAGE_BATCH);
+                        if (recordsToRecover.empty())
+                        {
+                            break;
+                        }
+
+                        const size_t drainedBatchSize = recordsToRecover.size();
+                        recordsRemaining -= std::min(recordsRemaining, drainedBatchSize);
+                        const size_t batchSaved = m_offlineStorageDisk->StoreRecords(recordsToRecover);
```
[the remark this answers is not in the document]
bmehta001: Fixed in ccf58965. Both batched and per-record flush paths now separate `EventPersistence_DoNotStoreOnDisk` records before invoking disk storage and return those records to the in-memory queue. Added focused tests for both batching modes that require only the normal record to be counted as persisted.

lib/offline/OfflineStorageHandler.cpp
```diff
@@ -36,6 +56,11 @@ namespace MAT_NS_BEGIN {
         m_cacheMemorySizeLimitInBytes(0),
         m_isStorageFullNotificationSend(false)
     {
+        if (!m_storageProvider)
+        {
+            throw std::invalid_argument("OfflineStorageHandler requires a storage provider");
```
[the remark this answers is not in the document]
bmehta001: Fixed in ccf58965. The provider validation now uses `MATSDK_THROW`, and every new `try`/`catch` handler in `OfflineStorageHandler.cpp` uses `MATSDK_TRY`/`MATSDK_CATCH`. The VS 2026 Release x64 `win32-mini-dll` target, which disables exception handling, rebuilds successfully.

lib/api/LogManagerImpl.cpp
```diff
@@ -368,9 +378,27 @@ namespace MAT_NS_BEGIN
...
+        }
+        try
+        {
+            LOCKGUARD(ILogManagerInternal::managers_lock);
+            ILogManagerInternal::managers.erase(this);
+        }
+        catch (...)
+        {
+            std::fputs("Log manager registry cleanup failed\n", stderr);
+        }
```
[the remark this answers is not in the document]
bmehta001: Fixed in 01b6781b. The destructor teardown/registry cleanup and EndActivity boundaries now use MATSDK_TRY with catches compiled only under HAVE_EXCEPTIONS, preserving direct execution in no-exception builds. Verified by the VS 2026 Debug|x64 win32-mini-lib build.

lib/pal/TaskDispatcher_CAPI.cpp
```diff
@@ -52,7 +72,55 @@ namespace PAL_NS_BEGIN {
                     LOG_ERROR("Unhandled non-standard exception in CAPI task");
                 }
```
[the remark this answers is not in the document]
bmehta001: Fixed in 01b6781b. The C API callback now uses MATSDK_TRY and exception-only MATSDK_CATCH handlers, so no-exception builds invoke the task directly. Verified by the VS 2026 Debug|x64 win32-mini-lib build.

lib/pal/WorkerThread.cpp
```diff
@@ -254,22 +399,46 @@ namespace PAL_NS_BEGIN {
                         catch (...) {
                             LOG_ERROR("Unhandled non-standard exception in worker task");
                         }
```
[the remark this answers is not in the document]
bmehta001: Fixed in 01b6781b. The worker callback now uses MATSDK_TRY and exception-only MATSDK_CATCH handlers, preserving direct task execution without exception support. Verified by the VS 2026 Debug|x64 win32-mini-lib build.

lib/offline/SQLiteWrapper.hpp
```diff
@@ -354,6 +392,11 @@ namespace MAT_NS_BEGIN {
                         *m_instanceCount -= 1;
                     } else if (*m_instanceCount == 1) {
                         *m_instanceCount = 0;
+                        if (m_ownsTempDirectory != nullptr && *m_ownsTempDirectory) {
+                            ::sqlite3_free(sqlite3_temp_directory);
+                            sqlite3_temp_directory = nullptr;
+                            *m_ownsTempDirectory = false;
+                        }
                         g_sqlite3Proxy->sqlite3_shutdown();
                     }
```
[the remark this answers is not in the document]
bmehta001: Fixed in 5797ed9c. Both normal teardown and initialization-failure cleanup now call sqlite3_shutdown() before touching sqlite3_temp_directory, and only free/clear the owned pointer when shutdown returns SQLITE_OK. A regression test injects SQLITE_BUSY, verifies the pointer remains valid, then verifies a later successful shutdown releases it. The full 42-test OfflineStorageTests_SQLite fixture passes.

baijumeswani (commented):
Thank you for the substantial hardening work here. The new response-size limits, stronger curl TLS defaults, WinHTTP certificate checks, and broader cancellation tests are valuable. I found several correctness, security, and compatibility concerns that should be addressed before merging. The most important are a WinInet credential-exposure window, callback and worker lifetime bugs, no-exception build failures, and a Windows 7 TLS regression. I left specific mitigation suggestions inline.

lib/http/HttpClient_WinInet.cpp
```diff
@@ -6,40 +6,422 @@
...
+            if (m_hWinInetRequest == nullptr || shouldStopSetup())
+            {
+                error = ERROR_INTERNET_OPERATION_CANCELLED;
+            }
+            else
+            {
+                m_sendIssued = true;
+                m_pendingApi = PendingApi::StagedHeaders;
+                ++m_asyncApiDepth;
+                result = ::HttpSendRequestExA(
```
baijumeswani, line 557: The Microsoft-root check happens only after `HttpSendRequestExA` completes. At that point WinInet may already have sent headers such as `APIKey`, `Tickets`, and authentication tokens to a Windows-trusted certificate that is not Microsoft-rooted. The request is rejected later, but the credentials may already be exposed. Please enforce this policy before any sensitive headers are sent. If WinInet cannot provide a safe pre-header callback, consider rejecting `msRootCheck` with WinInet or routing those requests through WinHTTP. An integration test should verify that a rejected server receives no sensitive headers.

lib/http/HttpClient_WinHttp.cpp
```diff
@@ -0,0 +1,1651 @@
...
+
+        // Install the callback context explicitly, before anything else can
+        // fail. Relying on WinHttpSendRequest's dwContext instead would strand
+        // the context (and the strong reference it holds) whenever the send
+        // fails before WinHTTP records it -- WinHTTP would then report
+        // HANDLE_CLOSING with a zero context and nothing would free it. Once
+        // the option is set, the handle owns the context and HANDLE_CLOSING is
+        // guaranteed to hand it back. Until then unique_ptr owns it, so no path
+        // out of this function can leak it.
+        std::unique_ptr<WinHttpCallbackContext> context(new WinHttpCallbackContext(shared_from_this()));
```
baijumeswani, line 947: This and later setup operations can throw after the wrapper has already been registered and saved the raw application callback. The caller treats a thrown `SendRequestAsync` as rejected and may destroy the callback, but the registered wrapper can call it later during cancellation or teardown, causing a use-after-free. Please make registration exception-safe: either remove the wrapper, disable later callback delivery, and close any native state before rethrowing, or convert every failure after registration into exactly one terminal callback without throwing. Please add fault-injection coverage for allocation and callback exceptions after registration.

lib/pal/WorkerThread.cpp
```diff
@@ -29,89 +33,181 @@ namespace PAL_NS_BEGIN {
...
+            std::thread threadToJoin;
+            bool joined = false;
+            {
+                LOCKGUARD(m_lock);
+                enqueueShutdownItemLocked();
+                if (!m_hThread.joinable()) {
+                    return;
+                }
+                if (m_hThread.get_id() == this_id) {
                     m_hThread.detach();
```
baijumeswani: Detaching here does not mark deletion as deferred. If this worker calls `Join()` and another thread then releases the final `shared_ptr` before the worker exits, the object is deleted while `threadFunc()` still uses it. Please do not detach during an ordinary self-`Join()`, or keep an explicit worker-owned lifetime reference until `threadFunc()` finishes. Add a test where a task calls `Join()` and another thread releases the final owner before the task returns.

lib/pal/WorkerThread.cpp
```diff
@@ -29,89 +33,181 @@ namespace PAL_NS_BEGIN {
...
+                } else {
+                    threadToJoin = std::move(m_hThread);
+                }
+            }
+            try {
+                if (threadToJoin.joinable()) {
+                    threadToJoin.join();
```
baijumeswani, line 131: `m_joinLock` is held while this blocks in `join()`. If the running worker callback also calls `Join()`, it waits for `m_joinLock` while this thread waits for that callback to return, causing a permanent deadlock. Please choose one join owner while holding the lock, release the lock before waiting, and let other callers wait on a separate completion condition. A deterministic external-join plus worker-join test would cover this case.

lib/http/HttpClient_Curl.cpp
```diff
@@ -65,100 +307,342 @@ namespace MAT_NS_BEGIN {
...
         }
 
-        auto curlOperation = std::make_shared<CurlHttpOperation>(curlRequest->m_method, curlRequest->m_url, callback, requestHeaders, curlRequest->m_body, false, HTTP_CONN_TIMEOUT, m_sslVerify, sslCaInfo);
-        curlRequest->SetOperation(curlOperation);
-        
-        // The lifetime of curlOperation is guarnteed by the call to result.wait() in the d'tor.  
-        curlOperation->SendAsync([this, callback, requestId](CurlHttpOperation& operation) {
-            this->EraseRequest(requestId);
+        std::shared_ptr<CurlHttpOperation> operation;
+        try
```
baijumeswani, line 342: This file and `HttpClient_Curl.hpp` now contain several unguarded `try`/`catch` blocks. The SDK supports builds with exceptions disabled through `HAVE_EXCEPTIONS` and `MATSDK_TRY`/`MATSDK_CATCH`; GCC and Clang reject raw exception syntax under `-fno-exceptions`. Please convert all new curl handlers to the repository abstraction and provide the direct no-exception paths. Please also add a Linux curl compile job using `-fno-exceptions`, since minimal downstream builds rely on this mode.

lib/http/HttpClient_WinHttp.cpp
```diff
@@ -0,0 +1,1651 @@
...
+    // without depending on a logged-on interactive user or that user's
+    // Internet Explorer settings -- unlike WinInet's
+    // INTERNET_OPEN_TYPE_PRECONFIG, which requires one. This is why WinHTTP,
+    // not WinInet, is Microsoft's documented recommendation for services and
+    // other non-interactive processes. On an older OS that rejects this access
+    // type, fall back to the machine-wide WinHTTP proxy configuration. This is
+    // the documented pre-Windows-8.1 behavior and avoids bypassing enterprise
+    // proxies entirely. Only fall back for the compatibility error; other
+    // failures should not be hidden by a second, unrelated WinHttpOpen call.
+    HINTERNET session = ::WinHttpOpen(
```
baijumeswani, line 1545: The project still targets Windows 7, but this new default WinHTTP session does not configure secure protocols. On Windows 7, WinHTTP may not enable TLS 1.2 unless the machine has the required update and registry configuration, so telemetry to TLS-1.2-only collectors can stop working even where the previous WinInet path worked. Please explicitly configure TLS 1.2 with a compatible fallback, retain WinInet for the Windows 7 target, or raise and document the minimum supported Windows version. CI currently tests only modern Windows.

lib/offline/OfflineStorageHandler.cpp
```diff
@@ -163,65 +243,147 @@ namespace MAT_NS_BEGIN {
...
+                            recordsToRecover.begin(), recordsToRecover.end(),
+                            [](StorageRecord const& record)
+                            {
+                                return record.persistence != EventPersistence_DoNotStoreOnDisk;
+                            });
+                        std::vector<StorageRecord> memoryOnlyRecords(
+                            std::make_move_iterator(memoryOnlyBegin),
+                            std::make_move_iterator(recordsToRecover.end()));
+                        recordsToRecover.erase(memoryOnlyBegin, recordsToRecover.end());
+                        ReturnRecordsToMemory(memoryOnlyRecords);
```
baijumeswani: Reinserting memory-only records inside this loop lets the next batch select the same records again because `MemoryStorage` reads and writes at the back of each queue. Those repeated records still reduce `recordsRemaining`, so older records that should be written to disk may never be reached. During shutdown they are then discarded with the memory queue. Please keep all memory-only records aside until the original queue snapshot has been fully processed, then reinsert them after the loop. Add a test with more than one batch of newer memory-only records followed by an older persistent record.

lib/pal/TaskDispatcher_CAPI.cpp
```diff
@@ -157,11 +241,32 @@ namespace PAL_NS_BEGIN {
...
         }
 
-        return (!taskId.empty()) ? m_cancelFn(taskId.c_str()) : false;
+        if (taskId.empty())
+        {
+            return false;
+        }
+
+        const bool wasRunning = capiTask->RequestCancel();
+        m_cancelFn(taskId.c_str());
```
baijumeswani: The Boolean result from the host cancellation callback is ignored. If the host rejects cancellation, this code still marks and erases the task and returns success. If the host later runs that task, the callback cannot find it and silently drops the work. Please preserve the return value and only mark or erase a queued task when the host confirms cancellation. If cancellation fails, leave the task registered so its normal callback can run.

Solutions/net48/net48.vcxproj
```diff
@@ -21,10 +21,10 @@
   </ItemGroup>
   <PropertyGroup Label="Globals">
     <ProjectGuid>{DC91621E-A203-42DF-8E03-3A23DD0602B1}</ProjectGuid>
-    <TargetFrameworkVersion>v4.0</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.8</TargetFrameworkVersion>
```
baijumeswani, line 24: Replacing the only .NET Framework 4.0 wrapper with a 4.8-only wrapper raises the runtime and consumer target requirement to 4.8. Applications targeting 4.0 through 4.7.2 will no longer be able to reference the new assembly. Please keep and package the existing compatible target alongside 4.8, or treat this as a documented major-version breaking change with package metadata that prevents incompatible upgrades.

tools/ports/cpp-client-telemetry/portfile.cmake
```diff
@@ -134,13 +134,19 @@ if(MATSDK_ROOT_CMAKE MATCHES "MATSDK_MINIMAL_SQLITE"
...
+endif()
+
 vcpkg_cmake_configure(
     SOURCE_PATH "${SOURCE_PATH}"
     OPTIONS
         ${MATSDK_PINNED_SOURCE_OPTIONS}
         -DMATSDK_SQLITE_PROVIDER=${MATSDK_VCPKG_SQLITE_PROVIDER}
         -DBUILD_SHARED_LIBS=${MATSDK_VCPKG_BUILD_SHARED_LIBS}
         -DMATSDK_ANDROID_HTTP_CLIENT=${MATSDK_ANDROID_HTTP_CLIENT}
+        -DMATSDK_USE_WININET=${MATSDK_USE_WININET}
```
baijumeswani, line 162: This option works for the in-repository overlay, but the normal fallback still downloads `v3.10.173.1`, which does not contain the WinHTTP selector. Published/fallback installs therefore remain WinInet-only and the new `wininet` feature has no effect. Please update the pinned source and hash in this PR, or fail clearly when the selected source does not support this option. The vcpkg CI should also exercise the pinned fallback rather than only the local overlay.

lib/http/HttpClientManager.cpp
```diff
@@ -86,16 +127,39 @@ namespace MAT_NS_BEGIN {
...
+        // HttpCallback and scheduled response tasks retain a reference to this
+        // manager, so non-reentrant destruction must be a full callback lifetime
+        // barrier. Reentrant destruction is unsupported because the active
+        // callback itself must still unwind through this object.
+#ifndef NDEBUG
+        {
+            std::lock_guard<std::mutex> lock(m_httpCallbacksMtx);
+            for (auto const& active : m_activeHttpCallbacks)
+            {
+                assert(active.second != std::this_thread::get_id());
```
baijumeswani: This check protects only debug builds. In release builds, destroying the manager from one of its active response or debug callbacks can finish destruction immediately; when the callback resumes, it accesses the destroyed mutex and callback containers. That turns unsupported callback behavior into a use-after-free. Please defer final destruction until the active callback unwinds, or move callback tracking into reference-counted state that survives the manager. Add a release-mode test that destroys a dynamically owned manager from inside `requestDone`.

Solutions/net48/net48.vcxproj
bmehta001: I will document the change

lib/http/HttpClient_WinInet.cpp
bmehta001: Valid. WinInet now rejects HTTPS requests requiring the Microsoft-root policy before creating or sending the request, so sensitive headers cannot leave the process. Added a WinInet regression that sets an APIKey header and verifies immediate failure.

lib/http/HttpClient_WinHttp.cpp
bmehta001: Valid. After registration, WinHTTP setup is now exception-contained and converts failures into terminal completion rather than exposing a rejected callback that native state could later invoke.

lib/pal/WorkerThread.cpp
bmehta001: Valid. Ordinary self-Join no longer detaches the worker. A deterministic regression now has the task call Join while another thread releases the final owner and verifies destruction waits for the worker to finish.

lib/pal/WorkerThread.cpp
bmehta001: Valid. Self-Join is detected before taking the external join-serialization lock, eliminating the lock inversion. Added a deterministic worker-Join plus external-Join regression.

lib/http/HttpClient_Curl.cpp
bmehta001: Valid. Curl exception handlers are now guarded for HAVE_EXCEPTIONS, and CI compiles the Curl translation unit with -fno-exceptions. The same validation passed locally in WSL.

lib/http/HttpClient_WinHttp.cpp
bmehta001: The legacy target macros are real, but Windows 7 is not in the repository supported-platform matrix; support starts at Windows 10 and Windows Server 2016. I documented explicitly that Windows 7/8/8.1 are unsupported and that legacy target macros do not extend the support matrix, rather than adding an untested Windows 7 TLS fallback.

lib/offline/OfflineStorageHandler.cpp
bmehta001: Valid. Batched flush now holds memory-only records aside until the original queue snapshot is fully drained, then reinserts them. Added a regression with two full memory-only batches followed by an older persistent record.

lib/pal/TaskDispatcher_CAPI.cpp
bmehta001: Valid. The C API dispatcher now honors the host cancellation result and leaves a rejected task registered and runnable. Added a regression that rejects cancellation and then runs the normal callback.

Solutions/net48/net48.vcxproj
bmehta001: Valid breaking-change concern. The PR description and Windows build documentation now state that the managed wrapper requires .NET Framework 4.8 and that 4.0-4.7.2 consumers must remain on an earlier SDK release or retarget.

tools/ports/cpp-client-telemetry/portfile.cmake
bmehta001: Valid. The port now detects a pinned source that predates MATSDK_USE_WININET and fails with an actionable message instead of silently producing the wrong transport. The release REF/SHA update remains the documented post-release step.

lib/http/HttpClientManager.cpp
bmehta001: Valid. Callback bookkeeping now lives in shared registry state retained by each callback, so it survives reentrant manager destruction until the callback unwinds. Added a release-mode regression that destroys a dynamically owned manager from requestDone.

lib/http/HttpClientManager.cpp
```diff
@@ -192,37 +423,84 @@ namespace MAT_NS_BEGIN {
...
+        // re-enters onHttpResponse().
         const auto cancelStart = std::chrono::steady_clock::now();
         cancelAllRequestsAsync(bestEffort ? m_cancelDrainTimeout : std::chrono::milliseconds::zero());
 
         // Drain callbacks through the condition variable signaled by onHttpResponse.
-        std::unique_lock<std::recursive_mutex> lock(m_httpCallbacksMtx);
+        auto registry = m_callbackRegistry;
+        std::unique_lock<std::mutex> lock(registry->mutex);
+        std::thread::id const callerThread = std::this_thread::get_id();
+        auto callbacksDrainedForCaller = [registry, callerThread] {
```
baijumeswani, line 536: This returns `true` as soon as it finds the callback running on the thread that is destroying the manager. It does not check whether other callbacks are still active or queued. Those callbacks can still use the manager through raw `this` pointers, including the scheduled task created in `scheduleOnHttpResponse`, after the manager has been freed. Please exclude only the current callback from the wait and continue waiting for every other callback. Scheduled response work should hold reference-counted state that does not depend on the manager remaining alive. A test with two callbacks, where the first destroys the manager while the second is queued or running, would cover this case.

lib/http/HttpClient_WinHttp.cpp
```diff
@@ -0,0 +1,1666 @@
...
+    }
+
+    void DispatchEvent(std::unique_lock<std::mutex>& lock, HttpStateEvent type)
+    {
+        if (m_appCallback != nullptr && !isCallbackCalled)
+        {
+            void* handle = static_cast<void*>(m_hRequest);
+            IHttpResponseCallback* callback = m_appCallback;
+            auto state = m_clientState;
+            ++m_stateCallbackDepth;
```
baijumeswani: The callback depth is incremented here, but the cleanup below is skipped if `OnHttpStateEvent` throws. The new catch in `send()` then aborts the request, but completion sees a nonzero depth and remains pending forever. This can leave the request registered and make cancellation or shutdown wait indefinitely. Please use an RAII guard that always reacquires the lock, decreases the callback counters, and releases any pending completion before the exception is handled. Please also add a test with a state callback that throws during setup.

lib/offline/OfflineStorageHandler.cpp
```diff
@@ -163,65 +243,149 @@ namespace MAT_NS_BEGIN {
...
+            size_t dbSizeBeforeFlush = (m_offlineStorageMemory != nullptr) ? m_offlineStorageMemory->GetSize() : 0;
+            if ((m_offlineStorageMemory) && (dbSizeBeforeFlush > 0) && (m_offlineStorageDisk))
+            {
+                size_t totalSaved = 0;
+                if (IsBatchedStorageFlushEnabled())
+                {
+                    // Drain only the records present when this flush started so
+                    // producers cannot keep the flush alive indefinitely.
+                    size_t recordsRemaining = m_offlineStorageMemory->GetRecordCount();
+                    std::vector<StorageRecord> memoryOnlyRecords;
```
baijumeswani: Holding memory-only records aside fixes the normal flush path, but this vector exists only inside the `try` block. If a later `StoreRecords()` call throws, the outer catch recovers only the current `recordsToRecover` batch. The memory-only records already moved into this vector are destroyed and lost. The SQLite implementation can rethrow from `StoreRecords`, so this path is reachable. Please keep this collection in the outer recovery scope and return it to memory on both success and exception paths. A test where an early batch contains memory-only records and a later disk batch throws would verify the fix.

baijumeswani (commented):
Thank you for addressing the earlier review. Most of the original concerns are now resolved. I found three remaining edge cases in the updated lifetime and storage fixes. They can still cause use-after-free, stalled shutdown, or record loss, so I recommend addressing them before merging.

lib/http/HttpClientManager.cpp
bmehta001: Fixed in 41bb6ed3. Queued response work now captures the shared callback registry instead of a raw manager pointer, manager access is counted per thread, and destruction waits for every peer callback while excluding only the current reentrant callback. `DestroyingManagerWaitsForPeerCallback` covers the two-callback lifetime case.

lib/http/HttpClient_WinHttp.cpp
bmehta001: Fixed in 41bb6ed3. WinHTTP state callbacks now use an RAII guard that always restores callback depth/thread accounting and releases pending completion during exception unwinding. `ThrowingStateCallbackStillCompletesRequest` verifies setup callback exceptions still terminate the request.

lib/offline/OfflineStorageHandler.cpp
bmehta001: Fixed in 41bb6ed3. The memory-only collection now lives in the outer recovery scope and is returned on both success and exception paths. `BatchedFlushRecoversMemoryOnlyRecordsWhenLaterBatchThrows` verifies recovery after a later disk batch throws.

lib/http/HttpClientManager.cpp
```diff
@@ -86,64 +298,104 @@ namespace MAT_NS_BEGIN {
...
+        }
+        catch (...)
+        {
+            LOG_ERROR("Unhandled non-standard exception while releasing failed HTTP request");
         }
 
-        delete callback;
+        try
+        {
+            requestFailureComplete(ctx);
```
baijumeswani: `requestFailed(ctx)` above calls external code and can destroy this manager. Reentrant destruction is allowed for the current active thread, so execution can return here after `this` has been freed and access `requestFailureComplete`, causing a use-after-free. This is reachable when `requestDone` throws and the failure handler tears down its owner. Please dispatch these routes through registry-owned state, or check that the manager is still attached after `requestFailed` before accessing another member. A test where `requestDone` throws and `requestFailed` destroys the manager would cover the path.

lib/http/HttpClient_WinInet.cpp
```diff
@@ -6,40 +6,475 @@
...
+      public:
+        StateCallbackGuard(
+            WinInetRequestWrapper& owner,
+            std::thread::id callbackThread) noexcept
+            : m_owner(owner),
+              m_callbackThread(callbackThread)
+        {
+        }
+
+        ~StateCallbackGuard() noexcept(false)
```
baijumeswani: This destructor can call `finishStateCallback()`, which may perform terminal completion, allocate a response, and invoke application `OnHttpResponse` code. If the state callback caused cancellation and then threw, this destructor runs while that exception is already unwinding. A second exception from allocation or the terminal callback then calls `std::terminate`. `SetupGuard` has the same problem, while the new WinHTTP guard contains completion exceptions. Please make both WinInet guards non-throwing and contain completion failures, or defer completion until the original callback exception has been handled. Add a test where a state callback cancels and throws.

baijumeswani (commented):
Thank you for fixing the three previous findings. I confirmed those fixes and their focused tests. I found two remaining reentrant exception paths in the new lifetime handling that can still cause use-after-free or process termination.

lib/http/HttpClient_WinInet.cpp
bmehta001: Fixed in b716c792. SetupGuard and StateCallbackGuard are now noexcept and contain terminal-completion exceptions, matching the WinHTTP guard behavior. The new WinInet regression cancels from OnConnecting and then throws, verifying setup unwinding still produces exactly one aborted terminal response without termination.

lib/http/HttpClientManager.cpp
bmehta001: Fixed in b716c792. Failure routing now stays in the callback registry-owned frame and rechecks ManagerUse::IsAttached() after requestFailed returns before accessing requestFailureComplete. The new RequestFailureCanDestroyManager regression makes requestDone throw, destroys the manager from requestFailed, and verifies the completion route is not touched afterward.

lib/http/HttpClient_WinInet.cpp
```diff
@@ -6,40 +6,497 @@
...
+        ~SetupGuard() noexcept
+        {
+#if HAVE_EXCEPTIONS
+            try
+            {
+#endif
+            m_owner.finishSetup();
+#if HAVE_EXCEPTIONS
+            }
+            catch (...)
```
baijumeswani, line 169: Catching the completion exception prevents `std::terminate`, but it can leave the request permanently registered. `onRequestComplete()` sets `m_terminalCallbackStarted` before allocating the response and well before `eraseRequest()`. If response construction or header processing throws, this catch suppresses the error after the terminal latch is set. Later cancellation returns immediately because it sees that latch, while `CancelAllRequests()` can wait forever for the request registry to drain. Please make terminal finalization exception-safe: prepare allocating state before claiming terminal ownership, or guarantee handle closure and registry removal on every failure after the claim. Fault-injection coverage for an allocation or header-processing failure during deferred completion would verify this path.

baijumeswani (commented):
Thank you for fixing the manager failure-route lifetime issue and the WinInet double-exception termination path. One exception-safety gap remains in the new WinInet completion handling.

lib/http/HttpClient_WinInet.cpp
bmehta001: Fixed in f4a47308. Terminal ownership now immediately installs a no-throw finalization guard that closes native handles and removes the request from the registry on every post-claim exit, including response allocation and header-processing exceptions. The same invariant is applied to WinHTTP after auditing the analogous default-transport path. TerminalFinalizationExceptionDoesNotStrandRequest injects a preparation failure for both transports and verifies CancelAllRequests still drains.

baijumeswani (approved):

PR #1529, opened 2026-09-01
Add maintainer onboarding and release guidance

bmehta001:
## Summary
- add a central maintainer onboarding guide with repository reading order and development workflow
- document Windows, WSL/Linux, vcpkg, Android, macOS, iOS, simulator, physical-device, and end-to-end Kusto validation
- document PR expectations and the complete SDK, SPM, vcpkg, and downstream release process
- replace obsolete version schedules with the current four-component version and support policy
- link the new guidance from the README

## Validation
- ran the repository's pinned `misspell` checker on all changed Markdown files
- checked whitespace errors with `git diff --check`
- verified relative links introduced by the new and rewritten documentation


docs/maintainer-onboarding.md
```diff
@@ -0,0 +1,351 @@
...
+```powershell
+# Platform, configuration, optional .props/.targets file, transport
+.\build-tests.cmd x64 Release "" WinHTTP
+.\build-tests.cmd x64 Debug "" WinHTTP
+
+# Exercise the opt-in legacy transport when transport code or selection changes
+.\build-tests.cmd x64 Release "" WinInet
+```
+
+The default desktop transport is WinHTTP. It supports services and
```
lalitb, line 127: Are you sure about this - The default used to be `WinInet` ?

bmehta001: Was waiting until #1520 went in to merge this PR, as that sets the default desktop transport to WinHTTP, so now this should be accurate

docs/maintainer-onboarding.md
```diff
@@ -0,0 +1,377 @@
...
+1. Start from current `main` after required changes have merged and required CI
+   is green.
+2. Generate `lib/include/public/Version.hpp`:
+
+   ```powershell
+   .\tools\gen-version.cmd
+   ```
+
+   ```bash
+   ./tools/gen-version.sh
```
baijumeswani: This command fails when run from the repository root. `version-node.js` reads `../Solutions/version.txt` relative to the current directory, so it looks outside the repository. I reproduced the failure with `ENOENT: ../Solutions/version.txt`.

Please either fix `tools/gen-version.sh` so it changes to its own directory before running Node, or document a working command such as `(cd tools && ./gen-version.sh)`. The same command also appears in `docs/versioning.md`.

docs/maintainer-onboarding.md
```diff
@@ -0,0 +1,377 @@
...
+1. Update local `main` to the merged release commit.
+2. Create the four-component tag `vX.Y.Z.W` on that commit and push it to the
+   canonical repository.
+3. Publish a non-draft, non-prerelease GitHub Release for the tag with release
+   notes.
+4. Verify that the GitHub source archives resolve and contain the generated
+   version.
+
+Do not move or force-update a published release tag. Correct a bad release with
+a new build/version component and release notes describing the correction.
```
baijumeswani: A correction that changes only the fourth version component cannot be published to SPM. Both `vX.Y.Z.1` and `vX.Y.Z.2` map to the same `X.Y.Z` SPM tag, and the workflow skips the second one. This could leave SPM users on the broken package.

Please say that a corrected release which must reach SPM needs a new `X.Y.Z` value, or update the SPM version mapping before using a fourth-component-only hotfix.

docs/maintainer-onboarding.md
```diff
@@ -0,0 +1,377 @@
...
+| Cross-platform or release-critical change | Windows, WSL/Linux, Android, macOS, iOS Simulator, package-consumer tests, and representative physical-device checks where feasible |
+| High-risk integration or ingestion change | Embed the local SDK in a real consuming application, exercise the changed scenario end to end, and verify expected events in the authorized telemetry destination |
+
+Manual review and CI can be sufficient for a small, low-risk change. A larger
+change should use more independent environments. Record what was and was not
+tested in the PR so reviewers can assess residual risk.
+
+### Leak analysis
+
+The [memory leak analysis workflow](../.github/workflows/memory-leak-analysis.yml)
```
baijumeswani: This workflow and the baseline linked below are not present in the current PR head, so both links are broken here. They were added by #1536, which has now merged into `main`. Please update this branch from current `main` before merging so the guide and the files it describes land together.

baijumeswani (commented):
The guide is useful and most commands match the repository. I found three documentation issues that should be corrected before approval.

docs/maintainer-onboarding.md
bmehta001: Already addressed by the a387a250 merge of current main into this PR branch before this review. Both .github/workflows/memory-leak-analysis.yml and .github/memory-leak-baseline.csv are present in the PR head (and current main is an ancestor). No additional change needed for these links.

docs/maintainer-onboarding.md
bmehta001: Fixed in 0f5d0d02: docs/maintainer-onboarding.md and docs/versioning.md now explain that an SPM correction requires a new X.Y.Z and that changing only W leaves the existing SPM package in place. Verified against the tag mapping and existing-tag skip in .github/workflows/spm-release.yml.

docs/maintainer-onboarding.md
bmehta001: Fixed in 0f5d0d02: tools/gen-version.sh now enters its own directory before calling version-node.js. The documented ./tools/gen-version.sh invocation works from the repository root (also verified from tools in Git Bash and WSL), so both docs can keep the same command.

baijumeswani (approved):

PR #1530, opened 2026-09-04
Restore buildless C# CodeQL analysis

bmehta001:
## Summary
- add a dedicated CodeQL job for C# source
- use buildless analysis so legacy .NET Framework and Xamarin projects do not need to compile
- upload the C# snapshot required by the SFI-PS2.1 Continuous SDL KPI

## Rationale
PR #1357 removed C# from the language matrix after the Windows 2022 job could not observe a successful C# compilation. CodeQL now supports `build-mode: none` for C#, which analyzes the checked-in source without invoking those legacy build paths.

## Validation
- parsed `.github/workflows/codeql-analysis.yml` as YAML
- checked the branch diff for whitespace errors

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -77,6 +77,27 @@ jobs:
     - name: Perform CodeQL Analysis
       uses: github/codeql-action/analyze@5595ccaf912efad79be6eef63a5619ff05969be3 # v4.37.6
 
+  analyze-csharp:
+    name: Analyze C#
+    runs-on: windows-2022
+    permissions:
+      contents: read
+      actions: read
+      security-events: write
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
+
+    - name: Initialize CodeQL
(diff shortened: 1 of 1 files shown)


lalitb (approved):
LGTM. The C# scan passed, though the [warning](https://github.com/microsoft/cpp_client_telemetry/actions/runs/33908359757/job/101138606956?pr=1530) shows gaps in resolving calls and types that could cause it to miss issues. We can track improving the analysis quality as a follow-up.

``
 Scanning C# code completed successfully, but the scan encountered issues. This may be caused by problems identifying dependencies or use of generated source code. Some metrics of the database quality are: Percentage of calls with call target: 51 % (threshold 85 %). Percentage of expressions with known type: 78 % (threshold 85 %). Ideally these metrics should be above their thresholds. Addressing these issues is advisable to avoid false-positives or missing results. If they cannot be addressed, consider scanning C# using either the `autobuild` or `manual` [build modes](https://docs.github.com/en/code-security/code-scanning/creating-an-advanced-setup-for-code-scanning/codeql-code-scanning-for-compiled-languages#comparison-of-the-build-modes).
``

ThomsonTan (approved):

PR #1535, opened 2026-09-14
Avoid session sidecars for in-memory storage

bmehta001:
## Summary

- treat SQLite's `:memory:` cache path as non-file-backed for session data
- prevent creation or deletion attempts for a physical `:memory:.ses` sidecar
- add regression coverage for create, reset, and delete

Borrowed from the fix in microsoft/onnxruntime#32538, where ONNX Runtime currently carries this as a dependency patch. This upstreams the behavior so consumers can receive it without maintaining a local patch.

## Validation

- WSL/Ubuntu 24.04: `LogSessionDataFuncTests.*` (5/5 passed)
- Windows/Visual Studio 2026: changed production and test translation units compile; the existing functional-test target then fails to link on unresolved `AITelemetrySystem` because Azure Monitor headers are auto-detected while its implementation is not linked
- `git diff --check`
- clang-format 20.1.8 on changed line ranges


bmehta001:
Fixed the cross-platform LoggerTests.LogSession_CanEventPropertiesBeSentReturnsTrue_CallsSubmit failure in 89ebcc5f. The :memory: path now generates ephemeral session timestamp/UID values while still suppressing the .ses sidecar. The exact failing unit test and the strengthened in-memory sidecar functional test pass on both Visual Studio 2026 and WSL.

bmehta001:
Merged microsoft/cpp_client_telemetry#1532 and fixed the two deterministic CI regressions from its new main state in 6bc46bff: older CMake FindSQLite3 modules are adapted from SQLite::SQLite3 to SQLite3::SQLite3, and Android setup no longer requests the removed SDK 	ools package. Local validation: installed system-package consumer 10/10 on CMake 3.28; focused session tests pass on Visual Studio 2026.

bmehta001:
Superseded by #1520

PR #1536, opened 2026-09-18
Fix Windows network detection leak and add periodic leak reports

### LINKED ISSUE #634: Periodic memory leak analysis reports
There are several known non-service impacting minor memory leaks in SDK:

- `RuntimeConfig_Default.hpp` - may have small leaks at initialization.

- FIFO Storage (used only for EXP configs, not used elsewhere) - appears to leak on config item read.

- initialization of NetworkDetector `lib/pal/desktop/WindowsDesktopNetworkInformationImpl.cpp:87`.

- possible leak in `void EventProperty::copydata(EventProperty const* source)` --- `as_string = new char[len + 1];`

- there could be a few others.

Need to allocate about a week of time to instrument CI runs with Dr. Memory ( https://drmemory.org/ ) on our Windows and Linux, Functional and Unit, test scenarios, as well as basic Sample app. And consistently track the # of leaks. Then eventually catch all leaks - log . Identify valid vs. false-positives, and resolve or document those that are non-service impacting.

bmehta001:
## Summary
- replace desktop `PublicNetworkListManager` use with `Windows.Networking.Connectivity`, preserving metered-network cost updates without loading the leaking `netprofm.dll`
- run Dr. Memory analysis on Windows and Linux after relevant changes reach `main`, with manual dispatch available on demand
- analyze unit tests, functional tests, and SampleCppMini with pinned, SHA-256-verified tooling
- publish leak counts to the job summary and retain raw reports for 90 days
- keep the current leak baseline non-gating while failing explicitly on instrumentation/target errors
- fail Windows leak analysis if `netprofm.dll` is loaded again

## Validation
- Visual Studio 2026 x64 Debug build of UnitTests and SampleCppMini
- focused Windows network/configuration tests: 31 passed
- GCC/WSL syntax check of SampleCppMini
- actionlint 1.7.12 and PowerShell parser validation
- hosted Dr. Memory runs on Windows and Linux, including the `netprofm.dll` regression gate

## Known instrumentation exclusion
- `BasicFuncTests.killSwitchWorks` remains covered by normal CI but is excluded under Dr. Memory because instrumentation changes its exact asynchronous drop count (400 observed versus 100 expected).
- `OfflineStorageTests_SQLite.StoreThousandEventsTakesLessThanASecond` remains covered by normal CI but is excluded under Dr. Memory because instrumentation invalidates its one-second wall-clock performance threshold (1.384 seconds observed).

Closes #634

Related external evidence: microsoft/onnxruntime-genai#2590


lib/pal/desktop/NetworkDetector.cpp
```diff
@@ -571,18 +302,21 @@ namespace MAT_NS_BEGIN
...
-                    // Wait for up to NETDETECTOR_COM_SETTLE_MS ms until:
-                    // - COM object is ready; OR
-                    // - COM object can't be started (pre-Win 8 scenario)
+                    // Wait for up to NETDETECTOR_START_TIMEOUT_MS ms until:
+                    // - the listener is subscribed; OR
+                    // - Windows Runtime network information is unavailable
                     int retry = 1;
                     constexpr int max_retries = 2;
-                    while (isRunning && cv.wait_for(lock, std::chrono::milliseconds(NETDETECTOR_COM_SETTLE_MS))
+                    while (isRunning && cv.wait_for(lock, std::chrono::milliseconds(NETDETECTOR_START_TIMEOUT_MS))
```
baijumeswani: `Start()` can stop waiting while the worker is still starting and has not created its message queue. If `Stop()` runs then, `PostThreadMessage()` can fail and the current code detaches the worker. The worker still uses fields from `NetworkDetector`, so destroying the detector can leave the worker using freed memory.

Please avoid detaching this thread. One option is to add a startup state such as `starting`, `ready`, or `failed`, signal it with the condition variable, and have `Start()` wait for a clear result. `Stop()` should request cancellation and join the worker on every path before the detector is destroyed. A test that delays startup past the timeout would help cover this case.

lib/pal/desktop/NetworkDetector.cpp
```diff
@@ -359,55 +149,37 @@ namespace MAT_NS_BEGIN
...
-            {
-                LOG_ERROR("cannot query IID_IUnknown!!!");
+            networkStatusCallbackState = std::make_shared<CallbackState>();
+            const auto callbackState = networkStatusCallbackState;
+            const auto currentNetworkCost = m_currentNetworkCost;
+            const auto networkInformation = networkInfoStats;
+            networkStatusChangedHandler = Callback<INetworkStatusChangedEventHandler>(
+                [callbackState, currentNetworkCost, networkInformation](IInspectable*) -> HRESULT {
+                    if (callbackState->acceptCallbacks.load(std::memory_order_acquire)) {
+                        RefreshNetworkCost(networkInformation.Get(), *currentNetworkCost);
```
baijumeswani: A callback can read `acceptCallbacks` as true just before shutdown changes it to false. That callback may then finish this call and broadcast `EVT_NET_CHANGED` after `Stop()` has returned and other SDK objects are being destroyed. The shared pointers keep the cache and COM object alive, but they do not keep event listeners safe.

A simpler shutdown model would be for this callback to post a refresh message to the listener thread instead of doing the refresh and broadcast here. Then unregistering the callback and joining the listener thread would guarantee that no network event is sent after `Stop()` returns. Please also add a test that races a network callback with shutdown.

.github/workflows/memory-leak-analysis.yml
```diff
@@ -0,0 +1,217 @@
+name: Memory leak analysis
+
+on:
```
baijumeswani, line 3: This workflow is described as periodic, but it has no scheduled trigger. It also reports the current counts without comparing them with an earlier result, so a leak increase can be missed unless someone checks each run by hand.

Please add a weekly or monthly `schedule` trigger. It would also help to keep a reviewed baseline and show a clear warning when any count increases. The warning can stay non-blocking at first, while tool false positives are sorted out.

lib/pal/desktop/NetworkDetector.cpp
```diff
@@ -8,94 +8,94 @@
...
+            }
+
+            boolean roaming = false;
+            boolean overDataLimit = false;
+            boolean approachingDataLimit = false;
+            NetworkCostType costType = NetworkCostType_Unknown;
+            if (FAILED(connectionCost->get_Roaming(&roaming)) ||
+                FAILED(connectionCost->get_OverDataLimit(&overDataLimit)) ||
+                FAILED(connectionCost->get_ApproachingDataLimit(&approachingDataLimit)) ||
+                FAILED(connectionCost->get_NetworkCostType(&costType))) {
```
baijumeswani: Windows also provides `BackgroundDataUsageRestricted`. This matters for a telemetry library because Windows may allow normal network use while asking apps to limit background transfers. Ignoring it can make this connection look less restricted than it is.

Please query `IConnectionCost2` when it is available and map `BackgroundDataUsageRestricted == true` to the SDK's restricted cost value. If the newer interface is not available, keep the current mapping. Add a mapping test for both the restricted and unsupported-interface cases.

baijumeswani (commented):
The WinRT change is useful and the leak jobs are a good start. I found four areas that should be tightened so shutdown is safe and the leak reporting fully meets the goal of #634. I added a suggested path forward to each inline comment.

lib/pal/desktop/NetworkDetector.cpp
bmehta001: Valid; fixed in fa092818. Start now waits for an explicit Ready/Failed startup result instead of returning after a timeout, and Stop always requests cancellation and joins the worker. The detach path is gone, so destruction cannot leave a worker using NetworkDetector fields.

lib/pal/desktop/NetworkDetector.cpp
bmehta001: Valid; fixed in fa092818. The WinRT callback now only queues NETDETECTOR_REFRESH to the listener thread and captures shared queue state, not detector resources. Stop disables further queueing, posts the stop message, and joins the listener, which guarantees no queued EVT_NET_CHANGED delivery after Stop returns. NetworkDetectorTests.QueuedNetworkCallbackRaceDoesNotOutliveStop exercises callback queueing concurrently with shutdown.

.github/workflows/memory-leak-analysis.yml
bmehta001: Valid; fixed in fa092818. The workflow now runs every Monday at 08:00 UTC and compares all nine leak/possible-leak/reachable metrics against a reviewed Windows/Linux baseline from successful run 35392664701. Increases emit explicit non-blocking GitHub warnings and are shown in the step summary.

lib/pal/desktop/NetworkDetector.cpp
bmehta001: Valid; fixed in fa092818. QueryCurrentNetworkCost now queries IConnectionCost2 when available, maps BackgroundDataUsageRestricted=true to the SDK restricted cost, and preserves the existing mapping when QueryInterface reports that IConnectionCost2 is unavailable. The restrictive mapping test now covers the new flag.

lib/pal/desktop/NetworkDetector.cpp
```diff
@@ -602,31 +393,27 @@ namespace MAT_NS_BEGIN
...
+                        networkStatusCallbackState->listenerThreadId.store(0, std::memory_order_release);
                     }
-                    else
+                    if (startupState == StartupState::Ready &&
+                        !PostThreadMessage(m_listener_tid, NETDETECTOR_STOP, 0, NULL))
                     {
-                        lk.unlock();
-                        netDetectThread.join();
-                        LOG_TRACE("NetworkDetector tid=%p has stopped.", m_listener_tid);
+                        LOG_WARN("NetworkDetector stop message could not be posted.");
```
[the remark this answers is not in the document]
bmehta001: Valid; fixed in 3dc04b1e. Shutdown now uses a manual-reset kernel event and MsgWaitForMultipleObjects, so it does not consume message-queue capacity and takes priority over queued refresh work. Stop disables new refresh posts, signals the event, and always joins. The callback/shutdown race passed 500 consecutive iterations with continuous queueing.

lib/pal/desktop/NetworkDetector.cpp
```diff
@@ -359,175 +182,146 @@ namespace MAT_NS_BEGIN
...
-
-            MSG msg;
-            PostThreadMessage(m_listener_tid, NETDETECTOR_START, 0, 0);
-            cv.notify_all();
+                const DWORD waitResult = MsgWaitForMultipleObjects(
+                    1,
+                    &stopEvent,
+                    FALSE,
+                    INFINITE,
+                    QS_ALLINPUT);
```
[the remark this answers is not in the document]
bmehta001: Valid; fixed in a30f5fd0. The listener now waits with MsgWaitForMultipleObjectsEx and MWMO_INPUTAVAILABLE, so already-queued refresh messages remain observable after one message is removed while the stop event still has priority. The continuous callback/shutdown race again passed 500 iterations.

lib/pal/desktop/NetworkDetector.cpp
```diff
@@ -540,18 +334,42 @@ namespace MAT_NS_BEGIN
...
+                    LOG_ERROR("Unable to create the network detector stop event.");
+                    startupState = StartupState::Failed;
+                    networkStatusCallbackState.reset();
+                    return false;
+                }
                 isRunning = true;
             }
 
             // Start a new thread. Notify waiters on exit.
             netDetectThread = std::thread([this]()
```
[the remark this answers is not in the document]
bmehta001: Valid; fixed in 1b3fc58d. Start and Stop now serialize every access to netDetectThread through m_lifecycleLock, while m_lock remains dedicated to worker state. Stop therefore cannot inspect joinable() between startup state publication and thread assignment. NetworkDetectorTests.ConcurrentStopWaitsForStartupPublication races Stop against startup and passed 25 repeated runs (500 lifecycle iterations). The same commit also adds TotalPossibleLeaks and TotalReachable to the job summary, addressing the suppressed reporting note.

bmehta001:
Addressed the latest review summary concern in commit d6bcef3a: `EVT_NET_CHANGED` is now dispatched through an independently owned, coalescing worker callback rather than synchronously on the detector listener. `Stop()` can therefore join the listener even when invoked reentrantly by an event handler, and the dispatch state prevents any later event. `NetworkDetectorTests.NetworkChangeListenerCanStopDetector` exercises the exact synchronous teardown path.

bmehta001:
Copilot's final suppressed note about excluding OfflineStorageTests_SQLite.StoreThousandEventsTakesLessThanASecond on Linux is not actionable. The Windows exclusion was added after the instrumented Windows run took 1.384 seconds. The hosted Linux Dr. Memory job in run 35392664701 executed the unfiltered unit-test binary, completed successfully, and produced its leak report. Keeping the test on Linux preserves coverage; only the platform demonstrated to violate the timing threshold is excluded.

lib/pal/desktop/NetworkDetector.cpp
```diff
@@ -359,199 +277,200 @@ namespace MAT_NS_BEGIN
...
-                if (GetNetworkInfoStats())
-                {
-                    LOG_INFO("create network list manager...");
-                    hr = CoCreateInstance(
-                        CLSID_NetworkListManager,
-                        nullptr,
-                        CLSCTX_ALL,
-                        IID_INetworkListManager,
-                        (void**)&pNlm);
+                    HRESULT hr = RoInitialize(RO_INIT_MULTITHREADED);
```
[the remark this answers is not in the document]
bmehta001: We don't need to support Windows 7-8.1

.github/memory-leak-baseline.csv
```diff
@@ -0,0 +1,7 @@
+Platform,Scenario,UniqueLeaks,TotalLeaks,LeakBytes,UniquePossibleLeaks,TotalPossibleLeaks,PossibleLeakBytes,UniqueReachable,TotalReachable,ReachableBytes
+Windows,unit-tests,10,113,4256,14,15,7206,467,644,227717
```
baijumeswani: Non-blocking: the latest successful run is already above this baseline for 8 Linux unit-test metrics and 4 Windows unit-test metrics. For example, possible-leak bytes increased from 3,452 to 208,600 on Linux and from 7,206 to 225,649 on Windows. Please check whether these increases are expected after the recent test and dependency changes. Then run the jobs again to confirm the numbers are stable and update this baseline with a short explanation if needed. This will keep future warnings useful.

baijumeswani (commented):
The earlier correctness concerns are addressed and the current CI checks pass. I have one non-blocking baseline follow-up.

baijumeswani (approved):

